PR #7, opened 2021-04-22
HARMONY-826: Clarify install, enforce node version and guard against GDAL_DATA

bilts:
Fixes some small issues new devs were seeing:

1. The README for NVM had an inverted command (use before install) and the recommendation to re-check node version wasn't clear enough.  A dev encountered issues with an incorrect PATH setup even though NVM looked correct
2. Node version isn't enforced at all, which meant an incorrect node version could cause subtle problems
3. If GDAL_DATA is set in the environment, it's likely to cause issues with the gdal installed by our NPM module.  I added a warning when this happens along with instructions to override.

To test item 2, run `npm install` and `nvm use 12.4.0` (or any node version you have installed below v12.14.1), then run `npm test` and `npm run start-dev`  It should error out.
To see the GDAL warning behavior:
No GDAL_DATA set:
```
ts-node -e 'require("./app/util/env"); console.log(require("gdal").SpatialReference.fromUserInput("EPSG:4326").toProj4())'
```

GDAL_DATA set.  This would fail before this change but now prints a warning and succeeds:
```
GDAL_DATA=/tmp ts-node -e 'require("./app/util/env"); console.log(require("gdal").SpatialReference.fromUserInput("EPSG:4326").toProj4())'
```

You can set GDAL_DATA to some non-gdal dir in .env and rerun either of the above commands, causing a failure

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,7 +47,7 @@ You must select "401" as the application type for Harmony to work correctly with
 
 Required:
 * A local copy of this repository.  Using `git clone` is strongly recommended
-* Node.js version 12.  We recommend installing [NVM](https://github.com/nvm-sh/nvm) to add and manage node versions
+* Node.js version 12.  We strongly recommend installing [NVM](https://github.com/nvm-sh/nvm) to add and manage node versions.
 * Mac OSX, Linux, or similar command line tooling.  Harmony is tested to run on OSX >= 10.14 and Amazon Linux 2.  Command-line instructions and bash helper files under [bin/](bin/) are tested on OSX >= 10.14.
 * [git](https://git-scm.com) - Used to clone this repository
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -76,20 +76,24 @@ If you have not yet cloned the Harmony repository, run
 $ git clone https://github.com/nasa/harmony.git
 ```
 
-Ensure node is available and is the correct version, 12.x.x
+Ensure node is available and is the correct version, 12.x.y, where "x" >= 14.
```
indiejames, line 79: maybe should add "y" >= 1?

indiejames (approved):

README.md
bilts: yeah, I really couldn't think of a concise way to say "^12.14.1".  Open to suggestions.  Technically y could be 0 if x > 14.

chris-durbin (approved):
All of the tests are passing for me.

PR #8, opened 2021-04-26
HARMONY-767: Pass user agent to backend services

vinnyinverso:
Tests were already in place for ensuring that clientId was passed to the backend. I felt that it might be overkill to test the formatting of the clientId itself (`harmony/${version} ${whichHarmony}`) but if anyone feels differently, please let me know.

### DIFF
--- a/bin/create-k8s-config-maps-and-secrets
+++ b/bin/create-k8s-config-maps-and-secrets
@@ -6,7 +6,7 @@ HARMONY_VERSION=$(grep version ./package.json \
   | sed 's/[",]//g' \
   | tr -d '[[:space:]]')
 
-USER_AGENT="harmony/${HARMONY_VERSION} ${CLIENT_ID}"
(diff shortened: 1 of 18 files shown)


vinnyinverso:
If you run this locally in conjunction with the [service example](https://github.com/nasa/harmony-service-example), with the corresponding branch (767) of the services lib, and log the full_user_agt, you should see something like **harmony/0.0.0 harmony-local** harmony-service-lib/0.0.1

bin/create-k8s-config-maps-and-secrets
```diff
@@ -1,5 +1,14 @@
 #!/bin/bash
 
+HARMONY_VERSION=$(cat ./package.json \
```
chris-durbin, line 3: Nit: could remove one line with `grep version ./package.json`

vinnyinverso: Updated! Thanks!

bilts (approved):

PR #10, opened 2021-04-28
HARMONY-118: Simplify environment variables

chris-durbin:
Changed example/dotenv to a file that is sourced to set default environment variables and changed some of the defaults to be better suited for local development. The `.env.ts` file no longer has code to set default values reducing some complexity.

I also fixed the template generation script in this PR to fail if any of the parallel template generation fails.

I will add Bamboo variables to all the deployments for places where the defaults changed and we can no longer rely on the original defaults in deployed environments.

In total there are 5 variables needed in .env.

To test: 

1. Move off your existing .env file so that you have your original values saved.
2. Run npm test and verify that the main harmony tests and the query-cmr tests work without any .env file.
3. Run `bin/create-dotenv`
4. Set the 3 blank OAUTH parameters to match what was in your original .env
5. Add any additional overrides you want (for example if you don't usually run on port 3000)
6. Run `bin/start-argo` and verify there are no errors
7. Bring up harmony and make sure a request completes successfully.


.github/workflows/github-actions.yml
```diff
@@ -17,6 +17,6 @@ jobs:
       uses: actions/setup-node@v1
       with:
         node-version: ${{ matrix.node-version }}
-    - run: cp test/test_dot_env .env
+    - run: npm install -g npm@latest
```
bilts, line 20: Is this going to impact installations from nvm?  I'd figure that the docker image would come packaged with the default version of npm, which would be the same as we'd get from nvm.

Wonder if we should just bump to node 14.

README.md
```diff
@@ -104,14 +104,17 @@ Recommended: Add `./node_modules/.bin` to your `PATH`.  This will allow you to r
...
-Copy the file [example/dotenv](example/dotenv) to a file named `.env` in the root project directory.  Follow the instructions in that file to populate any blank variables.  Variables that have values in the example can be kept as-is, as they provide good defaults for local development.  To check environment differences between the example and local, run:
+Harmony uses environment variables for managing much of its configuration. Most of the variables can be defaulted, and harmony provides those defaults suitable for local development in the `env-defaults` file. In order to set up the remaining variables, run the following from the harmony project root:
 
 ```
-$ git diff --no-index .env example/dotenv
+$ bin/create-dotenv
 ```
 
-We recommend doing this any time you receive an example/dotenv update to ensure there are no new variables needed.
+The script will create a file named `.env` in the root project directory containing only those parameters that cannot be defaulted. Open the file and update the values for any of the variables that are currently blank. Detailed information for the environment variables can be found in the `env-defaults` file.
```
bilts, line 113: I think this is all good info to have captured, but noting for future PR that we can probably skip it in a quick start.  I'm also wondering if maybe our README is starting to become what most projects put in their GitHub wikis.  Anyways, no action.

README.md
```diff
@@ -104,14 +104,17 @@ Recommended: Add `./node_modules/.bin` to your `PATH`.  This will allow you to r
...
+
+Harmony reads both the `env-defaults` and `.env` files at startup to determine the configuration. To override any default values, set the desired value in the `.env` file. There is no need to duplicate parameters in the `.env` file if using the default value.
+
+Note: The defaults are suitable for running locally with Mac OS X. If running on Linux there are a couple of parameters that will also need to be overridden and are documented as such in the `env-defaults` file.
```
bilts, line 117: May want to move this note closer to the top, since it will be important.  Or... use `uname` in the create-dotenv script to just figure out Linux

app/util/env.ts
```diff
@@ -65,58 +42,66 @@ const envVars: HarmonyEnv = {} as HarmonyEnv;
...
-    value = envValue;
+    envVars[_.camelCase(envName)] = stringValue;
   }
+  process.env[envName] = stringValue;
+}
 
-  envVars[_.camelCase(envName)] = value;
+const allEnv = { ...envDefaults, ...envOverrides, ...process.env };
+
+for (const k in allEnv) {
```
bilts, line 57: Does it not work to do:

```javascript
for (const k of Object.keys(allEnv)) {
```

And then remove the linter comment at the top of the file?  `for ... in` can be problematic in surprising ways, though this usage should be safe

app/util/env.ts
```diff
@@ -65,58 +42,66 @@ const envVars: HarmonyEnv = {} as HarmonyEnv;
...
+  oauthUid: string;
+  objectStoreType: string;
+  reapableJobAgeMinutes: number;
+  sameRegionAccessRole: string;
+  sharedSecretKey: string;
+  stagingBucket: string;
+  syncRequestPollIntervalMs: number;
+  uploadBucket: string;
+  useLocalstack: boolean;
+}
```
bilts, line 104: I really like this change.  Our defaults were all over the place.

bin/create-k8s-config-maps-and-secrets
```diff
@@ -31,7 +31,7 @@ type: Opaque
 data:
   OAUTH_UID: $(echo -n "${OAUTH_UID}" | base64)
   OAUTH_PASSWORD: $(echo -n "${OAUTH_PASSWORD}" | base64)
-  EDL_USERNAME: $(echo -n "${EDL_USERNAME}" | base64)
-  EDL_PASSWORD: $(echo -n "${EDL_PASSWORD}" | base64)
+  EDL_USERNAME: $(echo -n "${EDL_USERNAME:-unset}" | base64)
```
bilts, line 34: Do we have a note somewhere that this is being deprecated?  It'd be great if we could just delete it.

bin/create-workflow-templates
```diff
@@ -88,7 +88,29 @@ function inject_script_source {
...
+
+  install_template $workflow_file
+  install_template $batch_file
+  install_template $chain_file
+}
+
+# We install templates in parallel to significantly speed up deployment time.
+# We wait on each of the pids we fork so that we error out if the template
+# creation fails for any of the templates.
+function wait_for_pids {
```
bilts, line 106: I could swear I already made this change once.  Either way, glad you found it wasn't there and did it.

env-defaults
```diff
@@ -218,6 +237,12 @@ BUILT_IN_TASK_PREFIX=
 # Version to use for all built-in tasks.  "latest" (local / sandbox), "sit", "uat", or "prod"
 BUILT_IN_TASK_VERSION=latest
 
+# AWS region where harmony is deployed, irrelevant for local development
+AWS_DEFAULT_REGION=us-west-2
+
+# The object store used for storing shapefiles. Only S3 is supported currently.
+OBJECT_STORE_TYPE=s3
```
bilts, line 244: Good catch.  Thinking this should probably be used by defaultObjectStore.  Also thinking we may have over-engineered something to have an env var with only one valid value.

package.json
```diff
@@ -104,7 +104,7 @@
     "shpjs": "^3.5.0",
     "simple-oauth2": "git+https://github.com/bilts/simple-oauth2.git#fix-encoding",
     "source-map-support": "^0.5.19",
-    "sqlite3": "^4.2.0",
+    "sqlite3": "^5.0.2",
```
bilts, line 107: What required these changes, removing pre-gyp and updating sqlite3?

tasks/query-cmr/app/cli.ts
```diff
@@ -68,8 +67,8 @@ export default async function main(args: string[]): Promise<void> {
   const startTime = new Date().getTime();
   const appLogger = logger.child({ application: 'cmr-granule-locator' });
   const options = parser().parse(args);
-  const encrypter = createEncrypter(env.sharedSecretKey);
-  const decrypter = createDecrypter(env.sharedSecretKey);
+  const encrypter = createEncrypter(process.env.SHARED_SECRET_KEY);
```
bilts, line 70: Is the intent to make query-cmr independent from the .env config mechanism?  I think that seems right.

bilts (approved):
Nice work!  This has been a long time coming.  Only minor comments.

.github/workflows/github-actions.yml
chris-durbin: I think I tried with node 15, but ran into dependency problems I wasn't able to sort out quickly, so I just went with keeping the same version of node but updating npm. It looks like nvm with node12 defaults to npm version 6, which is why this was needed.

package.json
chris-durbin: `node-pre-gyp` is no longer supported, I switched it over to `@mapbox/node-pre-gyp`. No building from source was needed with the combination of changes for node-pre-gyp, npm, and this version of sqlite.

tasks/query-cmr/app/cli.ts
chris-durbin: I made these changes thinking that the main harmony .env file was specific to the harmony code (especially with the typescript variable declarations) and that we should just use environment variables directly in query-cmr. It turns out that other namespaces used from the main harmony code pull in the .env file anyway, so this change wasn't necessary in the end, but I still thought it was a better idea.

bin/create-workflow-templates
indiejames: I remember seeing this before as well.

indiejames (approved):

PR #15, opened 2021-05-13
DAS-1101 - Post schema for history_json provenance metadata.

owenlittlejohns:
This PR adds the first version of the JSON schema for the `history_json` global attribute, which is to be used to capture consistent provenance metadata.


app/schemas/history/0.1.0/history-0.1.0.json
```diff
@@ -0,0 +1,53 @@
...
+   "title": "Data Processing History",
+   "description": "A history record of processing that produced a given data file. For more information, see: https://wiki.earthdata.nasa.gov/display/TRT/In-File+Provenance+Metadata+-+TRT-42",
+   "type": [ "array", "object" ],
+   "items": { "$ref": "#/definitions/history_record" },
+   "definitions": {
+      "history_record": {
+         "type": "object",
+         "properties": {
+            "date_time": {
+               "description": "A Date/Time stamp in ISO-8601 format, including time-zone, GMT (or Z) preferred",
```
bilts, line 13: Nit: This should say what the field is for more than how it's formatted.  Something like "The date and time at which the processing occurred"

Even smaller nit: The JSON schema defines a format of "date-time" to be an RFC 3339 section 5.6 string (https://datatracker.ietf.org/doc/html/rfc3339#section-5.6), which is a subset of ISO-8601 and requires a time offset, so a lot of this is redundant and enforced by schema validation

owenlittlejohns: Agreed. As discussed in Slack, I've updated the JSON file to reflect the new description David has in the TRT-42 wiki page.

bilts:
I also just realized that by convention we put a `v` in front of the version in the filename.  I think we should do that here as well.

chris-durbin (approved):

fgreg (approved):

PR #16, opened 2021-05-13
Harmony 817: Renames images to match DockerHub and GitHub repos/orgs. Pushes tasks to DockerHub

bilts:


bilts:
Action succeeded: https://github.com/nasa/harmony/runs/2578426899?check_suite_focus=true

And pushed to DockerHub: https://hub.docker.com/repository/docker/harmonyservices/query-cmr

indiejames (approved):

.github/workflows/publish-tasks.yml
```diff
@@ -0,0 +1,40 @@
...
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v1
+        with:
+          node-version: 12.x
+      - name: Log into registry
+        run: echo "${{ secrets.DOCKERHUB }}" | docker login -u "${DOCKER_USER}" --password-stdin
+      - name: Setup Node
+        run: |
+          npm install -g npm@latest
```
chris-durbin, line 28: I started changing this to `npm install -g npm@7` to explicitly make sure we use version 7.

env-defaults
```diff
@@ -193,11 +193,11 @@ ASF_GDAL_SUBSETTER_IMAGE=asfdataservices/gdal-subsetter:latest
 ASF_GDAL_SUBSETTER_IMAGE_PULL_POLICY='IfNotPresent'
 ASF_GDAL_SUBSETTER_PARALLELISM=2
 
-HARMONY_GDAL_IMAGE=harmony/gdal:latest
-HARMONY_GDAL_IMAGE_PULL_POLICY='IfNotPresent'
-HARMONY_GDAL_PARALLELISM=2
+HARMONY_SERVICE_EXAMPLE_IMAGE=harmonyservices/service-example:latest
+HARMONY_SERVICE_EXAMPLE_IMAGE_PULL_POLICY='IfNotPresent'
```
chris-durbin, line 197: Make sure to set this image pull policy variable in Bamboo. In at least SIT and Sandbox you will need to set it to 'Always' if we're using the latest tag. As a cleanup after you've deployed you should delete any HARMONY_GDAL_* variables from Bamboo.

chris-durbin (approved):
Built and tested locally. Everything worked.

env-defaults
bilts: I wasn't planning to have our environments pull from DockerHub as part of this work, because that will mess with how we're producing and locking artifacts right now.  Those deployments are all going straight to ECR.  I will need to update variable names and image names, though.

.github/workflows/publish-tasks.yml
bilts: Done.  Also filed https://bugs.earthdata.nasa.gov/browse/HARMONY-842 to make us stop needing to do that

PR #17, opened 2021-05-14
HARMONY-830: Include OPeNDAP links with a type of #service

chris-durbin:
Vinny did most of the work for this PR.

There were a few changes:

1.) Switch var-subsetter to use the new chained workflow templates with the granule locator instead of the legacy argo templates. The reason for this is that the code uses two different paths for processing granule links. It made sense to switch to the new template type now so that only the new path for determining granule links mattered.
2.) Nothing was using data_url_pattern in services.yml anymore so that functionality was removed.
3.) Update the new granule links processing code to also look for links that had a rel of #service in addition to #data.

I tested the var subsetter with:
http://localhost:3000/C1234714698-EEDTEST/ogc-api-coverages/1.0.0/collections/%2Fgt1l%2Fland_segments%2Fcanopy%2Fh_canopy/coverage/rangeset?granuleId=G1238479514-EEDTEST&format=application%2Fnetcdf

to verify that on main and on this branch identical results are returned (with a rel of #data for the OPeNDAP link).

I also tested with a new collection that isn't configured for use with harmony, but does contain links with a rel of #service for the OPeNDAP links. I looked at the STAC catalogs generated by the granule locator and verified that for both collections the assets included the OPeNDAP link.

ARCTEST granule:
```
"assets": {
    "data": {
      "href": "https://acdisc.gesdisc.eosdis.nasa.gov/data//Aqua_AIRS_Level3/AIRS3QP5.006/2002/AIRS.2002.09.01.L3.RetQuant_IR005.v6.0.9.0.G13214160515.hdf",
      "title": "AIRS.2002.09.01.L3.RetQuant_IR005.v6.0.9.0.G13214160515.hdf",
      "roles": [
        "data"
      ]
    },
    "opendap": {
      "href": "https://acdisc.gesdisc.eosdis.nasa.gov/opendap/Aqua_AIRS_Level3/AIRS3QP5.006/2002/AIRS.2002.09.01.L3.RetQuant_IR005.v6.0.9.0.G13214160515.hdf",
      "title": "AIRS.2002.09.01.L3.RetQuant_IR005.v6.0.9.0.G13214160515.hdf",
      "description": "The OPENDAP location for the granule. (GET DATA : OPENDAP DATA)",
      "type": "application/x-hdf",
      "roles": [
        "data",
        "opendap"
      ]
    }
  }
```
EEDTEST granule:
```
  "assets": {
    "data": {
      "href": "https://harmony.uat.earthdata.nasa.gov/service-results/harmony-uat-staging/public/sds/staged/ATL08_20200716202251_03280806_003_01.h5",
      "title": "ATL08_20200716202251_03280806_003_01.h5",
      "type": "application/x-hdfeos",
      "roles": [
        "data"
      ]
    },
    "data1": {
      "href": "https://opendap.uat.earthdata.nasa.gov/collections/C1234714698-EEDTEST/granules/EEDTEST-ATL08-003-ATL08_20200716T202251",
      "title": "EEDTEST-ATL08-003-ATL08_20200716T202251",
      "roles": [
        "data"
      ]
    }
  }
```

You can see that the links look a little bit different but both do include the OPeNDAP link. The inconsistency with the roles and the properties of each link are related to the CMR json results returned. Ideally we would use the UMM_JSON CMR response, which consistently return whether links are OPeNDAP links, but that requires a significant change to the harmony CMR parsing code due to spatial representation being quite different in the two formats. We should investigate making that change at some point.


bilts:
It's not quite sitting right that the OPeNDAP links are showing up differently, especially when the one that's "wrong" is on the canonical OPeNDAP server for UAT.  Since we did away with the configured pattern, could we test the URL to see if it includes "opendap", and if so, put it under the "opendap" heading, which is the equivalent of what var subsetter was doing?

app/util/cmr.ts
```diff
@@ -453,3 +455,16 @@ export async function belongsToGroup(
...
+/**
+ * Return all non-inherited links with rel ending in /data# when no regex is provided.
+ *
+ * @param granule - The granule to obtain links from
+ * @returns An array of granule links
+ */
+export function filterGranuleLinks(
+  granule: CmrGranule,
+): CmrGranuleLink[] {
+  return granule.links.filter((g) => (g.rel.endsWith('/data#') || g.rel.endsWith('/service#'))
```
indiejames, line 468: nit: probably slightly more efficient to reorder the terms in your && clause so the test for inherited links short circuits the other tests if it fails. Also, you could replace the two `endsWtih` checks with a single regex.

indiejames, line 468: It looks like you are only ever. using the first link that matches, in which case you should use `find` here instead of `filter`.

chris-durbin:
> It's not quite sitting right that the OPeNDAP links are showing up differently, especially when the one that's "wrong" is on the canonical OPeNDAP server for UAT. Since we did away with the configured pattern, could we test the URL to see if it includes "opendap", and if so, put it under the "opendap" heading, which is the equivalent of what var subsetter was doing?

Made the change to also look at the href so now we get:
```
"assets": {
    "data": {
      "href": "https://harmony.uat.earthdata.nasa.gov/service-results/harmony-uat-staging/public/sds/staged/ATL08_20200716202251_03280806_003_01.h5",
      "title": "ATL08_20200716202251_03280806_003_01.h5",
      "type": "application/x-hdfeos",
      "roles": [
        "data"
      ]
    },
    "opendap": {
      "href": "https://opendap.uat.earthdata.nasa.gov/collections/C1234714698-EEDTEST/granules/EEDTEST-ATL08-003-ATL08_20200716T202251",
      "title": "EEDTEST-ATL08-003-ATL08_20200716T202251",
      "roles": [
        "data",
        "opendap"
      ]
    }
  }
```

app/util/cmr.ts
chris-durbin: Moved the inherited to the first check.

chris-durbin: I figured I would leave this function as is since it returns all the possible data links. I assume that's what we should be doing, it just so happens that right now this old code ends up only sending one data link in backend service messages. This is only relevant in legacy templates which pass links rather than using the STAC catalog from the query-cmr task. In the new style we do include multiple links in the STAC catalog.

bilts (approved):

indiejames (approved):

PR #18, opened 2021-05-14
HARMONY-834: Serve schemas on /schemas

bilts:
Example: http://localhost:3000/schemas/data-operation/0.10.0/data-operation-v0.10.0.json


chris-durbin (changes requested):
This provides access to all of the files under the schemas directory including READMEs and CHANGELOGs, I assume that's fine since they are available via github anyway.

A few things:

1. It's not obvious where to find a schema for an end user - I think it would be nice if /schemas provided a directory style listing of all the schemas that could be navigated to get to the schema you want (this could be a separate ticket if we just want to get something out there and the user making use of it knows the exact location of the schema.
2. We're requiring the user log in to get to the schemas. I don't think these should be behind EDL.
3. The home page documentation should probably link to the schemas route.
4. The route http://localhost:3000/schemas/eoss/0.1.0/eoss-v0.1.0.yml returns an Earthdata error page with the text: "Sorry! schemas/eoss/0.1.0 must be a collection short name or CMR collection identifier, but we could not find a matching collection. Please make sure the collection is correct and that you have access to it."
5. http://localhost:3000/schemas/ogc-api-coverages/1.0.0/ogc-api-coverages-v1.0.0.yml returns a JSON error similar to the eoss error.

I assume we should at least fix 2, 4, and 5.

bilts:
Good finds. So much for a trivial one-liner :)

bilts:
I updated to fix 2, 4, and 5 above.  1 and 3 are going to be a bit more involved and don't help with the underlying need of a documented place to point to for provenance schemas

chris-durbin (approved):
Tested that the eoss and OGC api schema errors are resolved and EDL is no longer required.

PR #28, opened 2021-05-26
HARMONY-375: correct use of secure cookie option

rwat:
The revised Acceptance Criteria for this card include:

1. Verify cookies / tokens are always encrypted in transit (HTTPS).
2. Tokens are never logged or stored unencrypted at rest.
3. Tokens cannot be manipulated by the user and remain valid; to that end, verify Harmony is doing HMAC and Harmony isn't relying on EDL for that functionality.

For item 1, harmony dev and SIT environments need HTTP whereas other environments (esp. PROD) need https. Currently we rely on upstream proxies to handle http-to-https redirects. We have also been using an environment variable, USE_HTTPS, to determine what cookie options to set for which environment. A side goal of this card was to remove the need for USE_HTTPS. To address all of these, I removed the 'secure' cookie option and changed it to 'samesite':'Lax'. HTTPS is ensured, per environment, by upstream web proxies and we will now have additional CSRF protections. We could set the value to 'Strict' but that would require an additional EDL challenge under some use conditions.

For item 2, I received assistance in identifying which part of the code was doing this. app/server.ts changes remove the logging of cookie values.

Item 3, this required an audit of the code and I believe things are behaving correctly. We do not rely on EDL for this functionality; rather, we are employing HMAC ourselves ('signed' option) and it ensures an EDL token can't be tampered with and still be valid to Harmony.


app/server.ts
```diff
@@ -24,9 +24,16 @@ import JobReaper from './workers/job-reaper';
  * @param appLogger - Request specific application logger
  */
 function addRequestLogger(appLogger: Logger): RequestHandler {
+  const requestFilter = (req, propName): unknown => {
```
chris-durbin, line 27: Neat.

chris-durbin (approved):
I can't comment on the correctness of this change because I don't completely understand the web and cookies in general. I did test in a sandbox environment that my requests using Insomnia continued to work without issue.

bilts (approved):
Confirmed updating logging behavior is working

PR #34, opened 2021-06-15
HARMONY-851: Move job links to a separate database table

chris-durbin:
I tested out the migrations locally and in a sandbox environment.

I decided to not drop the _json_links column from the existing jobs table just in case something doesn't work correctly when running the migrations in live environments (the sandbox one I tested had 18 existing jobs which all migrated fine).

After we've deployed to all the environments and the jobs listings look fine I can submit a new PR to drop the _json_links column from the jobs table.

I added Postgres to our local k8s deployment. By default you will continue to use sqlite, but if you want to use Postgres you can set up the database by running:
```
DATABASE_URL=postgresql://postgres:<pass>@localhost:5432/postgres NODE_ENV=production npx knex --cwd db migrate:latest
```
Then set DATABASE_URL and NODE_ENV=production in your .env file so that when you bring up harmony it uses Postgres.


vinnyinverso (approved):

app/models/job-link.ts
```diff
@@ -0,0 +1,162 @@
...
+  }
+}
+
+/**
+ * Returns the links for a given job
+ * @param transaction - the transaction to use for querying
+ * @param jobID - the UUID associated with the job
+ */
+export async function getLinksForJob(transaction, jobID): Promise<JobLink[]> {
+  const links = await transaction('job_links').select().where({ jobID }).forUpdate();
```
bilts, line 160: Is `forUpdate()` correct here?

app/models/job-link.ts
```diff
@@ -0,0 +1,162 @@
...
+    return errors.length === 0 ? null : errors;
+  }
+}
+
+/**
+ * Returns the links for a given job
+ * @param transaction - the transaction to use for querying
+ * @param jobID - the UUID associated with the job
+ */
+export async function getLinksForJob(transaction, jobID): Promise<JobLink[]> {
```
bilts, line 159: The next ticket is going to immediately need to support pagination on this.  I think it's okay to defer that, but I wanted to bring it up in case it alters your thinking on the change.

app/models/job-link.ts
bilts, line 160: Will these be fetched in any consistent sequence?  May want to order by created_at or updated_at, though if STAC catalog and cloud access are persisted also (I can't tell) then that wouldn't quite work.

app/models/job.ts
```diff
@@ -399,14 +395,20 @@ export class Job extends Record {
...
-    // Need to jump through serialization hoops due array caveat here: http://knexjs.org/#Schema-json
     const { links, originalStatus } = this;
     delete this.links;
     delete this.originalStatus;
-    this._json_links = JSON.stringify(links);
-    await super.save(transaction);
+    // Remove the _json_links field once we've dropped the column
+    await super.save(transaction, { ...this, _json_links: [] });
+    const promises = [];
+    for (const link of links) {
```
bilts, line 404: This will probably need to change to be a bit more O(1) once we get into pagination and no granule limit, i.e. we saving the link and attaching it to the job is probably going to need to happen on the link object rather than the link array in the job object.

app/models/job.ts
```diff
@@ -439,12 +441,13 @@ export class Job extends Record {
...
   /**
-   * Returns only the links with a rel that matches the passed in value
+   * Returns only the links with a rel that matches the passed in value.
    *
    * @param rel - the relation to return links for
    * @returns the job output links with the given rel
    */
   getRelatedLinks(rel: string): JobLink[] {
-    return this.links.filter((link) => link.rel === rel);
+    const links = this.links.filter((link) => link.rel === rel);
```
bilts, line 450: I can't recall quite how / where this is used, but I'm expecting it to need to be changed to do an actual query once pagination comes into play and we can't just store all of the links in an array.

app/util/date.ts
```diff
@@ -1,10 +1,9 @@
-/* eslint-disable import/prefer-default-export */
 /**
  * Converts a Date object into an ISO String representation (truncates milliseconds)
  *
  * @param date - The date to convert
  * @returns An ISO string representation of the date, with milliseconds truncated
  */
-export function toISODateTime(date: Date): string {
+export default function toISODateTime(date: Date): string {
```
bilts, line 7: Noted above, I probably just disagree with the linter rule in this case.  It makes sense for this to be the default export only if we don't plan to have other date utility methods.

config/local-argo-deployment.yml
```diff
@@ -655,3 +655,67 @@ data:
   # Fake credentials for localstack that Argo uses so it avoids looking up IAM
   accessKey: c2hoaGhoaGg=
   secretKey: aXRzYXNlY3JldA==
+---
+apiVersion: v1
+kind: Secret
+metadata:
```
bilts, line 661: It could just be my inability to read kube config, but I don't _think_ this is used, and it's just hard-coded below on line 704?

db/migrations/20210614142629_split_out_job_links.js
```diff
@@ -0,0 +1,88 @@
...
+  })
+  .then(async () => {
+    // Populate the job_links table
+    const jobs = await knex.select(['jobID', '_json_links']).from('jobs').forUpdate();
+    jobs.forEach(async (job) => {
+      let links = job._json_links;
+      if (typeof links === 'string') {
+        links = JSON.parse(links);
+      }
+      if (links && links.length > 0) {
```
bilts, line 59: Will definitely want to watch this as it deploys to SIT and UAT 😨

db/migrations/20210614142629_split_out_job_links.js
```diff
@@ -0,0 +1,88 @@
...
+        links.forEach(async (link) => {
+          const { href, type, title, rel, temporal, bbox } = link;
+          const temporalStart = temporal && temporal.start ? new Date(temporal.start) : undefined;
+          const temporalEnd = temporal && temporal.end ? new Date(temporal.end) : undefined;
+          const now = new Date();
+          const createdAt = now;
+          const updatedAt = now;
+          const bboxString = bbox ? bbox.join(',') : undefined;
```
bilts, line 67: Am I correct that we are not currently and this change will not be persisting things like the STAC catalog link and cloud access link?  I think that's the best option, but I can't tell from the code that that's not ever happening, because of the intermixing of those links.

app/frontends/stac-item.ts
```diff
@@ -126,11 +129,19 @@ export class HarmonyItem {
    * @param end - Data end datetime
    *
    */
-  addTemporalExtent(start: string, end: string): void {
-    // Validate
-    this.setProperty('start_datetime', start);
-    this.setProperty('end_datetime', end);
-    this.setProperty('datetime', start);
+  addTemporalExtent(start: Date | string, end: Date | string): void {
+    let startString = start as string;
```
indiejames, line 133: This logic is a little confusing to me. I think it might be more straight-forward like this
```
const startString = typeof start === 'string' ? start : start.toISOString();

File: package-lock.json [hunk 10]
```diff
@@ -8853,156 +8868,54 @@
...
-      "engines": {
-        "node": ">=0.10.0"
-      }
-    },
-    "node_modules/get-pkg-repo/node_modules/redent": {
-      "version": "1.0.0",
-      "resolved": "https://registry.npmjs.org/redent/-/redent-1.0.0.tgz",
-      "integrity": "sha1-z5Fqsf1fHxbfsggi3W7H9zDCr94=",
+      "version": "3.3.0",
+      "resolved": "https://registry.npmjs.org/meow/-/meow-3.3.0.tgz",
```
bilts, line 8872: Looks like this got _more_ out of date

package.json
```diff
@@ -89,6 +89,7 @@
     "knex": "^0.20.15",
     "knex-paginate": "^1.2.1",
     "lodash": "^4.17.21",
+    "meow": "^8.0.0",
```
bilts, line 92: I don't _think_ this is necessary in our own package.json, and as you noted it also doesn't fix the problem with dependencies.

app/models/job-link.ts
```diff
@@ -0,0 +1,162 @@
...
+}
+
+interface JobLinkRecord extends BaseJobLink {
+  bbox?: string;
+  temporalStart?: Date;
+  temporalEnd?: Date;
+}
+
+/**
+ * For the constructor we can take either a JobLink or a JobLinkRecord.
```
indiejames, line 24: This seems kind of convoluted.

app/models/job-link.ts
```diff
@@ -0,0 +1,162 @@
...
+ * which has a different representation for the bbox and temporal.
+ * Specifying JobLink | JobLinkRecord as the type causes many Typescript
+ * errors so we use this interface to avoid that.
+ */
+
+export interface JobLinkOrRecord extends BaseJobLink {
+  bbox?: string | number[];
+  temporalStart?: Date;
+  temporalEnd?: Date;
+  temporal?: {
```
indiejames, line 36: Why have this and `temporalStart`/`temporalEnd`?

app/backends/service-response.ts
```diff
@@ -93,7 +93,7 @@ export function updateJobFields(
     } else if (status) {
       job.updateStatus(status as JobStatus);
     } else if (redirect) {
-      job.addLink({ href: redirect, rel: 'data' });
+      job.addLink(new JobLink({ href: redirect, rel: 'data' }));
```
bilts, line 96: Lazy Patrick would have made `addLink` accept the original ad hoc object representation and converted it into a `JobLink` internally.  However, as noted later, I think we're actually going to want to persist the new `JobLink` separate from the overall job once we get to pagination, so this is probably better.

app/models/job-link.ts
```diff
@@ -0,0 +1,162 @@
...
+ * For the constructor we can take either a JobLink or a JobLinkRecord.
+ * When created from the database it will be a JobLinkRecord, but when
+ * constructed outside of the database it will likely be a JobLink
+ * which has a different representation for the bbox and temporal.
+ * Specifying JobLink | JobLinkRecord as the type causes many Typescript
+ * errors so we use this interface to avoid that.
+ */
+
+export interface JobLinkOrRecord extends BaseJobLink {
+  bbox?: string | number[];
```
indiejames, line 33: I think we should just pick one `string` or `number[]` for internal use and convert to the other as needed for responses, etc.

app/models/job-link.ts
```diff
@@ -0,0 +1,162 @@
...
+  };
+
+  bbox?: number[];
+
+  /**
+   * Creates a Job link from the links in a job.
+   *
+   * @param fields - Object containing fields to set on the record
+   */
+  constructor(fields: JobLinkOrRecord) {
```
indiejames, line 88: I think it might have been better to use constructor overloading and then in the implementation use `typeof` to check the type, rather than introducing the `JobLinkOrRecord` class.

indiejames (approved):
Some comments, but probably stuff you don't need to address in this PR.

app/backends/service-response.ts
```diff
@@ -75,15 +76,14 @@ export function updateJobFields(
...
-        link.temporal = { start, end };
+        link.temporal = { start: new Date(temporal[0]), end: new Date(temporal[1]) };
       }
       link.rel = link.rel || 'data';
       job.addLink(link);
     }
     if (progress) {
       if (Number.isNaN(+progress)) {
-        throw new TypeError('Job record is invalid: ["Job progress must be between 0 and 100"]');
+        throw new TypeError('Record is invalid: ["Job progress must be between 0 and 100"]');
```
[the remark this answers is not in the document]
chris-durbin: Job record (original message) is better - the problem was that the message was constructed inside the record class which all database records (now jobs and job links) use. Might need to construct the message in the child class instead.

app/frontends/jobs.ts
```diff
@@ -37,16 +38,16 @@ function getLinksForDisplay(job: Job, urlRoot: string): JobLink[] {
   let { links } = job;
   const dataLinks = job.getRelatedLinks('data');
   if (containsS3DirectAccessLink(job)) {
-    links.unshift(getCloudAccessJsonLink(urlRoot));
-    links.unshift(getCloudAccessShLink(urlRoot));
+    links.unshift(new JobLink(getCloudAccessJsonLink(urlRoot)));
+    links.unshift(new JobLink(getCloudAccessShLink(urlRoot)));
```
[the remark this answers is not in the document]
chris-durbin: I have no problem renaming the table - that wouldn't be bad and I'd definitely rather do that before we merge anything in. But I'm not quite following (it is late in the day for me). Are you suggesting the database table should be called job_results instead of job_links? What else did you want to change?

app/frontends/versions.ts
```diff
@@ -1,7 +1,7 @@
 import { Response } from 'express';
 import { defaultContainerRegistry, ECR } from 'util/container-registry';
 import { Logger } from 'winston';
-import { toISODateTime } from 'util/date';
+import toISODateTime from 'util/date';
```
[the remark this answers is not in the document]
chris-durbin: Oh oops - I was just getting rid of a linter line - I agree and can put that back since it was intentional.

app/models/job-link.ts
```diff
@@ -0,0 +1,162 @@
...
+/**
+ * For the constructor we can take either a JobLink or a JobLinkRecord.
+ * When created from the database it will be a JobLinkRecord, but when
+ * constructed outside of the database it will likely be a JobLink
+ * which has a different representation for the bbox and temporal.
+ * Specifying JobLink | JobLinkRecord as the type causes many Typescript
+ * errors so we use this interface to avoid that.
+ */
+
+export interface JobLinkOrRecord extends BaseJobLink {
```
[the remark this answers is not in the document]
chris-durbin: This is here to help illustrate how leaky our jobs and job links are - I kind of like how TypeScript makes you face your poor decisions and define them.

app/models/job-link.ts
```diff
@@ -0,0 +1,162 @@
...
+  temporalStart?: Date;
+  temporalEnd?: Date;
+}
+
+/**
+ * For the constructor we can take either a JobLink or a JobLinkRecord.
+ * When created from the database it will be a JobLinkRecord, but when
+ * constructed outside of the database it will likely be a JobLink
+ * which has a different representation for the bbox and temporal.
+ * Specifying JobLink | JobLinkRecord as the type causes many Typescript
```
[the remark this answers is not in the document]
chris-durbin: Yes, I spent a little time trying to make things better and went in circles - eventually I had to just timebox it to get something working and hope the next person can make it better with that ticket.

db/migrations/20210614142629_split_out_job_links.js
chris-durbin: Tested it in Sandbox - it looked fine for the 18 jobs there. I still have a bunch of worries - potentially taking awhile with many jobs, memory issues (I think forEach should just process one thing in memory at a time, but we're going to be loading all the jobs from the database at once), not loading all the jobs for some reason, unexpected data, etc. Will definitely need to watch.

db/migrations/20210614142629_split_out_job_links.js
chris-durbin: Those three links (cloud-access.sh, cloud-access.json, STAC link) are not saved in the database from my testing. But we do have one link representing the S3 bucket that does get stored for every single request. I was thinking that link shouldn't be stored in the DB and should just be presented by the frontend.

```
{ href: "s3://local-staging-bucket/public/harmony/service-example/f437c85c-4908-4ab9-8673-08d06e1606d8/",
title: "Results in AWS S3. Access from AWS us-west-2 with keys from /cloud-access.sh" }
```

package.json
chris-durbin: Yeah - already gone.

app/backends/service-response.ts
```diff
@@ -66,7 +67,7 @@ export function updateJobFields(
   const { error, item, status, redirect, progress } = query;
   try {
     if (item) {
-      const link = _.pick(item, ['href', 'type', 'rel', 'title']) as JobLink;
+      const link = new JobLink(_.pick(item, ['href', 'type', 'rel', 'title']));
```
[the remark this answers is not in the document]
chris-durbin: Without it you get:

Argument of type 'ArgoCallbackQueryItem' is not assignable to parameter of type 'JobLinkOrRecord'.
  Types of property 'temporal' are incompatible.
    Type 'string' has no properties in common with type '{ start?: Date; end?: Date; }'.ts(2345)

app/backends/service-response.ts
```diff
@@ -66,7 +67,7 @@ export function updateJobFields(
...
       if (item.bbox) {
```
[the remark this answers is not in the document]
chris-durbin: There's only 8.

app/frontends/stac-item.ts
chris-durbin: Done

app/models/job-link.ts
chris-durbin: +1

app/models/job-link.ts
chris-durbin: This is the point of the interface - you can either pass in the temporal JSON with start and end properties or temporalStart and temporalEnd. In most places we use the temporal JSON, but in the database we are storing temporalStart and temporalEnd as two separate timestamps.

app/models/job-link.ts
chris-durbin: Similar to the temporal and temporalStart/temporalEnd above. That particular interface allows either, but generally we use number[].

app/models/job-link.ts
chris-durbin: That interface was needed in order to compile without type errors.

Can you give me an example of what you're suggesting though (that's what I wanted to do to, but I couldn't figure out how to make it work because Typescript doesn't actually let you define multiple constructors that take different types).

app/models/job-link.ts
chris-durbin: I think we want to lock the table so that it cannot change as we're trying to pull the job links. I'm not sure if it's necessary, but I did see that we did the same thing when querying for jobs so wanted to be consistent.

chris-durbin: I decided to hold off on sorting for the pagination ticket. I assumed it will just default to row insert order so it will always be the same.

app/models/job.ts
chris-durbin: Oh... I think this is mostly used in STAC catalogs... how do they handle pagination?

config/local-argo-deployment.yml
chris-durbin: You're right - it's not doing an `envFrom`.

app/models/job-link.ts
indiejames: It does let you _declare_ multiple constructors. You can only have one implementation, but that can be generic and can do different things depending on the type of the input:
```typescript
constructor(fields: JobLink); 
constructor(fields: Record);
constructor(fields: JobLink | Record) {
    super(fields);
    if (fields instanceof JobLink) {
       // init with JobLink
    } else {
      // init with Record
    }
```

chris-durbin: It has the same issues that caused me to introduce the interface:

```
Property 'temporalStart' does not exist on type 'JobLinkRecord | JobLink'.
  Property 'temporalStart' does not exist on type 'JobLink'.
```

app/backends/argo-response.ts
```diff
@@ -49,7 +50,7 @@ function updateJobFields(
   try {
     if (items && items.length > 0) {
       items.forEach((item, _index) => {
-        const link = _.pick(item, ['href', 'type', 'rel', 'title']) as JobLink;
+        const link = new JobLink(_.pick(item, ['href', 'type', 'rel', 'title']));
```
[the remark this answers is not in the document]
chris-durbin: Yes, otherwise I get the type errors.

app/backends/service-response.ts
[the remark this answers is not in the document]
chris-durbin: Fixed.

app/frontends/versions.ts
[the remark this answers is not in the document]
chris-durbin: Fixed.

app/models/job.ts
bilts: They don't yet. Mostly just an observation that we can't really rely on storing all links in memory or we lose out on the benefits of pagination.  This can wait for those tickets but needs to be addressed there.

bilts (approved):

app/util/date.ts
chris-durbin: Fixed.

PR #35, opened 2021-06-16
HARMONY-865: Make queryVariables use POST not GET

vinnyinverso:


test/ogc-api-coverages/get-coverage-rangeset-shapefile.ts
```diff
@@ -277,7 +278,7 @@ describe('OGC API Coverages - getCoverageRangeset with shapefile', function () {
     describe('when the CMR returns a 4xx', function () {
       const cmrErrorMessage = 'Corrupt GeoJSON';
       const cmrStatus = 400;
-      hookCmr('cmrPostSearchBase', { status: cmrStatus, data: { errors: [cmrErrorMessage] } });
+      hookCmr('queryGranulesForCollection', () => { throw new CmrError(cmrStatus, cmrErrorMessage); });
```
vinnyinverso, line 281: The only reason I made this change is because queryVariables also uses POST, we need queryVariables to be able to call the non-stubbed version of cmrPostSearchBase().

test/helpers/stub-cmr.ts
```diff
@@ -4,16 +4,25 @@ import * as cmr from 'util/cmr';
 
 type CmrMethodName = 'cmrSearchBase' | 'fetchPost' | 'cmrPostSearchBase' | 'getCollectionsByIds' | 'getVariablesByIds' | 'getVariablesForCollection' | 'queryGranulesForCollection' | 'belongsToGroup' | 'cmrApiConfig';
 
+function stubCmr(functionName: CmrMethodName, response: object): void;
```
indiejames, line 7: I think these declaration might now be necessary.

indiejames (approved):
Just one comment.

bilts (commented):
I think it could use a test that actually queries a very large number of variables, otherwise it looks good.

test/helpers/stub-cmr.ts
vinnyinverso: Thanks I just removed them.

vinnyinverso:
Added the test here: test/util/cmr.ts

https://github.com/nasa/harmony/blob/f090bc6fb1bc5f025a3b8a1a106a9b49445788c2/test/util/cmr.ts

test/util/cmr.ts
```diff
@@ -0,0 +1,112 @@
...
+        'V1240763723-YOCLOUD', 'V1240763884-YOCLOUD', 'V1240764039-YOCLOUD',
+        'V1240764017-YOCLOUD', 'V1240763929-YOCLOUD', 'V1240764118-YOCLOUD',
+        'V1240763696-YOCLOUD', 'V1240763784-YOCLOUD', 'V1240763585-YOCLOUD',
+        'V1240764142-YOCLOUD', 'V1240763899-YOCLOUD', 'V1240764007-YOCLOUD',
+        'V1240763704-YOCLOUD', 'V1240763609-YOCLOUD', 'V1240763805-YOCLOUD',
+        'V1240763555-YOCLOUD', 'V1240763903-YOCLOUD', 'V1240763563-YOCLOUD',
+        'V1240764102-YOCLOUD', 'V1240763756-YOCLOUD', 'V1240763551-YOCLOUD',
+        'V1240763832-YOCLOUD',
+      ];
+      const variables = await getVariablesByIds(ids, '');
```
bilts, line 108: If the request fails, would it return something other than an empty list of variables?  (I assume yes, just checking because it's not 100% obvious).  You could include one valid ID in the list to be more accurate.

test/util/cmr.ts
```diff
@@ -0,0 +1,112 @@
+/* eslint-disable max-len */
+import { describe, it } from 'mocha';
+import { expect } from 'chai';
+import { getVariablesByIds } from '../../app/util/cmr';
+
+describe('util/cmr', function () {
+  describe('getVariablesByIds', function () {
+    it('returns a valid response, given a huge number of variables', async function () {
+      const ids = [
+        'V1240763852-YOCLOUD', 'V1240763854-YOCLOUD', 'V1240763844-YOCLOUD',
```
bilts, line 10: Impressive.  Could maybe make it more concise by using a for loop to generate bogus IDs.

bilts (approved):
Minor things.  No changes necessary, but some things to consider.

PR #38, opened 2021-06-24
Harmony 482

indiejames:
Changes to support paging of data links in job status page and removal of links from jobs listing page.


app/models/services/base-service.ts
```diff
@@ -160,7 +160,7 @@ export default abstract class BaseService<ServiceParamType> {
       do {
         // Sleep and poll for completion.  We could also use SNS or similar for a faster response
         await new Promise((resolve) => setTimeout(resolve, env.syncRequestPollIntervalMs));
-        job = await Job.byRequestId(db, requestId);
+        ({ job } = await Job.byRequestId(db, requestId));
```
chris-durbin, line 163: Can you explain this syntax? It looks like you are destructuring and calling the result as function?

env-defaults
```diff
@@ -82,6 +82,12 @@ DEBUG_KNEX=false
 # Log level (error, warn, info, debug)
 LOG_LEVEL=debug
 
+# Default number of results in a page
+DEFAULT_PAGE_SIZE=2000
```
chris-durbin, line 86: We're changing our default page size from 10 to 2000 and max page size from 2000 to 10000?

I think default should stay at 10 for performance and usability reasons. I don't have as strong of an opinion on the change from 2000 to 10000 max.

test/jobs/job-status-pagination.ts
```diff
@@ -0,0 +1,209 @@
...
+    // use a DEFAULT_PAGE size of 10 for these tests
+    ({ defaultPageSize } = env);
+    env.defaultPageSize = 10;
+
+    // Generate some links for the job - need to save the job after each link is added to
+    // circumvent ordering problems with sqlite
+    const jTrx = await db.transaction();
+    await aJob.save(jTrx);
+    jTrx.commit();
+    for (let i = 1; i < 51; i++) {
```
chris-durbin, line 27: I'm not sure how fast these tests run, but consider changing the default page size to say 5 and then 20 links if the test is slightly slow.

test/jobs/job-status-pagination.ts
```diff
@@ -0,0 +1,209 @@
...
+        });
+      });
+    });
+  });
+
+  describe('when `page` parameter and `limit` are not set', function () {
+    hookJobStatus({ jobID, username: 'joe', query: {} });
+    it('shows the first page of results with the default page size', function () {
+      const job = new Job(JSON.parse(this.res.text));
+      const outputLinks = job.getRelatedLinks('data');
```
chris-durbin, line 98: Note that this will need to change for Vinny's 483 PR if he's going to change getRelatedLinks to make a database call.

test/jobs/job-status-pagination.ts
```diff
@@ -0,0 +1,209 @@
...
+      hookJobStatus({ jobID, username: 'joe', query: { page: 2 } });
+      itIncludesPagingRelations(5, `jobs/${jobID}`, { first: null, prev: 1, self: 2, next: 3, last: 5 });
+    });
+
+    describe('on a middle page', function () {
+      hookJobStatus({ jobID, username: 'joe', query: { page: 3 } });
+      itIncludesPagingRelations(5, `jobs/${jobID}`, { first: 1, prev: 2, self: 3, next: 4, last: 5 });
+    });
+
+    describe('on the penultimate page', function () {
```
chris-durbin, line 179: Huh so second to last is what penultimate means?

test/jobs/job-status-pagination.ts
```diff
@@ -0,0 +1,209 @@
...
+      hookJobStatus({ jobID, username: 'joe', query: { page: 4 } });
+      itIncludesPagingRelations(5, `jobs/${jobID}`, { first: 1, prev: 3, self: 4, next: 5, last: null });
+    });
+
+    describe('on the last page', function () {
+      hookJobStatus({ jobID, username: 'joe', query: { page: 5 } });
+      itIncludesPagingRelations(5, `jobs/${jobID}`, { first: 1, prev: 4, self: 5, next: null, last: null });
+    });
+
```
chris-durbin, line 188: Can you add one more test with { page: 6  } to show the behavior? I'm not sure if it should be an error or just an empty set of links though I think I'd lean towards error.

chris-durbin (commented):
Is there any impact to our regression tests or notebooks with the change in behavior to job listings to no longer show data links?

Similarly are there any impacts to the regression tests or notebooks with the job status links having these extra pagination links and potentially not returning all the data links (if we keep the default to 2000 we'll always return all of the links because of our current granule limits, but if we reduce it there might be an impact)?

app/models/services/base-service.ts
indiejames: The parentheses are necessary to prevent the compiler from interpreting the destructuring brackets as a block. This is apparently the only way to use object destructuring of a variable that was previously declared with a `let`. You don't need them if you are declaring the variable at the same time as the destructuring, e.g., 
```
const { job } = await Job.byRequestId(db, requestId);
```
TypeScript weirdness.

env-defaults
indiejames: I'm just following the AC...

test/jobs/job-status-pagination.ts
indiejames: I didn't notice any slowdown.

test/jobs/job-status-pagination.ts
indiejames: He is no longer making that change.

test/jobs/job-status-pagination.ts
indiejames: yes

test/jobs/job-status-pagination.ts
indiejames: I'l add the test. The current behavior for jobs listings is to simply return an empty set. I agree I would think an error would make more sense, but maybe it's easier for people to write code that just keeps paging until it gets empty results than it is to handle an error response.

bilts (approved):
Tested and working, just a few style comments, none critical.

test/jobs/job-status-pagination.ts
indiejames: Done.

app/backends/argo-response.ts
```diff
@@ -109,7 +109,7 @@ export default async function responseHandler(req: Request, res: Response): Prom
 
   const trx = await db.transaction();
 
-  const job = await Job.byRequestId(trx, requestId);
+  const { job } = await Job.byRequestId(trx, requestId);
```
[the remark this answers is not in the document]
indiejames: job.queryAll does not return a list of jobs either. We need to pass the pagination information back to the code generating the page links. I could rename this to something else like `jobAndPaginationInformationByRequestId`, but that seems a bit wordy.

app/models/services/base-service.ts
```diff
@@ -189,7 +189,7 @@ export default abstract class BaseService<ServiceParamType> {
    * received.
    * @param _logger - the logger associated with the request
    */
-  protected abstract async _run(_logger: Logger): Promise<InvocationResult>;
+  protected abstract _run(_logger: Logger): Promise<InvocationResult>;
```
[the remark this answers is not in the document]
indiejames: I didn't remove it because of redundancy, I removed it because I was getting a compiler warning that said it was incorrect to label an abstract method as `async`.

test/jobs/job-status-pagination.ts
chris-durbin: Sounds good. I could see either way. I think CMR does the empty list behavior, so we'll at least be consistent.

app/models/job.ts
```diff
@@ -415,24 +438,26 @@ export class Job extends Record {
...
-        const serializedLink = link.serialize();
-        let { href } = serializedLink;
-        const { title, type, rel, bbox, temporal } = serializedLink;
-        // Leave the S3 output staging location as an S3 link
-        if (rel !== 's3-access') {
-          href = createPublicPermalink(href, urlRoot, type, linkType);
-        }
-        return removeEmptyProperties({ href, title, type, rel, bbox, temporal });
-      }) as unknown as JobLink[];
+      if (linkType !== 'none') {
```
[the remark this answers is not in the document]
indiejames: Fixed. The default behavior is acceptable for `linkType`.

chris-durbin (approved):
Approved with the change to have a separate default of 10 jobs returned with a job listing.

indiejames:
I separated the environment variables for the default number of jobs in the jobs list endpoint from the environment variable used for the default number of links in the job status and STAC catalog pages.

PR #40, opened 2021-06-25
DAS-1097 - Update history_json schema to allow parameters attribute to be an object

owenlittlejohns:
This PR makes a change to the `history_json` v0.1.0 schema to allow the parameters attribute to be a raw object (with unspecified properties) as proposed by David Auty. Because it only extends the options (and so is entirely backwards compatible) it has been proposed that the current semantic version (0.1.0) is overwritten, rather than a new version added.

I'm hopeful that this is the last change to the schema in the short term, and definitely hope it's the last update to v0.1.0.

### DIFF
--- a/app/schemas/history/0.1.0/history-v0.1.0.json
+++ b/app/schemas/history/0.1.0/history-v0.1.0.json
@@ -34,7 +34,7 @@
             },
             "parameters": {
                "description": "The list of parameters to the program when generating this data file",
-               "type": [ "array", "string" ],
+               "type": [ "array", "string", "object" ],
                "items": { "type": "string" }
             },
             "program_ref": {


chris-durbin:
FYI - the failure on the license and security checks seems to be an issue with Snyk, no problems with the file that failed. Jon opened a ticket with Snyk when we saw this same error before.

chris-durbin (approved):

owenlittlejohns:
Thanks, Chris, for the information on Snyk - should I be filing a ticket against Snyk, too?

chris-durbin:
> Thanks, Chris, for the information on Snyk - should I be filing a ticket against Snyk, too?

Nope, we're good.

bilts (approved):

PR #42, opened 2021-06-29
Harmony 483 - Add paging to STAC catalog route

vinnyinverso:


app/frontends/stac.ts
```diff
@@ -60,13 +79,33 @@ export async function getStacCatalog(req, res): Promise<void> {
...
+        job.jobID, job.request, job.links, pagingLinks, linkType,
+      ), pagingParams,
+    );
   } catch (e) {
     req.context.logger.error(e);
-    res.status(500);
-    res.json({
-      code: 'harmony:ServerError',
-      description: `Error: Internal server error trying to retrieve STAC catalog for job ${jobId}` });
+    if (e instanceof RangeError) {
```
chris-durbin, line 93: We generally let our middleware handle creating the appropriate JSON and just throw exceptions that are defined in `app/util/errors.ts`.

E.g.
```
if (e instanceof RangeError) {
  throw new RequestValidationError(e.message)
} ...
```

app/models/job.ts
```diff
@@ -1,5 +1,5 @@
 import { pick } from 'lodash';
-import { IWithPagination } from 'knex-paginate'; // For types only
+import { IPagination, IWithPagination } from 'knex-paginate'; // For types only
```
chris-durbin, line 2: It looks like we should either use `IPagination` or `IWithPagination` everywhere. I assume `IWithPagination` since that is what was here before.

app/models/job.ts
```diff
@@ -407,6 +408,24 @@ export class Job extends Record {
...
+   *
+   * @param transaction - transaction to use for the query
+   * @returns true or false
+   */
+  async hasStacLinks(
+    transaction,
+  ): Promise<boolean> {
+    const links = await transaction('job_links').select()
+      .where({ jobID: this.jobID, rel: 'data' })
+      .whereNotNull('bbox')
```
chris-durbin, line 422: Just want to confirm that for STAC items we only want data links that have bbox, and both a temporal start and end. Is there the concept of an item only having one of temporal start or  end?

test/helpers/jobs.ts
```diff
@@ -73,6 +73,20 @@ export function areJobLinksEqual(jobLinks: JobLink[], serializedJobLinks: JobLin
...
 }
 
+/**
+ * Compare link titles from a saved Job to those of the STAC catalog output
+ * @param jobLinks - links from a saved job
+ * @param stacLinks - links from STAC catalog
+ * @returns true if the links are the same, false otherwise
+ */
+export function areStacJobLinksEqual(jobLinks: JobLink[], stacLinks: JobLink[]): boolean {
+  const cleanedStacLinks = stacLinks.map((link) => _.pick(link, ['title']));
```
chris-durbin, line 83: Shouldn't we compare all the STAC link fields like temporal and bbox fields as well?

test/helpers/stac.ts
```diff
@@ -12,9 +12,8 @@ import { auth } from './auth';
 export function stacCatalog(
   app: Express.Application,
   jobId: string,
-  linkType?: string,
+  query: object = {},
```
chris-durbin, line 15: Document the parameter.

chris-durbin: Wondering if you should make the same change to line 34 for `stacItem`.

test/helpers/stac.ts
```diff
@@ -45,13 +44,13 @@ export function stacItem(
 export function hookStacCatalog(
   jobId: string,
   username?: string,
-  linkType?: string,
+  query: object = {},
```
chris-durbin, line 47: Document the parameter.

test/stac/stac-pagination.ts
```diff
@@ -0,0 +1,236 @@
...
+    for (let i = 1; i < 51; i++) {
+      links.push(
+        new JobLink({
+          href: `http://example.com/${i}`,
+          title: `Example ${i}`,
+          type: i % 2 === 0 ? 'text/plain' : 'text/ornate',
+          rel: 'data',
+          bbox: '-180,-90,180,90',
+          temporal: {
+            start: new Date(1577836800000),
```
chris-durbin, line 46: We should use a string when constructing a date rather than ms since epoch.

test/stac/stac.ts
```diff
@@ -144,17 +145,22 @@ describe('STAC catalog route', function () {
           expect(catalog.description).to.equal('Harmony output for http://example.com/harmony?job=completedJob');
           expect(catalog.id).to.equal(completedJob.requestId);
           expect(catalog.links).to.eql([
-            { href: '.', rel: 'self', title: 'self' },
             { href: '.', rel: 'root', title: 'root' },
             { href: './0', rel: 'item' },
+            {
+              href: `http://127.0.0.1:4000/stac/${completedJob.requestId}?page=1&limit=${env.defaultResultPageSize}`,
```
chris-durbin, line 151: This is relying on the environment. We should either construct from envs or use a regex when comparing the href. It looks like trying to use a regex might be a pain. I'm not sure if we have a frontendHost env var - can you use `env.backendHost` and env.frontendPort`?

app/frontends/stac.ts
vinnyinverso: Ah OK. Which middleware handler in particular? I was just trying to follow what the other route handler was doing (getStacItem): https://github.com/nasa/harmony/blob/main/app/routers/router.ts#L206 There were a lot of tests written that expect these custom responses.

app/models/job.ts
vinnyinverso: Typescript wouldn't let me pass the pagination object to getPagingLinks without using `IPagination`. I'm not sure why James's code didn't complain when the type was `IWithPagination` but it may have to do with how he wrote his variable assignment?

```
let pagination;
      if (!req.context.isAdminAccess) {
        ({
          job,
          pagination,
        } = await Job.byUsernameAndRequestId(tx, req.user, jobID, false, page, limit));
      } else {
        ({ job, pagination } = await Job.byRequestId(tx, jobID, page, limit));
      }
    });
    if (job) {
      const urlRoot = getRequestRoot(req);
      const pagingLinks = getPagingLinks(req, pagination).map((link) => new JobLink(link));
```

app/models/job.ts
vinnyinverso: I think this implies that both should be required if using an interval?

```If there's clearly no meaningful single 'nominal' time, it is allowed to use null instead. In this case it is required to specify a temporal interval with the fields start_datetime and end_datetime from common metadata. For example, if your data is a time-series that covers 100 years, it's not very meaningful to set the datetime to a single timestamp as it would not be found in most searches that searches for a decade of data in that period although the Item actually covers the decade. ```

Comment (vinnyinverso) [reply app/models/job.ts:422]: https://github.com/radiantearth/stac-spec/blob/master/item-spec/item-spec.md

File: test/helpers/jobs.ts (hunk 4 above)
Comment (vinnyinverso) [reply test/helpers/jobs.ts:83]: From my understanding, for the STAC _catalog_, temporal and bbox fields aren't included in the output links. HREF also won't ever be equal to the original data links HREF. And the rel for STAC links in a catalog are 'item', so they will never be equal to 'data'. Title was the only comparable property I could find.

e.g. STAC catalog:

```
{
  "id": "faf56600-98e4-45d6-871e-0a26ae0a3651",
  "stac_version": "0.9.0",
  "title": "Harmony output for faf56600-98e4-45d6-871e-0a26ae0a3651",
  "description": "Harmony output for http://localhost:3000/C1233800302-EEDTEST/ogc-api-coverages/1.0.0/collections/all/coverage/rangeset?granuleId=G1233800343-EEDTEST&forceAsync=true",
  "links": [
    {
      "href": ".",
      "rel": "root",
      "title": "root"
    },
    {
      "href": "./0",
      "rel": "item",
      "title": "001_00_7f00ff_global.nc.tif"
    },
    {
      "title": "The current page",
      "href": "http://localhost:3000/stac/faf56600-98e4-45d6-871e-0a26ae0a3651?limit=500&page=1",
      "rel": "self",
      "type": "application/json"
    }
  ]
}
```

File: test/stac/stac.ts (hunk 8 above)
Comment (vinnyinverso) [reply test/stac/stac.ts]: Is this better? https://github.com/nasa/harmony/pull/42/commits/e2352b8d2f866b19d58297d1332c65be39582e34

File: app/models/job.ts (hunk 2 above)
Comment (vinnyinverso) [reply app/models/job.ts]: https://github.com/nasa/harmony/pull/42/commits/1b6f5b9c9b3b3053c02904160ec4c06b868ee656

Here's the new change. I can revert it if you have other thoughts.

File: app/frontends/stac.ts (hunk 1 above)
Comment (vinnyinverso) [reply app/frontends/stac.ts:93]: (`getStacCatalog` is a middleware function too).

https://github.com/nasa/harmony/blob/2f38b7f5beb5254f43af8a7265627cd0d20c9cb9/app/routers/router.ts#L205

I might be misunderstanding what you're saying.

Comment (chris-durbin) [reply app/frontends/stac.ts:93]: The middleware that formats the error messages is `app/middleware/error-handler.ts`. All other code should just throw the correct exception type with an appropriate message and the error handler will handle the presentation to the end user.

It does create messages with a 'code' and a 'description' when returning json errors. Given that the existing STAC code wasn't throwing exceptions to be handled by the error handler middleware it's probably out of scope for this ticket to refactor.

File: app/models/job.ts (hunk 2 above)
Comment (chris-durbin) [reply app/models/job.ts]: I just do what the Typescript gods tell me to do - this looks fine to me.

**Review by chris-durbin — APPROVED** [review]

File: app/frontends/jobs.ts [hunk 9]
```diff
@@ -174,7 +174,7 @@ export async function getJobStatus(
         ({
           job,
           pagination,
-        } = await Job.byUsernameAndRequestId(tx, req.user, jobID, true, page, limit));
+        } = await Job.byUsernameAndRequestId(tx, req.user, jobID, false, page, limit));
```
chris-durbin, line 177: I misread this - isn't this going to make it so that links are not returned by the Job status endpoint? Maybe I'm missing some other change.

vinnyinverso: I made a change to the function signature (swapping out the old boolean param for another). The old param was no longer needed because everything calling byUsernameAndRequestId currently requests links. I changed the boolean to signify whether you want stacOnly links, which for the job status code is false.

https://github.com/nasa/harmony/blob/1b6f5b9c9b3b3053c02904160ec4c06b868ee656/app/models/job-link.ts#L182

app/frontends/jobs.ts
```diff
@@ -174,7 +174,7 @@ export async function getJobStatus(
         ({
           job,
           pagination,
-        } = await Job.byUsernameAndRequestId(tx, req.user, jobID, true, page, limit));
```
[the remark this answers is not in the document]
vinnyinverso: See this comment: https://github.com/nasa/harmony/pull/42#discussion_r661450020

app/frontends/jobs.ts
vinnyinverso: Here is the change actually. https://github.com/nasa/harmony/pull/42/files#diff-cdf5c9e86f11ce00f32337e6562e6bbd9dfa25d732fa892573c8335bb04fa046L224-L225

app/models/job.ts
```diff
@@ -407,6 +408,24 @@ export class Job extends Record {
     return terminalStates.includes(this.status);
   }
 
+  /**
+   * Check if the job has any STAC data links
+   *
+   * @param transaction - transaction to use for the query
+   * @returns true or false
+   */
+  async hasStacLinks(
```
bilts, line 417: Similar comments to `getLinksForJob`, I think this bleeding dependencies on presentation into the model a little too much.

If you implement the changes suggested there, I'd carry them over here as params.  `hasLinks(rel=null, requireSpatioTemporal=false)` or similar.  Then the method body could and probably should just call `getLinksForJob` setting `perPage=1`, which amounts to the same query but means that the query logic exists in one method rather than 2, making it easier to update in the future.

bilts (changes requested):
Mostly good.  A couple of minor things to address.

app/models/job.ts
vinnyinverso: Agreed, thanks for the suggestions. I'll probably break it into multiple params rather than adding a new method.

app/frontends/stac.ts
```diff
@@ -81,11 +122,19 @@ export async function getStacItem(req, res): Promise<void> {
...
+    if (itemIndexInt === undefined) {
+      throw new RequestValidationError('STAC item index should be a valid integer');
+    }
+    const pagingParams: PagingParams = { page: itemIndexInt + 1, limit: 1 };
     await handleStacRequest(
       req,
       res,
-      (data) => stacItemCreate.apply(null, [data, itemIndex, linkType]),
+      (job: Job) => stacItemCreate.apply(
+        null, [job.jobID, job.request, job.links[0], itemIndexInt, linkType, job.createdAt],
```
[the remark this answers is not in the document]
vinnyinverso: handleStackRequest, which calls this callback method `(job: Job) => stacItemCreate.apply(...)`: 

1. validates that the STAC links requested by the user actually exist before invoking the callback 
2. loads only the requested links, (one of them in this case), into job.links via Job.byUsernameAndRequestId()
 
Assuming handleStackRequest doesn't change too much, you can assume that the requested STAC link does exist, and only one was loaded into job.links (since that's all that we need).

app/frontends/stac.ts
vinnyinverso: I'll refactor it.

app/frontends/stac-catalog.ts
```diff
@@ -88,29 +86,35 @@ class HarmonyCatalog implements SerializableCatalog {
...
+ * @param stacDataLinks - JobLink's to add to catalog
+ * @param pagingLinks - links pointing to the next, previous, and current page
  * @param linkType - the type of data links that the stac-items should use
  *
  * @returns - STAC Catalog JSON
  *
  * @example
  * const catalog = require('HarmonyCatalog');
- * let jsonObj = catalog.create(job);
+ * let jsonObj = catalog.create(job.jobID, job.request, stacDataLinks, pagingLinks);
```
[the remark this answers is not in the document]
vinnyinverso: I can't remember perfectly because there were a number of constraints I was thinking about early on (that I thought this satisfied well), but I'll try my best to explain, first by explaining a separate but similar change.

stacItem.create was previously indexing into an array of job.links based on the index requested by the user and generating the STAC item based on that link. Since we're no longer loading all job links into memory, it's no longer safe to assume in stacItem.create that job.links[index] has the link that we want. So that change decouples the structure/ordering of job.links from the catalog.create logic.

Once I went down that road, it made sense for consistency to have catalog.create also force the caller to specify up front which links they want turned into STAC items/links rather than relying directly on a job. 

Happy to revert if you disagree. I could definitely see how the simplicity of the former might be cleaner.

bilts (approved):

PR #44, opened 2021-07-08
WIP: Harmony-843:  Avoid request re-submit running additional jobs

rwat:
### Purpose:

Current WIP as unit tests and database migration still need to be done.

The purpose of this branch is to avoid re-submitting a Harmony job in the event a synchronous request times out client-side and the HTTP client re-submits (not seen in a dev environment but happens in a cloud environment).

The solution implemented does the following: when a request is submitted to harmony, it will check the request URL to determine if it's the same as another request which is already running ( JobStatus.ACCPTED || JobStatus.RUNNING ) and, if so, "attaches" the request to the already running job and waits for that job's results. In that case, the results from one job would be returned to more than one request. If there is a shapefile present then the logic goes one step further and compares an MD5 hash of shapefile content to know if two or more requests were the same. This is implemented for both sync and async requests and is transparent to the end-user.

Regarding the need for implementing our own MD5 hash for S3 objects: it allows the hash to be reliably available regardless of how the object was uploaded: https://docs.aws.amazon.com/AmazonS3/latest/API/RESTCommonResponseHeaders.html

### For review:

At this time I'd like a review of the approach and code and I'll continue working on unit tests and knex db migration.

Example longer running request:
http://localhost:3000/C1234088182-EEDTEST/ogc-api-coverages/1.0.0/collections/all/coverage/rangeset?subset=lat(52%3A77)&subset=lon(-165%3A-140)

Example request with shapefile:
curl -Ln -bj  "http://localhost:3000/C1233800302-EEDTEST/ogc-api-coverages/1.0.0/collections/red_var/coverage/rangeset" -F "shapefile=@test/resources/southern_africa.geojson;type=application/geo+json"  -F "subset=lon(17:98)"


app/models/job.ts
```diff
@@ -428,6 +442,51 @@ export class Job extends Record {
...
+      const originalId = this.requestId;
+      // if a shapefile is not present then database query returned all matching running requests
+      if (this.shapeFileUrl === '') {
+        const assumedId = rows.shift()['requestId'];
+        this.attachedStatus = { didAttach: true, originalId, assumedId };
+      // otherwise, compare checksum of current shapefile to other matching requests shapefile
+      } else {
+        const store = defaultObjectStore();
+        const thisSum = await objectStoreChecksum(store, this.shapeFileUrl);
+        let otherSum = '';
```
indiejames, line 469: I don't think this needs to be declared here, it can be a `const` in your `for` loop.

app/models/job.ts
```diff
@@ -428,6 +442,51 @@ export class Job extends Record {
...
+        let otherId = '';
+        for (const row of rows) {
```
indiejames, line 471: Nit: this might be a little more concise (and functional, if that matters) if you used `Array.find` here instead of a `for` loop.

app/models/services/base-service.ts
```diff
@@ -116,6 +116,12 @@ export default abstract class BaseService<ServiceParamType> {
       const startTime = new Date().getTime();
       logger.info('timing.save-job-to-database.start');
       job = await this._createJob(logger, requestUrl, this.operation.stagingLocation);
+      await job.maybeAttach(db);
+      if (job.attachedStatus.didAttach) {
+        const { originalId, assumedId } = job.attachedStatus;
+        logger.info(`This job attached to a previous job: ${originalId} is now ${assumedId}.`);
+        this.operation.requestId = assumedId;
+      }
```
indiejames, line 124: It might be simpler to just return a promise at this point that either resolves to a redirect (if `isAsync` === true) or calls `this_waitForSyncResponse` rather than saving the job that we know is attached and adding the 'mabyeRun' construct below.

db/db.sql
```diff
@@ -15,7 +15,8 @@ CREATE TABLE `jobs` (
   `updatedAt` datetime not null,
   `request` varchar(4096) not null default 'unknown',
   `isAsync` boolean,
-  `numInputGranules` integer not null default 0);
+  `numInputGranules` integer not null default 0,
+  `shapeFileUrl` varchar(4096) not null);
```
indiejames, line 19: Does it make sense to pre-compute the shapefile md5 and just store that instead of the url?

rwat: You're right - I'll get that changed.

app/models/job.ts
```diff
@@ -428,6 +442,51 @@ export class Job extends Record {
...
+   *
+   * @param transaction - The transaction to use for saving the job
+   * @throws {@link Error} if the job is invalid
+   */
+  async maybeAttach(transaction: Transaction): Promise<void> {
+    const rows = await transaction('jobs')
+      .select('requestId', 'shapeFileUrl')
+      .where({ username: this.username, request: this.request })
+      .andWhere('status', JobStatus.ACCEPTED)
+      .orWhere('status', JobStatus.RUNNING);
```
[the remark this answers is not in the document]
rwat: Good call - I'll get that changed.

app/models/job.ts
rwat: I went with a `for` loop instead of `Array.find` because I wanted the ability to `break` early / not process the whole sequence necessarily.

app/models/job.ts
rwat: It looks like `const`'s and `let`'s are block scoped so declaring outside of the `for` is needed here because the variable is used after that block. Also explains why a `const` can be used inside of a `for` loop: a variable declared this way falls out of scope and doesn't violate `const`'s prevention of variable reassignment.

app/models/job.ts
indiejames: Array.find stops after the first success

app/models/job.ts
indiejames: It doesn't look like `otherSum` is used outside the `for` loop, unless I'm misreading something.

app/models/job.ts
rwat: This code has been changed too: db rows object is `shift()`ed to if one or more matches are found.

db/db.sql
rwat: This is now done.

.eslintrc.yml
```diff
@@ -41,6 +41,7 @@ rules:
   no-await-in-loop: ["off"]
   object-curly-newline: ["off"]
   no-plusplus: ["off"]
+  dot-notation: ["off"]
```
chris-durbin, line 44: I think we do prefer dot notation - not sure if we should turn this off.

.mocharc.yml
```diff
@@ -14,3 +14,4 @@ require:
   - tsconfig-paths/register
 recursive: true
 exit: true
+timeout: 3000
```
chris-durbin, line 17: Curious why you needed to increase the test timeout. New tests, or just in general some tests were timing out on you?

app/middleware/shapefile-converter.ts
```diff
@@ -13,6 +13,7 @@ import { RequestValidationError, HttpError, ServerError } from '../util/errors';
 import { defaultObjectStore } from '../util/object-store';
 import { listToText } from '../util/string';
 import { cookieOptions } from '../util/cookies';
+import fileCheckSum from '../util/file';
```
chris-durbin, line 16: nit: `fileChecksum`

app/models/job.ts
```diff
@@ -526,6 +539,30 @@ export class Job extends Record implements JobRecord {
...
+   * current job's requestId with that of the running job's. Add an attachedStatus property for
+   * later referencing.
+   *
+   * @param transaction - The transaction to use for saving the job
+   * @throws {@link Error} if the job is invalid
+   */
+  async maybeAttach(transaction: Transaction): Promise<void> {
+    const rows = await transaction('jobs')
+      .select('requestId', 'shapeFileHash')
+      .where({ username: this.username, request: this.request, shapefileHash: this.shapeFileHash })
```
chris-durbin, line 553: Checking for `this.request` should work for GET requests, but I'm worried that for POST requests we're going to consider different requests as the same request because we're not tracking the form fields.

app/models/services/base-service.ts
```diff
@@ -168,17 +168,35 @@ export default abstract class BaseService<ServiceParamType> {
...
+    } catch (e) {
+      logger.error(e.stack);
+      throw new ServerError('Failed to save job to database.');
+    }
 
     const { isAsync, requestId } = job;
     this.operation.callback = `${env.callbackUrlRoot}/service/${requestId}`;
+    // If the current job attached to a running job then skip execution in the backend
+    this['maybeRun'] = job.attachedStatus.didAttach
```
chris-durbin, line 195: It looks like this might be the reason you changed the linter settings? I'm not entirely sure what it is going on here, but i don't think you need to set a variable on `this`.

Can't you just have:

```
const runFn = job.attachedStatus.didAttach...
 ...
runFn(logger).then...
 ```

File: db/db.sql [hunk 11]
```diff
@@ -14,8 +14,8 @@ CREATE TABLE `jobs` (
   `request` varchar(4096) not null default 'unknown',
   `isAsync` boolean,
   `numInputGranules` integer not null default 0,
-  `collectionIds` text not null
-);
+  `collectionIds` text not null,
+  `shapeFileHash` varchar(255));
```
chris-durbin, line 18: Not sure if we want to set it to `varchar(32)` since an md5 checksum should be 32 characters and if not set it will be an empty string.

chris-durbin (changes requested):
I tested out with GET requests for both Argo and turbo workflows and the behavior worked as expected attaching to running jobs when I submitted identical requests.

We need a knex database migration for postgres, and I believe we'll need some updates to ensure that POST requests are not always considered identical. I think I'd be ok with just always considering POST requests new requests - browsers are going to be sending GETs which is the most likely cause of these duplicate requests.

PR #48, opened 2021-07-13
Harmony 875

indiejames:
Adds support for using UMM-C/UMM-S associations to connect collections to services instead of declaring them directly in services.yml.


app/middleware/service-selection.ts
```diff
@@ -1,7 +1,35 @@
...
+function addCollectionsToServicesByAssociation(req: HarmonyRequest): ServiceConfig<unknown>[] {
+  const configs = getServiceConfigs();
+  const { collections } = req;
+  if (collections) {
+    for (const coll of collections) {
+      if (coll.associations?.services) {
+        for (const serviceId of coll.associations?.services) {
+          for (const config of configs) {
+            if (config.umm_s?.includes(serviceId)
+              && config.collections
```
vinnyinverso, line 20: Is it ever possible that the service would not have a collections property (i.e. not specified in the YAML file), but _would_ have a umm_s property, in which case, the collections associated with the umm_s record(s) wouldn't get added to the service config here?

indiejames: Argh - I made a commit that changed all this, but forgot to push it. Take a look at the revised code.

vinnyinverso: Ah cool -- I see the added logic.

app/middleware/service-selection.ts
```diff
@@ -1,7 +1,41 @@
...
-import { chooseServiceConfig } from '../models/services';
+import { chooseServiceConfig, getServiceConfigs } from '../models/services';
+
+/**
+ * Add collections to service configs
+ * @param req - The client request
+ * @returns An array of service configurations
+ */
+function addCollectionsToServicesByAssociation(req: HarmonyRequest): ServiceConfig<unknown>[] {
+  const configs = getServiceConfigs();
```
vinnyinverso, line 12: Would it be useful to push the logic of adding these collections to the configs into getServiceConfigs, e.g. getServiceConfigs(req.collections), where the collections are an optional parameter? 

I'm envisioning some other use cases that might want this same kind of behavior. E.g. if there's ever a web GUI component of Harmony, someone might want to look up which services are available with their collection(s) without actually making a data transformation request. 

What do you think?

config/services.yml
```diff
@@ -411,8 +411,12 @@ https://cmr.uat.earthdata.nasa.gov:
         env:
           <<: *default-argo-env
           STAGING_PATH: public/sds/swot-reproject
+    umm-s:
+      - S1237974711-EEDTEST
     collections:
-      - C1233860183-EEDTEST
+      # DO NOT UNCOMMENT THE FOLLOWING LINE - this collection is added via UMM-S and is left here
```
vinnyinverso, line 417: I wonder if it'd be desirable to have our tests rely on a separate services.yml.

vinnyinverso (approved):

config/services.yml
indiejames: I started doing that, but that meant making some extensive changes just to support testing. I also thought it might be good to have some mention in the services.yml that these collections don't need to be there. If anyone has a strong preference I could go the other route.

app/middleware/service-selection.ts
indiejames: That's an interesting idea. I can see potential utility, but I also like the idea of keeping it as a simple getter. I'm going to leave it as is for now.

config/services.yml
vinnyinverso: That sounds like enough work to be its own ticket (if we did go that way).

indiejames:
Closing this PR for now as I have discovered an issue when testing locally.

indiejames:
Re-opening after fixing typo in services.yml. To test, run the following curl after launching from this branch:

```
curl -Ln -bj "http://localhost:3000/C1233860183-EEDTEST/ogc-api-coverages/1.0.0/collections/all/coverage/rangeset?outputCrs=%2Bproj%3Dlcc%20%2Blat_1%3D43%20%2Blat_2%3D62%20%2Blat_0%3D30%20%2Blon_0%3D10%20%2Bx_0%3D0%20%2By_0%3D0%20%2Bellps%3Dintl%20%2Bunits%3Dm%20no_defs&interpolation=near&scaleExtent=-7000000,1000000,8000000,8000000&maxResults=10"
```

This should run swot-repr even though collection C1233860183-EEDTEST is not directly listed in the services.yml `collections` entry for the swot-repr config.

vinnyinverso:
Should there be a test for a service with multiple UMM-S records or is it not necessary?

From the AC: "Harmony's service configuration (services.yml) allows service workflows to list **one or more** UMM-S records that describe them"

chris-durbin (approved):
Verified locally that a collection using UMM-S for association works correctly:

http://localhost:3000/C1233800302-EEDTEST/ogc-api-coverages/1.0.0/collections/blue_var/coverage/rangeset?subset=lat(20%3A60)&subset=lon(-140%3A-50)&granuleId=G1233800343-EEDTEST&outputCrs=EPSG%3A31975&format=image%2Fpng

app/middleware/service-selection.ts
```diff
@@ -1,7 +1,41 @@
...
+  const collections = req.collections || [];
+  // for every collection in the request...
+  for (const coll of collections) {
+    const services = coll.associations?.services || [];
+    // for every service associated with the current collection...
+    for (const serviceId of services) {
+      for (const config of configs) {
+        // if the service config contains a 'umm_s' entry that includes the current service id
+        if (config.umm_s?.includes(serviceId)) {
+          if (config.collections) {
```
bilts, line 22: This method gets about 8 levels deep in nesting.  I'm wondering if there are ways to simplify for clarity.  It could get a little easier if we ensured config.collections was an array (possibly empty) either externally or at the start of the block with an `if (!config.collections) { config.collections = []; }` and then another clause that's `if (config.umm_s?.includes(serviceId) && !config.collections.includes(coll.id))` or just dropping the check for duplication entirely?

config/services.yml
bilts: alternatively, you could mock `getServiceConfigs` to return whatever you want

test/models/services.ts
```diff
@@ -472,3 +475,36 @@ describe('granule limits', function () {
...
+      hookRangesetRequest(version, conversionCollection, 'all', { headers, query: granulQuery });
+      it('uses the backend service from the association', function () {
+        expect(this.service.name).to.equal('harmony/netcdf-to-zarr');
+      });
+    });
+
+    describe('when matching services are provided directly and through associations', function () {
+      StubService.hook({ params: { redirect: 'http://example.com' } });
+      hookRangesetRequest(version, reprojectCollection, 'all', { headers, query: reprojQuery });
+      it('it uses the first matching service', function () {
```
bilts, line 505: Double "it"

bilts (approved):
Minor style comments, but otherwise good

PR #51, opened 2021-07-19
Harmony 880/881: Sharing STAC / job results

vinnyinverso:


vinnyinverso:
This is ready for review. Two questions remaining in my mind are:

1 - The request for the CMR permissions map is made with the accessToken of the user who originated the request. Will they always have permission to query the CMR ACL endpoints? We need them to have permission to do this or else we won't be able to check if guests can read the CMR collections in question.

2 - I made collectionIds required on the Job class but not the JobRecord interface. I want it to be as restrictive as possible rather than making both optional, but I'm not fully sure what the implications of this are for any associated logic. I may need to look into that more.

app/models/job.ts
```diff
@@ -443,7 +507,10 @@ export class Job extends Record {
     this.validateStatus();
     this.message = truncateString(this.message, 4096);
     this.request = truncateString(this.request, 4096);
-    const { links, originalStatus } = this;
+    const { links, originalStatus, collectionIds } = this;
+    // eslint-disable-next-line @typescript-eslint/ban-ts-ignore
```
vinnyinverso, line 511: This is not pretty but it's the best solution I could find, without resorting to using a bunch or '|' in the Job classes/interfaces.

chris-durbin:
> 1 - The request for the CMR permissions map is made with the accessToken of the user who originated the request. Will they always have permission to query the CMR ACL endpoints? We need them to have permission to do this or else we won't be able to check if guests can read the CMR collections in question.

Guests and registered users have access to the ACL endpoint and anyone is allowed to check permissions for guest users, so you should be fine.

app/models/job.ts
vinnyinverso: Using '|' might be better?

test/jobs/job-sharing.ts
```diff
@@ -0,0 +1,225 @@
...
+      });
+      describe('Accessing the STAC Item page', function () {
+        hookStacItem(jobIDWithEULANonexistent, 0, 'jill');
+        it('returns a 404 response', function () {
+          expect(this.res.statusCode).to.equal(404);
+        });
+      });
+    });
+
+    describe('No collections are used in job', function () {
```
vinnyinverso, line 181: I guess for backward compatibility (where we haven't saved a value to the collectionIds column when the job was originally kicked off) this should actually return a 404 when someone other than the owner tries to access the job results?

app/models/services/no-op-service.ts
```diff
@@ -57,6 +57,7 @@ export default class NoOpService extends BaseService<void> {
       createdAt: now,
       updatedAt: now,
       message: this.message,
+      collectionIds: this.operation.collectionIds,
```
chris-durbin, line 60: I'm not sure if we want to display the collection IDs as part of the job status response.

Counterpoint - we have been adding fields that aren't part of the OGC processing spec and seeing the collections used for a job does seem like useful information, so maybe we keep this.

app/util/cmr.ts
```diff
@@ -26,6 +26,21 @@ const acceptJsonHeader = {
   Accept: 'application/json',
 };
 
+export enum CmrPermission {
+  Read = 'read',
```
chris-durbin, line 30: TIL PascalCase for enums.

app/util/cmr.ts
```diff
@@ -26,6 +26,21 @@ const acceptJsonHeader = {
...
+  Update = 'update',
+  Delete = 'delete',
+  Order = 'order',
+}
+
+export interface CmrPermissionsMap {
+  [key: string]: CmrPermission[];
+}
+
+export interface CmrTags {
```
chris-durbin, line 40: Why are we looking at tags?

test/jobs/job-sharing.ts
vinnyinverso: https://github.com/nasa/harmony/pull/51/commits/8f3cccdb9a8c05ff14f037af015a93af741279c8

app/util/cmr.ts
```diff
@@ -91,6 +107,8 @@ export interface CmrQuery
   concept_id?: string | string[];
   page_size?: number;
   downloadable?: boolean;
+  user_id?: string;
```
chris-durbin, line 110: These seem specific to an ACL Query - I think I'd add another interface that extends CmrQuery called something like CmrAclQuery that adds those fields and is only used in the ACL query.

app/util/cmr.ts
```diff
@@ -348,15 +370,22 @@ async function queryGranuleUsingMultipartForm(
...
+  ids: Array<string>,
+  token: string,
+  includeTags?: string,
 ): Promise<Array<CmrCollection>> {
-  return queryCollections({
-    concept_id: ids,
-    page_size: 2000,
-  }, token);
+  const query = {
+    ...(includeTags && { include_tags: includeTags }),
```
chris-durbin, line 382: That's an interesting trick.

test/jobs/job-sharing.ts
```diff
@@ -0,0 +1,225 @@
...
+      start: new Date('2021-01-01T00:00:00.000Z'),
+      end: new Date('2021-01-01T01:00:00.000Z'),
+    },
+  }],
+  request: 'http://example.com/harmony?job=completedJob',
+  status: JobStatus.SUCCESSFUL,
+};
+
+const jobWithNoCollections = buildJob({
+  username: 'joe',
```
chris-durbin, line 41: Maybe define:
```
const jobOwner = 'joe';
const notJobOwner = 'jill';
```
and then use those constants in all the tests so you can tell from just looking at a single line the tests are trying to access a job that isn't theirs?

Only commenting because when I saw 'jill' in the tests I wasn't clear whether or not they owned the job.

By the way these tests were really easy to follow otherwise - nice job with creating all the jobs upfront with easy to understand names.

test/jobs/job-sharing.ts
chris-durbin: I think you are right - we should change it so if we can't tell what the collections for the job are - only the owner has access to it. Hopefully an easy change to use `await Job.byUsernameAndRequestId` in that case instead of `await Job.byRequestId`.

app/models/job.ts
chris-durbin: I went with a different strategy for JobLink (feel free to ignore this):

I added an interface `JobLinkOrRecord` that allowed for multiple types and then I performed the transformations in the `save` call and the constructor. That way I wasn't mucking around with setting things on `this` and then setting them back.

I don't entirely like that approach either.

app/models/job.ts
```diff
@@ -408,6 +416,62 @@ export class Job extends Record {
...
+   * @param accessToken - the token to make the request with
+   * @returns true or false
+   */
+  async collectionsHaveEulaRestriction(accessToken: string): Promise<boolean> {
+    const cmrCollections = await getCollectionsByIds(this.collectionIds, accessToken, 'harmony.has-eula');
+    if (cmrCollections.length !== this.collectionIds.length) {
+      return true;
+    }
+    return !cmrCollections.every((collection) => (collection.tags
+      && collection.tags['harmony.has-eula'].data === false));
```
chris-durbin, line 432: I see the tests for collection has no `harmony.has-eula` tag set so it looks like this line does default to not allowing access unless the tag is explicitly set:

```
collection.tags['harmony.has-eula'].data === false
```

chris-durbin (commented):
Adding my comments now - things look good to me. I'll help test out the migration locally and do some testing before approving.

One question I have is who is tagging these collections? Are we asking the provider to tag their collections? I think we need to update some documentation to explain how that will work and who is responsible.

app/models/job.ts
```diff
@@ -408,6 +416,62 @@ export class Job extends Record {
...
+  /**
+   * Checks whether sharing of this job is restricted by any EULAs for
+   * any collection used by this job.
+   * Defaults to true if any collection does not have the harmony.has-eula tag
+   * associated with it.
+   * @param accessToken - the token to make the request with
+   * @returns true or false
+   */
+  async collectionsHaveEulaRestriction(accessToken: string): Promise<boolean> {
+    const cmrCollections = await getCollectionsByIds(this.collectionIds, accessToken, 'harmony.has-eula');
```
chris-durbin, line 427: We should pull out `harmony.has-eula` to a named constant.

Here's some examples of other groups tags (not sure if we might want to structure things so we have a single JSON tag like). There are pros and cons to having a single harmony tag vs. potentially many. A single tag makes it easy to pull back all harmony information, but would require updates to the tags to include all of the harmony info:
```
harmony.collection_info
```
  data: {
    has_eula: false
  }
}
```
edsc.extra.serverless.collection_capabilities: {
  data: {
    cloud_cover: false,
    day_night_flag: false,
    granule_online_access_flag: false,
    orbit_calculated_spatial_domains: false,
    updated_at: "2021-07-20T12:00:35.493Z"
  }
}
```

```
opensearch.granule.osdd: {
  data: "http://fedeo.esa.int/opensearch/description.xml?parentIdentifier=EOP:ESA:FEDEO:COLLECTIONS:CryoSat.products"
}
```

vinnyinverso:
> One question I have is who is tagging these collections? Are we asking the provider to tag their collections? I think we need to update some documentation to explain how that will work and who is responsible.

@chris-durbin I added a ticket for the documentation and linked it to the "shared job results" epic https://bugs.earthdata.nasa.gov/browse/HARMONY-897. I think the DAACs would be responsible for adding the tag, but Patrick probably knows the right answer.

app/models/services/no-op-service.ts
vinnyinverso: Oh interesting. I thought by deleting it here https://github.com/nasa/harmony/blob/harmony-880/app/models/job.ts#L555 it would be inaccessible from job status endpoints.

app/util/cmr.ts
vinnyinverso: Good point! Updated here: https://github.com/nasa/harmony/commit/2c80857ead78e60e45c7425bdb946ae9f98213e8

test/jobs/job-sharing.ts
vinnyinverso: Thanks! Yea that makes sense. I updated it here: https://github.com/nasa/harmony/commit/7f57a795a039e8d9e837911aeda28ae41db4608d

app/models/services/no-op-service.ts
chris-durbin: You're correct. I didn't fully expand this code snippet - I thought this was the output from the noOpService which mimics a job status response, but this is just constructing a job.

app/models/job.ts
vinnyinverso: Those are some good options. I think we can also return tags by wildcard -- e.g. harmony*. I made the key a constant for now: https://github.com/nasa/harmony/commit/41d36f98e3cfed748c8dee0433bb0f1c5c824070

chris-durbin (approved):
I finished locally testing things out. The database migration is working, and I verified that any jobs created prior to the migration (no collection IDs stored) are inaccessible with a different user from the one that created the job, but for requests after the migration using  a public collection I can access the links.

fixtures/cmr.uat.earthdata.nasa.gov-443/162670989255318096
```diff
@@ -0,0 +1,27 @@
+GET /search/collections.json?concept_id=C1233147317-EEDTESTT&page_size=2000&include_tags=true
```
bilts, line 1: It looks like some of these fixtures are outdated (include_tags=true vs include_tags=harmony.has-eula)

bilts (approved):
Verified working with combinations of has_eula and public / non-public collections.  One minor comment about obsolete fixtures to address but otherwise good.

fixtures/cmr.uat.earthdata.nasa.gov-443/162670989255318096
vinnyinverso: Thanks for catching that. Updated here: https://github.com/nasa/harmony/pull/51/commits/5667f0b9bd3f1fa54fc5782b10a661de804b332b

PR #55, opened 2021-07-27
Harmony 290

hailiangzhang:
This turns out an easy fix. The original code overwrites query parameters by copying from the form parameters, which lost the temporal selection from the query string. The fix simply changed it from copy to merge.

### DIFF
--- a/app/frontends/ogc-coverages/post-coverage-rangeset.ts
+++ b/app/frontends/ogc-coverages/post-coverage-rangeset.ts
@@ -1,7 +1,7 @@
 import { Response, NextFunction } from 'express';
+import { RequestValidationError } from 'util/errors';
 import getCoverageRangeset from './get-coverage-rangeset';
 import HarmonyRequest from '../../models/harmony-request';
-import { RequestValidationError, NotFoundError } from 'util/errors';
 
 /**
  * Express middleware that responds to OGC API - Coverages coverage
@@ -24,7 +24,7 @@ export default function postCoverageRangeset(
   // merge form parameters into the query
   const queryKeys = Object.keys(req.query);
   const bodyKeys = Object.keys(req.body);
-  const duplicateKeys = queryKeys.filter(x => bodyKeys.includes(x));
+  const duplicateKeys = queryKeys.filter((x) => bodyKeys.includes(x));
   if (duplicateKeys.length) {
     throw new RequestValidationError(`Duplicate keys "${duplicateKeys}" found from request body and query string!`);
   }
--- a/app/frontends/ogc-coverages/post-coverage-rangeset.ts
+++ b/app/frontends/ogc-coverages/post-coverage-rangeset.ts
@@ -1,6 +1,7 @@
 import { Response, NextFunction } from 'express';
 import getCoverageRangeset from './get-coverage-rangeset';
 import HarmonyRequest from '../../models/harmony-request';
+import { RequestValidationError, NotFoundError } from 'util/errors';
 
 /**
  * Express middleware that responds to OGC API - Coverages coverage
@@ -21,10 +22,11 @@ export default function postCoverageRangeset(
   next: NextFunction,
 ): void {
   // merge form parameters into the query
-  let queryKeys = Object.keys(req.query);
-  let bodyKeys = Object.keys(req.body);
-  if (queryKeys.filter(x => bodyKeys.includes(x)).length) {
-    req.context.logger.warn("Duplicate keys found from request body and query string. We Will use that from request body.");
+  const queryKeys = Object.keys(req.query);
+  const bodyKeys = Object.keys(req.body);
+  const duplicateKeys = queryKeys.filter(x => bodyKeys.includes(x));
+  if (duplicateKeys.length) {
+    throw new RequestValidationError(`Duplicate keys "${duplicateKeys}" found from request body and query string!`);
   }
   req.query = { ...req.query, ...req.body };
 
--- a/app/frontends/ogc-coverages/post-coverage-rangeset.ts
+++ b/app/frontends/ogc-coverages/post-coverage-rangeset.ts
@@ -21,7 +21,7 @@ export default function postCoverageRangeset(
   next: NextFunction,
 ): void {
   // copy form parameters into the query
-  req.query = req.body;
+  req.query = {...req.query, ...req.body};
 
   getCoverageRangeset(req, res, next);
 }
--- a/app/frontends/ogc-coverages/post-coverage-rangeset.ts
+++ b/app/frontends/ogc-coverages/post-coverage-rangeset.ts
@@ -21,7 +21,7 @@ export default function postCoverageRangeset(
   next: NextFunction,
 ): void {
   // copy form parameters into the query
-  req.query = {...req.query, ...req.body};
+  req.query = { ...req.query, ...req.body };
 
   getCoverageRangeset(req, res, next);
 }
--- a/app/frontends/ogc-coverages/post-coverage-rangeset.ts
+++ b/app/frontends/ogc-coverages/post-coverage-rangeset.ts
@@ -20,7 +20,7 @@ export default function postCoverageRangeset(
   res: Response,
   next: NextFunction,
 ): void {
-  // copy form parameters into the query
+  // merge form parameters into the query
   req.query = { ...req.query, ...req.body };
 
   getCoverageRangeset(req, res, next);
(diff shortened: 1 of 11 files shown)


chris-durbin:
It seems like we should add a test that passing a shapefile and query parameters results in the query parameters being set. We'd need to call something similar to `postRangesetRequest` in `harmony/test/helpers/ogc-api-coverages.ts` that also takes a query parameter along with the form parameters.

app/frontends/ogc-coverages/post-coverage-rangeset.ts
```diff
@@ -20,8 +20,8 @@ export default function postCoverageRangeset(
   res: Response,
   next: NextFunction,
 ): void {
-  // copy form parameters into the query
-  req.query = req.body;
+  // merge form parameters into the query
+  req.query = { ...req.query, ...req.body };
```
chris-durbin, line 24: This means that if the same key is passed in both as a query parameter and a form parameter we will use the value from the form. Does that make sense? Should we do any kind of validation, or log anything?

hailiangzhang: Good point. So, if same key exists from both query and form parameter, which one should we use?

indiejames: I think if the same key is passed in both places with different values it is probably a mistake and should result in an error.

indiejames (changes requested):
I think if the same key is passed in both places with different values it is probably a mistake and should result in an error.

chris-durbin (changes requested):
I see several linter errors reported.

hailiangzhang:
> I see several linter errors reported.

Yes, trying to add warning message as you suggested, and didn't clean the code yet.

hailiangzhang:
> I think if the same key is passed in both places with different values it is probably a mistake and should result in an error.

In my post recent fix I issue a warning and choose to use the value from body, but, yes, probably user may miss it, and I will return an error instead.

hailiangzhang:
> I think if the same key is passed in both places with different values it is probably a mistake and should result in an error.

Ok, now I have issued an error instead warning when duplicate values found.

hailiangzhang:
> I see several linter errors reported.

Linter issues fixed.

indiejames (changes requested):
Need to add test that reproduces original bug to verify fix

hailiangzhang:
> Need to add test that reproduces original bug to verify fix

Oh yes, Chris did mention it as well ... will do.

hailiangzhang:
Tests added to duplicate/mixed items from form and query parameter ... ready to be reviewed again.

test/ogc-api-coverages/get-coverage-rangeset.ts
```diff
@@ -198,6 +198,44 @@ describe('OGC API Coverages - getCoverageRangeset', function () {
...
+      StubService.hook({ params: { redirect: 'http://example.com' } });
+      hookPostRangesetRequest(
+        version,
+        collection,
+        variableName,
+        queryLocal,
+        queryParameterString,
+      );
+
+      it('successfully queries CMR and accepts the request', function () {
```
chris-durbin, line 215: This check would not have been failed due to the original bug. I think you should add one more check that the temporal range is included in what gets passed into the backend service.

chris-durbin: Something like these checks:
```
      it('passes the temporal range to the backend service', function () {
        const { start, end } = this.service.operation.temporal;
        expect(start).to.equal('2020-01-02T00:00:00.000Z');
        expect(end).to.equal('2020-01-02T01:00:00.000Z');
      });

      it('identifies the correct granule based on time range', function () {
        const source = this.service.operation.sources[0];
        expect(source.granules.length === 1);
        expect(source.granules[0].id).to.equal('G1233800352-EEDTEST');
      });
```

hailiangzhang: Yes we should make sure this info is passed to backend successfully ... I added your suggestions (I don't think I can improve that:).

bilts:
I noticed some weird behavior in the jobs response:

This request, mixing time and forceAsync params:
```
curl -n -Lbj -F "shapefile=@docs/notebook_helpers/namibia.geo.json;type=application/geo+json" -F 'forceAsync=true' "http://localhost:3000/C1234088182-EEDTEST/ogc-api-coverages/1.0.0/collections/red_var/coverage/rangeset?subset=time("2020-02-16T00:00:00Z":"2020-03-02T00:00:00Z")"
```
Resulting json snippet:
```
"request":"http://localhost:3000/C1234088182-EEDTEST/ogc-api-coverages/1.0.0/collections/red_var/coverage/rangeset?forceAsync=true"
```
The "time" param is eaten but the "forceAsync" from the POST is placed in the query string

If "time" is placed in the POST body, it appears in the "request" URL in the output as expected

app/frontends/ogc-coverages/post-coverage-rangeset.ts
```diff
@@ -20,8 +21,14 @@ export default function postCoverageRangeset(
...
 ): void {
-  // copy form parameters into the query
-  req.query = req.body;
+  // merge form parameters into the query
+  const queryKeys = Object.keys(req.query);
+  const bodyKeys = Object.keys(req.body);
+  const duplicateKeys = queryKeys.filter((x) => bodyKeys.includes(x));
+  if (duplicateKeys.length) {
+    throw new RequestValidationError(`Duplicate keys "${duplicateKeys}" found from request body and query string!`);
+  }
```
bilts, line 30: Re: the next line, note that we do use `req.query` to reverse out which URL to provide in the jobs table and possibly in other places.  I don't see any issues being caused by this, but it's something to keep an eye out for.

bilts (changes requested):
I don't think this is working, or I am missing something:

POST with no query parameters.  Note "numInputGranules" in the response:
```
$ curl -n -Lbj -F "shapefile=@docs/notebook_helpers/namibia.geo.json;type=application/geo+json" -F 'forceAsync=true' -F 'subset=time("2020-02-16T00:00:00Z":"2020-03-02T00:00:00Z")' "http://localhost:3000/C1234088182-EEDTEST/ogc-api-coverages/1.0.0/collections/red_var/coverage/rangeset"
{"username":"pquinn","status":"running","message":"The job is being processed","progress":0,"createdAt":"2021-08-18T14:27:35.218Z","updatedAt":"2021-08-18T14:27:35.218Z","links":[{"href":"http://localhost:3000/jobs/cc2db1b3-31a7-4558-9f97-1d38dd7fd061?page=1&limit=2000","title":"The current page","type":"application/json","rel":"self"}],"request":"http://localhost:3000/C1234088182-EEDTEST/ogc-api-coverages/1.0.0/collections/red_var/coverage/rangeset?forceAsync=true&subset=time(%222020-02-16T00%3A00%3A00Z%22%3A%222020-03-02T00%3A00%3A00Z%22)","numInputGranules":1,"jobID":"cc2db1b3-31a7-4558-9f97-1d38dd7fd061"}
```

Same request, this time moving the "subset=time" parameter to the query:
```
curl -n -Lbj -F "shapefile=@docs/notebook_helpers/namibia.geo.json;type=application/geo+json" -F 'forceAsync=true' "http://localhost:3000/C1234088182-EEDTEST/ogc-api-coverages/1.0.0/collections/red_var/coverage/rangeset?subset=time("2020-02-16T00:00:00Z":"2020-03-02T00:00:00Z")"
{"username":"pquinn","status":"running","message":"The job is being processed","progress":0,"createdAt":"2021-08-18T14:31:56.073Z","updatedAt":"2021-08-18T14:31:56.073Z","links":[{"href":"http://localhost:3000/jobs/a7886f17-55a5-4d1c-a377-1f178ad43961?page=1&limit=2000","title":"The current page","type":"application/json","rel":"self"}],"request":"http://localhost:3000/C1234088182-EEDTEST/ogc-api-coverages/1.0.0/collections/red_var/coverage/rangeset?forceAsync=true","numInputGranules":32,"jobID":"a7886f17-55a5-4d1c-a377-1f178ad43961"}
```

Note that `numInputGranules` is 32 in the second case vs 1 in the first, which is likely because it is not seeing/using the temporal constraint.  The `curl` in the ticket is writing a TIFF output file.  I suspect this is hiding that one of the requests is returning a JSON job while the other is returning an actual TIFF.

app/frontends/ogc-coverages/post-coverage-rangeset.ts
hailiangzhang: @bilts , I have addressed the issues as we discussed by implementing the parameter merging in `cookies.ts`. I also moved the merging function into `util` so that it can be used by multiple places.

bilts (approved):

indiejames (approved):

PR #56, opened 2021-07-28
HARMONY-897: Add documentation for sharing jobs

vinnyinverso:


indiejames (approved):

chris-durbin (approved):
I think we also need to make providers aware of this change so that they go in and add tag associations for any collections that should allow job sharing. Who is going to contact them and do we have contact info for all of our providers?

For the harmony collections in UAT and Prod, one of us should add the tag associations now.

vinnyinverso:
> I think we also need to make providers aware of this change so that they go in and add tag associations for any collections that should allow job sharing. Who is going to contact them and do we have contact info for all of our providers?
> 
> For the harmony collections in UAT and Prod, one of us should add the tag associations now.

@chris-durbin I can ask Jeanne or Justin for DAAC contact / mailing list unless you have a better idea. 

For associations, I think UAT EED-Test collections are mostly taken care of since we tagged a lot of the Harmony collections for the tests? Which prod collections would we need to associate with the new tag?

I still need to create the harmony.has-eula tag in prod once I have permission.

chris-durbin:
> Which prod collections would we need to associate with the new tag?

C1756916832-XYZ_PROV

vinnyinverso:
> C1756916832-XYZ_PROV

@chris-durbin the tag was created, however, when tagging the above collection, the response was:

```[
  {
    "errors": [
      "Collection [C1756916832-XYZ_PROV] does not exist or is not visible."
    ],
    "tagged_item": {
      "concept_id": "C1756916832-XYZ_PROV"
    }
  }
]```

PR #59, opened 2021-08-05
HARMONY-904:Fix security Vulnerabilities

blackone-sudo:


package.json
```diff
@@ -115,7 +115,7 @@
     "unzipper": "^0.10.11",
     "uuid": "^8.3.2",
     "winston": "^3.3.3",
-    "xmldom": "^0.6.0",
+    "xmldom": "github:xmldom/xmldom#0.7.0",
```
chris-durbin, line 118: FYI - this is the workaround required until they can get the library published to npm. Once published to npm we should change this to:
`"xmldom": "^0.7.0"`

chris-durbin (approved):

package.json
```diff
@@ -103,7 +103,7 @@
     "shpjs": "^3.5.0",
     "simple-oauth2": "git+https://github.com/bilts/simple-oauth2.git#fix-encoding",
     "source-map-support": "^0.5.19",
-    "sqlite3": "^5.0.2",
+    "sqlite3": "^4.2.0",
```
indiejames, line 106: Was there no 5.x version with a fix for the vulnerability? Not sure dropping down a major version is the right thing to do.

indiejames (approved):

PR #69, opened 2021-09-15
Harmony 937

indiejames:
Add resiliency to pull worker to prevent it from crashing.


tasks/service-runner/app/service/service-runner.ts
```diff
@@ -72,63 +72,71 @@ function _getErrorMessage(logStr: string): string {
...
-          const errMsg = `${env.harmonyService}: ${logErr}`;
-          stdOut.destroy();
-          resolve({ error: errMsg });
-        }
-      },
-    ).catch((e) => {
-      clearTimeout(timeout);
-      logger.error(e.message);
-      resolve({ error: e.message });
+export async function runQueryCmrFromPull(workItem: WorkItem): Promise<ServiceResponse> {
```
bilts, line 75: Is async correct / necessary here, given that there is no await statement inside the function?

tasks/service-runner/app/service/service-runner.ts
```diff
@@ -72,63 +72,71 @@ function _getErrorMessage(logStr: string): string {
...
+      // create a writable stream to capture stdout from the exec call
+      // using stdout instead of stderr because the service library seems to log ERROR to stdout
+      const stdOut = new LogStream();
+
+      // timeout if things take too long
+      const timeout = setTimeout(() => {
+        resolve({ error: `Worker timed out after ${workerTimeout / 1000.0} seconds` });
+      }, workerTimeout);
+
+      exec.exec(
```
bilts, line 100: It's been too long since I had to worry about promises like this: If this method throws synchronously, does the exception get automatically propagated to the `.catch`?

bilts (approved):
This all looks reasonable.  I do find it really tricky to follow the promise flow through both process.exec and timeout, which makes me feel less confident that everything is caught, and handled.  I don't yet have a better suggestion, though, so 👍

tasks/service-runner/app/service/service-runner.ts
indiejames: I added async there so the `return` statement in the final `catch` block can just return an object instead of a promise. The `async` here wraps that return value in a promise.

tasks/service-runner/app/service/service-runner.ts
indiejames: As I understand it `.catch` is just handling rejected promises, so if the `exec` call fails synchronously it will bubble up to the outer `try/catch` and be handled there.

tasks/service-runner/app/service/service-runner.ts
```diff
@@ -55,11 +55,11 @@ function _getStacCatalogs(dir: string): string[] {
  */
 function _getErrorMessage(logStr: string): string {
   // expect JSON logs entries
-  const regex = /\{"message":.*?"application":.*?\}/g;
-  const matches = logStr.match(regex);
+  const regex = /\{.*?\}/gs;
```
chris-durbin, line 58: Do you see this being a problem if we have a large number of messages logged by the backend service?

tasks/service-runner/app/workers/pull-worker.ts
```diff
@@ -12,6 +12,9 @@ const activeSocketKeepAlive = 6_000;
 const maxSockets = 1;
 const maxFreeSockets = 1;
 const maxRetries = 3;
+let pullCounter = 0;
+// how many pulls to execute before logging - used to keep log message count reasonable
+const pullLogPeriod = 10;
```
chris-durbin, line 17: 😬  - this is going to add logging every 5 seconds when there is no work. I'm going to miss the nice clean logs that only logged messages when there were requests going on.

chris-durbin (approved):

tasks/service-runner/app/service/service-runner.ts
indiejames: Thats a non-greedy regex, so it will match the first '{' to the first following '}'. So it should get one valid log entry. I have added a `try/catch` around this anyway in case the extracted match is not valid JSON for some reason.

tasks/service-runner/app/workers/pull-worker.ts
indiejames: We could add another env var 😬

indiejames:
The SNYK finding does not affect us due to the limited way in which we are using the module.

PR #70, opened 2021-09-17
HARMONY-939: Sanitize sensitive information from image-related strings

vinnyinverso:


vinnyinverso:
I was on the fence about whether to combine `sanitizeImageStrict` and `sanitizeImage`.

There seems to be a tradeoff between specificity/rigidity of the regexp and flexibility/robustness. The original `sanitizeImage` could afford to be a bit more lax in specificity in exchange for being more flexible because the inputs (full path to an image) are well known. For this newer use case `sanitizeImageStrict`, where me may be sanitizing arbitrary strings, we have to be a bit pickier about what we match on so that we don't accidentally remove more than we want.

I was reluctant to force the older `sanitizeImage` to be stricter about what it removes as it will become more brittle, however it does seem more ideal to only have one method for maintainability.

vinnyinverso:
In addition to this change, I'm wondering if it makes sense to scrub any of this further down the chain (e.g. does the workflow step really need the full path of the image) to reduce the chance that it leaks again? I'm not sure how much that extra bit of info helps with debugging if we can retrieve it elsewhere?

app/util/string.ts
```diff
@@ -58,3 +58,39 @@ export function truncateString(s: string, n: number): string {
...
+    .replace(/m[a-z]+.earthdata.nasa.gov\//i, '');
+}
+
+/**
+ * Returns true if the image repository for the given image is ECR
+ *
+ * @param image - the full image string
+ * @returns true if the image is in ECR and false otherwise
+ */
+export function inECR(image: string): boolean {
```
indiejames, line 81: nit: proper camel casing would be `inEcr`

app/util/string.ts
```diff
@@ -58,3 +58,39 @@ export function truncateString(s: string, n: number): string {
...
 }
+
+/**
+ * Removes AWS account ECR information or *.earthdata.nasa.gov from the image name
+ * since we may not want to expose that information.
+ *
+ * @param image - The image name to sanitize
+ * @returns the sanitized string
+ */
+export function sanitizeImage(image: string): string {
```
indiejames: All of these functions should have unit tests

indiejames (changes requested):
There are some small changes mentioned in my comments, but the main thing is that you are not addressing sanitizing error messages, which is mentioned in the AC for this ticket. See lines 127 and 200 of service-runner.ts. Although it might be better to sanitize messages in the harmony code so we catch things at the edges.

vinnyinverso:
> There are some small changes mentioned in my comments, but the main thing is that you are not addressing sanitizing error messages, which is mentioned in the AC for this ticket. See lines 127 and 200 of service-runner.ts. Although it might be better to sanitize messages in the harmony code so we catch things at the edges.

Ah right. For some reason I was only thinking of the messages in the UI 🤦. Makes much more sense to catch it as early as possible. I'll address these issues Monday.

vinnyinverso:
[] revert changes to sanitize method
[] remove sanitization of job.message in favor of sanitizing in service runner
[] tests for service runner and string util methods

app/util/string.ts
chris-durbin: I spent way too long trying to figure out if your nit was correct for acronyms, and clearly there is no right answer.

vinnyinverso:
Issues mentioned are fixed now. (Image names are sanitized in logging code and unit tests have been added.)

vinnyinverso:
Also added some turbo guidance in the readme for harmony developers who are not running harmony in a box. Feel free to suggest any modifications: https://github.com/nasa/harmony/pull/70/commits/46edca9396169dbaf99cb70e834c34cafeee5ecd

app/frontends/jobs.ts
```diff
@@ -363,13 +363,12 @@ export async function getWorkItemsForWorkflowUI(
...
       const previousPage = pageLinks.find((l) => l.rel === 'prev');
       setPagingHeaders(res, pagination);
-      const workflowSteps = await getWorkflowStepsByJobId(db, job.jobID);
       res.render('jobs/workflow-items-table', {
         workItems,
         workflowItemUpdated() { return (new Date(this.updatedAt).toISOString()); },
         workflowItemCreated() { return (new Date(this.createdAt).toISOString()); },
         workflowItemBadge() { return badgeClasses[this.status]; },
-        workflowItemStep() { return workflowSteps[this.workflowStepIndex - 1].serviceID; },
+        workflowItemStep() { return sanitizeImage(this.serviceID); },
```
bilts, line 371: Is `this` the workflow item in this context?

vinnyinverso: Yup!

indiejames (approved):

PR #73, opened 2021-09-22
Harmony 926

hailiangzhang:
This PR handles the user cancel requests for TURBO workflow.


app/models/work-item.ts
```diff
@@ -121,7 +121,9 @@ export async function getNextWorkItem(
           .on(`${WorkflowStep.table}.stepIndex`, `${WorkItem.table}.workflowStepIndex`)
           .on(`${WorkflowStep.table}.jobID`, `${WorkItem.table}.jobID`);
       })
-      .where({ 'work_items.serviceID': serviceID, status: WorkItemStatus.READY })
+      .join(Job.table, `${WorkItem.table}.jobID`, '=', `${Job.table}.jobID`)
+      .where({ 'work_items.serviceID': serviceID, 'work_items.status': WorkItemStatus.READY })
+      .whereIn('jobs.status', [JobStatus.RUNNING, JobStatus.ACCEPTED])
```
bilts, line 126: This clause seems possibly redundant, given where things are in the if-statement, and I could see it breaking things in the future if we ever added another job status.

app/util/job.ts
```diff
@@ -29,13 +30,18 @@ export default async function cancelAndSaveJob(
...
     }
+
     if (job) {
       if (job.status !== JobStatus.CANCELED || !shouldIgnoreRepeats) {
         job.status = JobStatus.CANCELED;
         job.validateStatus();
         job.cancel(message);
         await job.save(tx);
-        if (shouldTerminateWorkflows) {
+        const isTurboWorkflow = await checkIfTurboWorkflow(tx, jobID, logger);
```
bilts, line 40: Knowing that the plan is to deprecate Argo, I think you could probably rearrange this to something like the following to make it easier to find and remember to delete the Argo stuff:

```
        const updatedAt = new Date();
        await tx('work_items').where({ jobID: job.jobID }).update({ status: WorkItemStatus.CANCELED, updatedAt });

        // The following can be removed once Argo is removed
        const isArgoWorkflow = !await checkIfTurboWorkflow(tx, jobID, logger);
        if (isArgoWorkflow && shouldTerminateWorkflows) {
          await terminateWorkflows(job, logger);
        }
```

test/work-backends.ts
```diff
@@ -32,6 +32,17 @@ describe('Work Backends', function () {
   hookServersStartStop({ skipEarthdataLogin: true });
 
   describe('when getting a work item', function () {
+    const runningJob = new Job({
```
bilts, line 35: Does this actually test the issue?

bilts (approved):

app/backends/workflow-orchestration.ts
```diff
@@ -153,8 +153,11 @@ export async function updateWorkItem(req: HarmonyRequest, res: Response): Promis
     workItem = await getWorkItemById(tx, parseInt(id, 10));
     const job: Job = await Job.byJobID(tx, workItem.jobID);
 
+    // If the job was already canceled then send 400 response
+    if (job.status === JobStatus.CANCELED) {
+      res.status(400).send();
```
chris-durbin, line 158: I think we send a 409 in other places to indicate a conflict. It's probably best to also have a message indicating the issue.

It would probably be best to throw a `ConflictError` here with a message and then the harmony code at the top level will make sure to format the message with the appropriate content type and status code of 409 (that is you shouldn't have to modify any other code - should just work).

app/util/workflows.ts
```diff
@@ -121,3 +123,30 @@ export async function terminateWorkflows(job: Job, logger: Logger): Promise<void
...
+  transaction: Transaction,
+  jobID: string,
+  logger: Logger,
+): Promise<boolean> {
+  let isTurboWorkflow = false;
+  try {
+    const hasWorkItemsTable = await transaction.schema.hasTable('work_items');
+    if (hasWorkItemsTable) {
+      const workItemCount = await workItemCountForJobID(transaction, jobID);
+      if (workItemCount) isTurboWorkflow = true;
```
chris-durbin, line 145: while this works because 0 is falsey, it probably is easier to follow if we treat workItemCount as a number:
```
if (workItemCount > 0) isTurboWorkflow = true;
```

chris-durbin (changes requested):
I think we need tests for canceling turbo workflows, and tests that attempt to update work items for a job that has been canceled.

app/backends/workflow-orchestration.ts
```diff
@@ -153,8 +153,11 @@ export async function updateWorkItem(req: HarmonyRequest, res: Response): Promis
     workItem = await getWorkItemById(tx, parseInt(id, 10));
     const job: Job = await Job.byJobID(tx, workItem.jobID);
 
+    // If the job was already canceled then send 400 response
+    if (job.status === JobStatus.CANCELED) {
+      res.status(409).send('Job was already canceled.');
```
chris-durbin, line 158: Did attempting to throw a ConflictError not work here?

test/util/job.ts
```diff
@@ -28,15 +39,26 @@ describe('Canceling a job', async function () {
...
     });
+
+    it('does not terminates the workflow', async function () {
+      await cancelAndSaveJob(anotherArgoJob.requestId, 'Canceled by admin', log, false, 'joe');
+      expect(terminateWorkflowsStub.callCount).to.equal(0);
+    });
   });
 
-  describe('when workflow termination is not requested', async function () {
+  describe('when cancelation is requested for a turbo workflow', async function () {
```
chris-durbin, line 52: Add one more test with work items in each of the states and verify it only updates the status to canceled for the ready and running ones.

chris-durbin (commented):
I noticed that any work items which were actively being worked when the cancel was sent continue to retry a few times on the error they receive back from harmony. Ideally they would not attempt to retry when they receive a 409 (this retry is in the service-runner code). When they are retrying they are keeping one of the service workers from trying any other requests so it is probably worth us changing to not retry on a 409.

hailiangzhang:
> I noticed that any work items which were actively being worked when the cancel was sent continue to retry a few times on the error they receive back from harmony. Ideally they would not attempt to retry when they receive a 409 (this retry is in the service-runner code). When they are retrying they are keeping one of the service workers from trying any other requests so it is probably worth us changing to not retry on a 409.



> I noticed that any work items which were actively being worked when the cancel was sent continue to retry a few times on the error they receive back from harmony. Ideally they would not attempt to retry when they receive a 409 (this retry is in the service-runner code). When they are retrying they are keeping one of the service workers from trying any other requests so it is probably worth us changing to not retry on a 409.

So, do you think it's clean if I add the following in [`tasks/service-runner/app/workers/pull-worker.ts:_pullAndDoWork`](https://github.com/nasa/harmony/blob/d2e11a89c026a72f8de85dca9020bd9e92a1a9d9/tasks/service-runner/app/workers/pull-worker.ts#L113)
```
          if (response.status == 409) {
            complete = true;
         }
```

chris-durbin:
> So, do you think it's clean if I add the following in [`tasks/service-runner/app/workers/pull-worker.ts:_pullAndDoWork`](https://github.com/nasa/harmony/blob/d2e11a89c026a72f8de85dca9020bd9e92a1a9d9/tasks/service-runner/app/workers/pull-worker.ts#L113)
> 
> ```
>           if (response.status == 409) {
>             complete = true;
>          }
> ```
> 
> (or change it to `if (response.status >= 400 && response.status !== 409) {`)

We should probably log if we get a 409 back with the message that gets returned, and then mark it as complete so I'd go with your first one and add a log message there.

hailiangzhang:
> > So, do you think it's clean if I add the following in [`tasks/service-runner/app/workers/pull-worker.ts:_pullAndDoWork`](https://github.com/nasa/harmony/blob/d2e11a89c026a72f8de85dca9020bd9e92a1a9d9/tasks/service-runner/app/workers/pull-worker.ts#L113)
> > ```
> >           if (response.status == 409) {
> >             complete = true;
> >          }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > (or change it to `if (response.status >= 400 && response.status !== 409) {`)
> 
> We should probably log if we get a 409 back with the message that gets returned, and then mark it as complete so I'd go with your first one and add a log message there.

Tried hard to make debugging easy in service-runner (by replacing the entry point by things like`watch-debug`), but no luck yet.
With the time-consuming way, I was trying to print `response` and `response.data` but all seem empty.  So, what should I log `with the message that gets returned`?

hailiangzhang:
> > > So, do you think it's clean if I add the following in [`tasks/service-runner/app/workers/pull-worker.ts:_pullAndDoWork`](https://github.com/nasa/harmony/blob/d2e11a89c026a72f8de85dca9020bd9e92a1a9d9/tasks/service-runner/app/workers/pull-worker.ts#L113)
> > > ```
> > >           if (response.status == 409) {
> > >             complete = true;
> > >          }
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > (or change it to `if (response.status >= 400 && response.status !== 409) {`)
> > 
> > 
> > We should probably log if we get a 409 back with the message that gets returned, and then mark it as complete so I'd go with your first one and add a log message there.
> 
> Tried hard to make debugging easy in service-runner (by replacing the entry point by things like`watch-debug`), but no luck yet. With the time-consuming way, I was trying to print `response` and `response.data` but all seem empty. So, what should I log `with the message that gets returned`?

I think I may have found an issue when we handle the response from updating the workitem via `axios.put`. If the request ends up with some error code, it seems raising an exception, and therefore we need to handle it inside the `catch(e)` block.

I have fixed the 409 status code and committed the changes.

hailiangzhang:
> I noticed that any work items which were actively being worked when the cancel was sent continue to retry a few times on the error they receive back from harmony. Ideally they would not attempt to retry when they receive a 409 (this retry is in the service-runner code). When they are retrying they are keeping one of the service workers from trying any other requests so it is probably worth us changing to not retry on a 409.

I think I have addressed this issue (so that it doesn't retry when receiving 409 response) from [this comment](https://github.com/nasa/harmony/pull/73#issuecomment-930338828)

test/util/job.ts
```diff
@@ -28,15 +47,38 @@ describe('Canceling a job', async function () {
...
+    });
+
     it('does not terminates the workflow', async function () {
-      await cancelAndSaveJob(anotherJob.requestId, 'Canceled by admin', log, true, 'joe');
-      expect(terminateWorkflowsStub.callCount).to.equal(1);
+      await cancelAndSaveJob(anotherTurboJob.requestId, 'Canceled by admin', log, true, 'doe');
+      expect(terminateWorkflowsStub.callCount).to.equal(0);
+    });
+
+    it('fails to cancel an already-canceled worklow', async function () {
```
indiejames, line 72: 'workflow' typo

test/util/job.ts
```diff
@@ -28,15 +47,38 @@ describe('Canceling a job', async function () {
...
+      await expect(
+        cancelAndSaveJob(anotherTurboJob.requestId, 'Canceled by admin', log, true, 'doe'),
+      ).to.be.rejectedWith('Job status cannot be updated from canceled to canceled.');
+    });
+
+    it('fails to cancel an already-finished worklow', async function () {
```
indiejames, line 78: 'workflow' typo

indiejames (commented):
There are still some unanswered questions in the comments. I will approve this if you address Chris's request to add a test case to verify that only work items in the 'READY' or 'RUNNING' state get set to 'CANCELED'.

hailiangzhang:
> There are still some unanswered questions in the comments. I will approve this if you address Chris's request to add a test case to verify that only work items in the 'READY' or 'RUNNING' state get set to 'CANCELED'.

Oh, I thought I have answered [here](https://github.com/nasa/harmony/pull/73#discussion_r716720825):


> There are still some unanswered questions in the comments. I will approve this if you address Chris's request to add a test case to verify that only work items in the 'READY' or 'RUNNING' state get set to 'CANCELED'.

Sorry I thought I did [answer Chris's request for more tests](https://github.com/nasa/harmony/pull/73#discussion_r716720825), but you are correct, it's still missing one more test to make sure it can cancel job in "READY" state (although it seems all jobs will be initialized with "RUNNING" state but I could be wrong).

Alright, more tests committed which will make sure we can cancel job in "READY" and "RUNNING" states, but not for those in "FINISHED", "FAILED", or "CANCELED" states.

indiejames (approved):

PR #77, opened 2021-10-04
Harmony 943

vinnyinverso:
Cancel work items when a job fails.

### DIFF
--- a/app/backends/workflow-orchestration.ts
+++ b/app/backends/workflow-orchestration.ts
@@ -7,7 +7,7 @@ import { readCatalogItems } from '../util/stac';
 import HarmonyRequest from '../models/harmony-request';
 import { Job, JobStatus } from '../models/job';
 import JobLink from '../models/job-link';
-import WorkItem, { getNextWorkItem, WorkItemStatus, updateWorkItemStatus, getWorkItemById, workItemCountForStep } from '../models/work-item';
+import WorkItem, { getNextWorkItem, WorkItemStatus, updateWorkItemStatus, getWorkItemById, workItemCountForStep, updateWorkItemsStatusByJobId } from '../models/work-item';
 import WorkflowStep, { getWorkflowStepByJobIdStepIndex } from '../models/workflow-steps';
 
 const MAX_TRY_COUNT = 1;
@@ -159,16 +159,18 @@ export async function updateWorkItem(req: HarmonyRequest, res: Response): Promis
     // If the response is an error then set the job status to 'failed'
     } else if (workItem.status === WorkItemStatus.FAILED) {
       if (![JobStatus.FAILED, JobStatus.CANCELED].includes(job.status)) {
-        job.status = JobStatus.FAILED;
         let message: string;
         if (errorMessage) {
           message = `WorkItem [${workItem.id}] failed with error: ${errorMessage}`;
         } else {
           message = 'Unknown error';
         }
-        job.message = message;
+        job.fail(message);
         try {
           await job.save(tx);
+          await updateWorkItemsStatusByJobId(
+            tx, job.jobID, [WorkItemStatus.READY, WorkItemStatus.RUNNING], WorkItemStatus.CANCELED,
+          );
         } catch (e) {
           logger.error('Failed to update job');
           logger.error(e);
--- a/app/models/work-item.ts
+++ b/app/models/work-item.ts
@@ -168,6 +168,31 @@ export async function updateWorkItemStatus(
 }
 
 /**
+ * Update the status of work items by job ID.
+ * @param tx - the transaction to use for the update
+ * @param jobID - the jobID associated with the work items
+ * @param oldStatuses - restricts the updates to work items where the status is in oldStatuses
+ * @param newStatus - the value of the updated status
+ */
+export async function updateWorkItemsStatusByJobId(
+  tx: Transaction,
+  jobID: string,
+  oldStatuses: WorkItemStatus[],
+  newStatus: WorkItemStatus,
+): Promise<void> {
+  const updatedAt = new Date();
+  await tx(WorkItem.table)
+    .where({ jobID })
+    .modify((queryBuilder) => {
+      if (oldStatuses.length) {
+        queryBuilder
+          .whereIn('status', oldStatuses);
+      }
+    })
+    .update({ status: newStatus, updatedAt });
+}
+
+/**
  * Returns the next work item to process for a service
  * @param tx - the transaction to use for querying
  * @param id - the work item ID
--- a/app/util/job.ts
+++ b/app/util/job.ts
@@ -1,7 +1,7 @@
 import { Logger } from 'winston';
 import db from './db';
 import { Job, JobStatus } from '../models/job';
-import { WorkItemStatus } from '../models/work-item';
+import { updateWorkItemsStatusByJobId, WorkItemStatus } from '../models/work-item';
 import { NotFoundError, RequestValidationError } from './errors';
 import { terminateWorkflows, checkIfTurboWorkflow } from './workflows';
 import isUUID from './uuid';
@@ -38,11 +38,9 @@ export default async function cancelAndSaveJob(
         job.validateStatus();
         job.cancel(message);
         await job.save(tx);
-        const updatedAt = new Date();
-        await tx('work_items')
-          .where({ jobID: job.jobID })
-          .whereIn('status', [WorkItemStatus.READY, WorkItemStatus.RUNNING])
-          .update({ status: WorkItemStatus.CANCELED, updatedAt });
+        await updateWorkItemsStatusByJobId(
+          tx, job.jobID, [WorkItemStatus.READY, WorkItemStatus.RUNNING], WorkItemStatus.CANCELED,
+        );
         // The following can be removed once Argo is removed
         const isArgoWorkflow = !await checkIfTurboWorkflow(tx, jobID, logger);
         if (isArgoWorkflow && shouldTerminateWorkflows) {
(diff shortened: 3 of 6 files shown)


app/backends/workflow-orchestration.ts
```diff
@@ -159,16 +159,18 @@ export async function updateWorkItem(req: HarmonyRequest, res: Response): Promis
...
         if (errorMessage) {
           message = `WorkItem [${workItem.id}] failed with error: ${errorMessage}`;
         } else {
           message = 'Unknown error';
         }
-        job.message = message;
+        job.fail(message);
         try {
           await job.save(tx);
+          await updateWorkItemsStatusByJobId(
```
bilts, line 171: I'm wondering if this should all be taken a step further and put in a method on `job`.  Something like `job.complete(tx, finalStatus, message)`.  For that matter, I was a little surprised that this all didn't end up inside the `fail` method and that we have to save and update stuff after calling `fail`.  Maybe that's similar to your question from the stand up about the behavior of `save`.

Either way, I think a method on `Job` that knows all the non-terminal statuses could be helpful if we ever end up adding one

app/models/work-item.ts
```diff
@@ -167,6 +167,31 @@ export async function updateWorkItemStatus(
...
 }
 
+/**
+ * Update the status of work items by job ID.
+ * @param tx - the transaction to use for the update
+ * @param jobID - the jobID associated with the work items
+ * @param oldStatuses - restricts the updates to work items where the status is in oldStatuses
+ * @param newStatus - the value of the updated status
+ */
+export async function updateWorkItemsStatusByJobId(
```
bilts, line 177: Related to prior message, I could see this being a method on the job, i.e. `job.updateWorkItemStatuses`

test/workflow-orchestration.ts
```diff
@@ -138,6 +138,70 @@ describe('Workflow chaining for a collection configured for swot reprojection an
...
+        await updateWorkItem(this.backend, workItem);
+        // since there were multiple query cmr results,
+        // multiple work items should be generated for the next step
+        const currentWorkItems = (await getWorkItemsByJobId(db, workItem.jobID)).workItems;
+        expect(currentWorkItems.length).to.equal(4);
+        expect(currentWorkItems.filter((i) => i.status === WorkItemStatus.READY && i.serviceID === 'sds/swot-reproject:latest').length).to.equal(3);
+      });
+
+      describe('when the first swot-reprojection service work item fails', function () {
+        it('the job fails, and all further work items are canceled', async function () {
```
bilts, line 176: `it('fails the job and cancels all further work items...`

test/workflow-orchestration.ts
```diff
@@ -138,6 +138,70 @@ describe('Workflow chaining for a collection configured for swot reprojection an
...
+          const res = await getWorkForService(this.backend, 'sds/swot-reproject:latest');
+          expect(res.status).to.equal(200);
+          const workItem = JSON.parse(res.text);
+          workItem.status = WorkItemStatus.FAILED;
+          workItem.results = [];
+          await updateWorkItem(this.backend, workItem);
+          expect(workItem.serviceID).to.equal('sds/swot-reproject:latest');
```
bilts, line 183: Here and above probably belong in a `before` block

app/util/job.ts
```diff
@@ -38,11 +38,9 @@ export default async function cancelAndSaveJob(
         job.validateStatus();
         job.cancel(message);
         await job.save(tx);
-        const updatedAt = new Date();
-        await tx('work_items')
-          .where({ jobID: job.jobID })
-          .whereIn('status', [WorkItemStatus.READY, WorkItemStatus.RUNNING])
-          .update({ status: WorkItemStatus.CANCELED, updatedAt });
+        await updateWorkItemsStatusByJobId(
+          tx, job.jobID, [WorkItemStatus.READY, WorkItemStatus.RUNNING], WorkItemStatus.CANCELED,
```
bilts, line 42: I know this isn't quite your change, but if something is `RUNNING` and then gets `CANCELED` like this, presumably whatever call was working on it will call back and try to update its status.  Is that behavior handled properly?

test/workflow-orchestration.ts
```diff
@@ -138,6 +138,70 @@ describe('Workflow chaining for a collection configured for swot reprojection an
...
+          // work item failure should trigger job failure
+          const job = await Job.byJobID(db, workItem.jobID);
+          expect(job.status === JobStatus.FAILED);
+          // job failure should trigger cancellation of any pending work items
+          const currentWorkItems = (await getWorkItemsByJobId(db, workItem.jobID)).workItems;
+          expect(currentWorkItems.length).to.equal(4);
+          expect(currentWorkItems.filter((i) => i.status === WorkItemStatus.SUCCESSFUL && i.serviceID === 'harmonyservices/query-cmr:latest').length).to.equal(1);
+          expect(currentWorkItems.filter((i) => i.status === WorkItemStatus.CANCELED && i.serviceID === 'sds/swot-reproject:latest').length).to.equal(2);
```
bilts, line 191: One way to guard against the issue I mentioned earlier, where we could add a status and it ends up not getting handled correctly, would be to loop over the statuses in the `WorkItemStatus` enum and ensure that one item per status exists, then the test could assert that all statuses end up in the correct bucket.

bilts (approved):
Minor comments.  None are strictly necessary for a merge, though, but they may offer some small improvements.

test/workflow-orchestration.ts
```diff
@@ -138,6 +138,70 @@ describe('Workflow chaining for a collection configured for swot reprojection an
...
+      scaleExtent: '0,2500000.3,1500000,3300000',
+      scaleSize: '1.1,2',
+      format: 'application/x-zarr',
+      turbo: true,
+    };
+
+    hookRangesetRequest('1.0.0', collection, 'all', { query: reprojectAndZarrQuery });
+    hookRedirect('joe');
+
+    describe('when checking for a query-cmr work item', function () {
```
indiejames, line 155: It's a little confusing that you are testing to see if query-cmr gets executed here instead of just executing it to get the side effect that you want (creation of the work items for the next step). In fact, query-cmr is not really special and should have the same behavior as any other service when the job is canceled (shouldn't get any work items), so you could just use it instead of a subsequent service. Not a hug deal, just a bit confusing to see a 'success' test case inside a 'job fails' `describe`.

test/workflow-orchestration.ts
```diff
@@ -138,6 +138,70 @@ describe('Workflow chaining for a collection configured for swot reprojection an
...
+          workItem.results = [];
+          await updateWorkItem(this.backend, workItem);
+          expect(workItem.serviceID).to.equal('sds/swot-reproject:latest');
+          // work item failure should trigger job failure
+          const job = await Job.byJobID(db, workItem.jobID);
+          expect(job.status === JobStatus.FAILED);
+          // job failure should trigger cancellation of any pending work items
+          const currentWorkItems = (await getWorkItemsByJobId(db, workItem.jobID)).workItems;
+          expect(currentWorkItems.length).to.equal(4);
+          expect(currentWorkItems.filter((i) => i.status === WorkItemStatus.SUCCESSFUL && i.serviceID === 'harmonyservices/query-cmr:latest').length).to.equal(1);
```
indiejames, line 190: nit: prefer `item` to `i`

test/workflow-orchestration.ts
```diff
@@ -138,6 +138,70 @@ describe('Workflow chaining for a collection configured for swot reprojection an
...
+          expect(currentWorkItems.filter((i) => i.status === WorkItemStatus.CANCELED && i.serviceID === 'sds/swot-reproject:latest').length).to.equal(2);
+          expect(currentWorkItems.filter((i) => i.status === WorkItemStatus.FAILED && i.serviceID === 'sds/swot-reproject:latest').length).to.equal(1);
+        });
+
+        describe('when checking to see if further reprojection work is queued', function () {
+          hookGetWorkForService('sds/swot-reproject:latest');
```
indiejames, line 196: style nit: Feel free to disagree, but I dislike using hooks for a single test (probably have done it myself, though). I think it would be clearer if you just put a call to `getWorkForService` in the `it`. For me hooks make it harder to know what is going on and I only want to use them if they save a lot of redundant code.

indiejames: Actually, looking at the code above (line 177) it would make sense to move this hook up so both of these test cases are using it since they are making the same call.

indiejames (changes requested):
Only one real  change request regarding use of `hookGetWorkForService` in test/workflow-orchestration.ts.

app/util/job.ts
vinnyinverso: This is what happens in that case:

```
// If the job was already canceled then send 400 response
    if (job.status === JobStatus.CANCELED) {
      res.status(409).send('Job was already canceled.');
```

indiejames (approved):

app/util/job.ts
vinnyinverso: Actually, the status of the work item would get updated again (undesirable) on the call back. I think I can just move the update invocation around to account for this.

vinnyinverso:
@bilts @indiejames 

I think the most recent commit addresses all of the comments.

For the tests, I moved a bunch of logic into the describe blocks and cleaned things up.

For the actual implementation, I did my best to adhere to Patrick's suggestion by adding a helper function `completeJob` in the util/job module. Also, within the `updateWorkItem` route, I had to move the invocation of `updateWorkItemStatus()` after the block that checks whether its job is failed or canceled, to prevent a work item from updating its status (from cancelled to something else) after the job has failed or been canceled.

indiejames (approved):

PR #78, opened 2021-10-05
HARMONY-991: Add CONCISE to services.yml and deploy locally

indiejames:
Initial changes to add CONCISE to Harmony. Deployment (terraform) changes will be in a separate PR in the CI/CD repo.


tasks/service-runner/config/query-cmr-sidecar.yaml
```diff
@@ -97,4 +97,4 @@ spec:
       protocol: TCP
   selector:
     name: query-cmr
-  type: LoadBalancer
+  type: ClusterIP
```
indiejames, line 100: Made this change because 'LoadBalancer' could result in the services being exposed outside the cluster which is not needed and could cause problems later with local development on Docker Desktop.

tasks/service-runner/config/concise-sidecar.yaml
```diff
@@ -0,0 +1,105 @@
...
+            value: $BACKEND_HOST
+          - name: BACKEND_PORT
+            value: "$BACKEND_PORT"
+          - name: HARMONY_SERVICE
+            value: $PODAAC_CONCISE_IMAGE
+          - name: WORKING_DIR
+            value: "/opt/harmony-concise"
+          - name: INVOCATION_ARGS
+            value: |-
+              concise_harmony
```
bilts, line 67: I understand why this is necessary; however, it seems like we're moving toward needing more one-off config per service intermingled with boilerplate.  For the "Definitely some day, but not in this ticket" I'd like us to be trying to put the onus for setting service-specific values on service providers (and making that as low-effort as possible) while getting the boilerplate centralized / generated.

PR #79, opened 2021-10-11
HARMONY-978: Warning messages for turbo requests

chris-durbin:
Previously a generic message was shown for turbo requests "The job is being processed", but now the message matches what is being displayed for Argo requests. 

For example for the request http://localhost:3000/C1233800302-EEDTEST/ogc-api-coverages/1.0.0/collections/all/coverage/rangeset?outputCrs=EPSG%3A4326&format=image%2Ftiff&turbo=true&maxResults=1&forceAsync=true the message is now:
"CMR query identified 177 granules, but the request has been limited to process only the first 1 granules because you requested 1 maxResults."

### DIFF
--- a/app/middleware/cmr-granule-locator.ts
+++ b/app/middleware/cmr-granule-locator.ts
@@ -137,7 +137,7 @@ async function cmrGranuleLocatorNew(
 
       const limitedMessage = getResultsLimitedMessage(operation);
       if (limitedMessage) {
-        req.context.messages.push(getResultsLimitedMessage(operation));
+        req.context.messages.push(limitedMessage);
       }
     });
     await Promise.all(queries);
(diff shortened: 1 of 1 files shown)


app/middleware/cmr-granule-locator.ts
```diff
@@ -134,6 +134,11 @@ async function cmrGranuleLocatorNew(
 
       operation.cmrHits += hits;
       operation.scrollIDs.push(scrollID);
+
+      const limitedMessage = getResultsLimitedMessage(operation);
+      if (limitedMessage) {
+        req.context.messages.push(getResultsLimitedMessage(operation));
```
indiejames, line 140: No need to call `getResultsLimitedMessage()` here since you have already saved the result of that call in `limitedMessage`.

indiejames: Makes me miss `(if-let`

indiejames (approved):
Just one nit

app/util/db.ts
```diff
@@ -1,12 +1,12 @@
...
 // eslint-disable-next-line import/order
 import knexfile from '../../db/knexfile';
-import knex from 'knex';
+import { knex, Knex } from 'knex';
 import { attachPaginate } from 'knex-paginate';
 import env from './env';
 import logger from './log';
 
-export type Transaction = knex.Transaction | knex;
+export type Transaction = Knex;
```
bilts, line 9: Is this required by the update to the Knex library, where `knex.transaction` returns a `Knex`?

bilts (approved):

PR #81, opened 2021-10-12
Harmony 990

indiejames:
This PR combines HARMONY-990 and HARMONY-991. It sets up services.yml and some other things to support PODAAC CONCISE (including starting the service in Turbo), but it does not actually make CONCISE available to Harmony requests yet (that comes later in HARMONY-993, etc.).

You can test this locally using the bootstrap startup and verifying that the 'podaac-concise' pods are running with `kubectl get pods -n argo`. Or if you are brave do a sandbox deployment (you need the harmony-990 branch of harmon-ci-cd for that).


bilts (approved):

tasks/service-runner/config/concise-sidecar.yaml
```diff
@@ -0,0 +1,103 @@
...
+            requests:
+              memory: $PODAAC_CONCISE_REQUESTS_MEMORY
+          envFrom:
+          - configMapRef:
+              name: harmony-env
+          - secretRef:
+              name: harmony-secrets
+          env:
+            - name: TEXT_LOGGER
+              value: "false"
```
chris-durbin, line 33: Should read from env variable (probably all of the sidecars need updated)

tasks/service-runner/config/service-example-sidecar.yaml
```diff
@@ -101,4 +102,4 @@ spec:
       protocol: TCP
   selector:
     name: harmony-service-example
-  type: LoadBalancer
+  type: ClusterIP
```
chris-durbin, line 105: Why these changes from LoadBalancer to ClusterIP?

app/models/data-operation.ts
```diff
@@ -231,6 +233,15 @@ export default class DataOperation {
     this.decrypter = decrypter;
   }
 
+  /**
+   * Returns true if the operation is requesting subsetting
+   *
+   * @returns true if the operation requests subsetting
+   */
+  get shouldSubset(): boolean {
```
chris-durbin, line 241: This seems too generic to me. I think we need different subset parameters - variableSubset, temporalSubset, spatialSubset, etc.

config/services.yml
```diff
@@ -411,6 +411,73 @@ https://cmr.uat.earthdata.nasa.gov:
...
+          <<: *default-argo-env
+          STAGING_PATH: public/podaac/l2-subsetter-concise
+    collections:
+      - C1234208436-POCLOUD
+      - C1234208437-POCLOUD
+      - C1234208438-POCLOUD
+    capabilities:
+      # TODO implement this in HARMONY-993
+      concatenation: true
+      subsetting:
```
chris-durbin, line 465: This configuration doesn't look right to me - it says to chain together the L2 subsetter with concise, but says that it has no subsetting capabilities. It should have bbox, shapefile, or variable subsetting.

config/services.yml
```diff
@@ -411,6 +411,73 @@ https://cmr.uat.earthdata.nasa.gov:
...
+        bbox: false
+        variable: false
+      output_formats:
+        - application/netcdf # Incorrect mime-type, remove when no longer needed
+        - application/x-netcdf4
+    steps:
+      - image: !Env ${CMR_GRANULE_LOCATOR_IMAGE}
+      - image: !Env ${PODAAC_L2_SUBSETTER_IMAGE}
+        operations: ['subset']
```
chris-durbin, line 474: I think both operations and conditional should be more specific about the type of subsetting as mentioned in earlier comment.

app/models/services/base-service.ts
```diff
@@ -98,14 +100,19 @@ const conditionToOperationField = {
  *
  * @returns true if the workflow step is required
  */
-function stepRequired(step: ServiceStep, operation: DataOperation): boolean {
+function stepRequired(step: ServiceStep, operation: DataOperation, logger: Logger): boolean {
+  logger.info('OPERATION');
```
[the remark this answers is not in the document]
indiejames: done

app/models/services/base-service.ts
```diff
@@ -300,7 +310,9 @@ export default abstract class BaseService<ServiceParamType> {
     if (this.config.steps) {
       let i = 0;
       this.config.steps.forEach(((step) => {
-        if (stepRequired(step, this.operation)) {
+        this.logger.debug('STEP');
```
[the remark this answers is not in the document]
indiejames: removed

config/services.yml
indiejames: fixed

config/services.yml
indiejames: done

chris-durbin (approved):

PR #84, opened 2021-10-19
Harmony 988 - Work items that take too long are marked as failed

vinnyinverso:


app/server.ts
```diff
@@ -194,7 +196,16 @@ export function start(config: Record<string, string>): {
     workReaper.start();
   }
 
-  return { frontend, backend, workflowTerminationListener: listener, jobReaper, workReaper };
+  let workFailer;
+  if (config.startWorkFailer !== 'false') {
+    const failerConfig = {
+      logger: logger.child({ application: 'workflow-events' }),
```
chris-durbin, line 202: I assume we should call this application 'work-failer'.

env-defaults
```diff
@@ -147,6 +147,12 @@ WORK_REAPER_PERIOD_SEC=360
 # WorkItems and WorkflowSteps (in a terminal state) older than this many minutes are checked by the work reaper
 REAPABLE_WORK_AGE_MINUTES=1440
 
+# The time (in seconds) between invocations of the work failer service
+WORK_FAILER_PERIOD_SEC=360
+
+# WorkItems and associated job (in a running state) older than this many minutes are set to failed by the work failer
+FAILABLE_WORK_AGE_MINUTES=20
```
chris-durbin, line 154: I would make this a giant number by default so we don't accidentally fail legitimate but slow requests - I think we allowed steps in Argo workflows 4 hours by default, so maybe do the same here and use 240 minutes.

PR #90, opened 2021-11-03
Harmony 923 - Pass Color Map related URLs to backend services

vinnyinverso:
I'll proceed with the rest of the data-operation/README.md checklist (service lib updates, etc.) once I get approval on the schema updates. I wanted to get as much as possible out of the way while things are sorted out on CMR's end.

Note: I deleted `interface CmrVariable` because it wasn't used and tests still pass but if it's somehow needed I can revert that. Also I stubbed a test that will be implemented once CMR deploys to UAT and I can add some RelatedUrls to a variable (or have someone do that for me).


app/models/data-operation.ts
```diff
@@ -175,6 +203,25 @@ export interface SRS {
...
+ * @param relatedUrl - A CmrRelatedUrl
+ * @returns the equivalent HarmonyRelatedUrl 
+ */
+export function cmrRelatedUrlToHarmony(relatedUrl: CmrRelatedUrl): HarmonyRelatedUrl {
+  const harmonyRelatedUrl: HarmonyRelatedUrl = {
+    url: relatedUrl.URL,
+    urlContentType: relatedUrl.URLContentType,
+    type: relatedUrl.Type,
+  };
+  if (relatedUrl.Subtype) harmonyRelatedUrl.subtype = relatedUrl.Subtype;
```
chris-durbin, line 217: Wondering why we can't assign these potentially undefined/null fields and need to perform these checks.

Edit: I'm guessing the JSON passed in the message was including things like `{ 'Description': null }`.

app/models/data-operation.ts
```diff
@@ -298,17 +345,30 @@ export default class DataOperation {
...
-    })) : undefined;
+    const variables = vars ? vars.map(({ umm, meta }) => {
+      const schemaVar: HarmonyVariable = {
+        id: meta['concept-id'],
+        name: umm.Name,
+        fullPath: umm.Name,
+      };
+      if (umm.RelatedURLs) {
+        const filteredUrls = umm.RelatedURLs
+          .filter((url) => varUrlTypes.includes(url.Type as CmrRelatedUrlType))
```
chris-durbin, line 364: So we're ignoring any urls that do not include a `Type` attribute? What's the rationale?

test/util/cmr.ts
```diff
@@ -11,5 +11,10 @@ describe('util/cmr', function () {
       const variables = await getVariablesByIds(ids, '');
       expect(variables.length).to.eql(1);
     });
+
+    it('response contains related URLs when the CMR variable has them', async function () {
```
chris-durbin, line 15: You can use `xit` to tell it not to run. Normally we would include a ticket number in the comment to explain why it is not running.

test/models/data-operation.ts
```diff
@@ -241,9 +241,22 @@ describe('DataOperation', () => {
       url: 'https://example.com/foo',
       temporal: {},
     }];
+    const relatedUrl = {
```
chris-durbin, line 244: I think this test should include a couple of other URLs - one with a Type that isn't 'Color Map' and one that is missing the Type field altogether and then add verifications for both URLs that should be included and URLs that should not be included.

app/models/data-operation.ts
vinnyinverso: Yep -- I was thinking it'd be best to not end up with properties like { 'Description': undefined } in the final data operation.

app/models/data-operation.ts
vinnyinverso: It's actually not filtering based on the existence of the Type property. It's just ignoring any urls that don't have Type = any of the types specified in the `varUrlTypes` argument. The thinking was to make the default argument be `[CmrRelatedUrlType.ColorMap]` since that's the only behavior that we want for now.

test/util/cmr.ts
vinnyinverso: Updated.

test/models/data-operation.ts
vinnyinverso: Updated, thanks!

app/models/data-operation.ts
bilts: I think it's still redundant, though:
```
> JSON.stringify({foo: undefined, bar: null})
'{"bar":null}'
```

app/models/data-operation.ts
bilts: I could go either way on this, but I could see this tripping us up in the future if we assume "Oh yeah, we're passing on related URLs, so you can put that info there"

Also if we want to pass on the minimum, we technically only want to pass on ColorMap URLs if they have a subtype of "GIBS"

app/models/data-operation.ts
```diff
@@ -175,6 +203,25 @@ export interface SRS {
   epsg?: string;
 }
 
+/**
+ * Transform a related URL from CMR to a Harmony related URL that can be used in the DataOperation.
+ * @param relatedUrl - A CmrRelatedUrl
+ * @returns the equivalent HarmonyRelatedUrl 
+ */
+export function cmrRelatedUrlToHarmony(relatedUrl: CmrRelatedUrl): HarmonyRelatedUrl {
+  const harmonyRelatedUrl: HarmonyRelatedUrl = {
```
bilts, line 212: Let's say there's a future where we just want to pass on the UMM-Var object in its entirety.  Are we making it harder for ourselves by lower-camel-casing all the fields?  I'm kind of torn on whether we should just pass the object as we receive it vs trying to interpret it.

(Also, :shaking-angry-fist: at whomever decided Var fields should be upper-camel-case)

bilts (commented):
I think this is probably merge-able as-is.  A few minor comments, though.

app/models/data-operation.ts
vinnyinverso: Interesting point. I was only thinking of the current conventions of the JSON schema and the service-lib, and how not following those conventions might confuse things downstream.

app/models/data-operation.ts
vinnyinverso: I'm happy to get rid of it if that's more ideal.

app/models/data-operation.ts
vinnyinverso: > Also if we want to pass on the minimum, we technically only want to pass on ColorMap URLs if they have a subtype of "GIBS"

Good to know. 

My thinking was that in case someone finds a use case that results in the relatedUrls field being abused, we'd not want the data operation to blow up without our knowledge.

app/models/data-operation.ts
vinnyinverso: Got rid of this code.

app/models/data-operation.ts
vinnyinverso: Passing all related URLs now.

app/models/data-operation.ts
vinnyinverso: Got rid of this function.

chris-durbin (approved):

indiejames (approved):

PR #93, opened 2021-11-04
HARMONY-959: Add support for indicating aggregated output on a workfl…

indiejames:
Adds `hasAggregatedOutput` column to workflow_steps table and sets this to true for steps that have the `concatenate` operation


app/models/services/base-service.ts
```diff
@@ -88,6 +88,23 @@ const conditionToOperationField = {
...
+  'concatenate',
+];
+
+/**
+ *  Returns true if the workflow step aggregates output from the previous step
+ * (and therefore must wait for all output before executing)
+ * @param step - the step in a workflow
+ * @returns true if the step is an aggregating step, false otherwise
+ */
+function _stepHasAggregatedOutput(step: ServiceStep): boolean {
```
chris-durbin, line 104: Nit: I thought our convention was to only prefix with _ on methods within a class. At the top level our indication a function was private was whether or not it was exported.

vinnyinverso (approved):

app/models/services/base-service.ts
```diff
@@ -318,6 +335,7 @@ export default abstract class BaseService<ServiceParamType> {
               this.config.data_operation_version,
               step.operations || [],
             ),
+            hasAggregatedOutput: _stepHasAggregatedOutput(step),
```
chris-durbin, line 338: I think we're marking this flag on the step that needs to aggregate all the outputs from the prior step's work items. In that case should it be `aggregateInputs`? Well I guess when the step completes it does have aggregated outputs... Names are hard, maybe yours is good.

chris-durbin (approved):

app/models/services/base-service.ts
indiejames: Oh - for some reason I thought it was anything that wasn't exported. I'll change it.

app/models/services/base-service.ts
indiejames: Yeah - I was dithering about that as well, but the ticket says "When services set the aggregated output flag in services.yml, the saved job record indicates its aggregated output status"

app/models/services/base-service.ts
indiejames: Done.

chris-durbin: I want to say this convention was your (previous James') idea, and we all liked it better.

bilts (approved):

PR #95, opened 2021-11-08
Harmony 960 wait for all inputs to aggregation

indiejames:
This ticket has a single AC:

- When a step in a job is marked as an aggregated output job, Harmony does not invoke it until all prior steps complete

The work item that is generated for the aggregation step currently does not use all the output from the preceding step (although it does wait for all the output). That is the subject of HARMONY-961.

### DIFF
--- a/app/backends/workflow-orchestration.ts
+++ b/app/backends/workflow-orchestration.ts
@@ -7,7 +7,7 @@ import { readCatalogItems } from '../util/stac';
 import HarmonyRequest from '../models/harmony-request';
 import { Job, JobStatus } from '../models/job';
 import JobLink from '../models/job-link';
-import WorkItem, { getNextWorkItem, WorkItemStatus, updateWorkItemStatus, getWorkItemById, workItemCountForStep } from '../models/work-item';
+import WorkItem, { getNextWorkItem, WorkItemStatus, updateWorkItemStatus, getWorkItemById, workItemCountForStep, WORK_ITEM_SUCCESS_STATUSES } from '../models/work-item';
 import WorkflowStep, { getWorkflowStepByJobIdStepIndex } from '../models/workflow-steps';
 
 const MAX_TRY_COUNT = 1;
@@ -118,6 +118,7 @@ async function createNextWorkItems(
     return createAggregatingWorkItem(tx, currentWorkItem, nextStep, results);
   } else {
     // Create a new work item for each result using the next step
+    // FIXME Do this as a bulk insertion when working NO GRANULES LIMIT TICKET (HARMONY-276)
     for await (const result of results) {
       const newWorkItem = new WorkItem({
         jobID: currentWorkItem.jobID,
@@ -190,7 +191,7 @@ export async function updateWorkItem(req: HarmonyRequest, res: Response): Promis
         tx,
         workItem.jobID,
         workItem.workflowStepIndex,
-        WorkItemStatus.SUCCESSFUL,
+        WORK_ITEM_SUCCESS_STATUSES,
       );
       const thisStep = await getWorkflowStepByJobIdStepIndex(
         tx,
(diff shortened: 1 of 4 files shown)


app/backends/workflow-orchestration.ts
```diff
@@ -80,27 +80,54 @@ async function _handleWorkItemResults(
   }
 }
 
+/**
+ *
```
chris-durbin, line 84: Add documentation

app/models/services/base-service.ts
```diff
@@ -313,6 +313,23 @@ export default abstract class BaseService<ServiceParamType> {
     return workItems;
   }
 
+  /**
+   * Return the number of work items that should be created for a give step
```
chris-durbin, line 317: given

chris-durbin (approved):

app/backends/workflow-orchestration.ts
indiejames: Done.

app/models/services/base-service.ts
indiejames: Fixed

test/workflow-orchestration.ts
```diff
@@ -6,7 +6,90 @@ import { Job, JobStatus } from '../app/models/job';
...
+    await updateWorkItem(this.backend, savedWorkItem);
+  });
+
+  this.afterEach(async function () {
+    await db.table('work_items').del();
+  });
+
+  describe('and a work item for the first step is completed', async function () {
+    describe('and it is not the last work item for the step', async function () {
+      it('the next step has no work available', async function () {
```
bilts, line 64: Grammar nit: `it('does not supply work for the next step')` or something like that.  Similar for the other `it`

app/backends/workflow-orchestration.ts
```diff
@@ -80,27 +80,55 @@ async function _handleWorkItemResults(
...
-      serviceID: nextStep.serviceID,
-      status: WorkItemStatus.READY,
-      stacCatalogLocation: result,
-      workflowStepIndex: nextStep.stepIndex,
-    });
+  if (nextStep.hasAggregatedOutput) {
+    return createAggregatingWorkItem(tx, currentWorkItem, nextStep, results);
+  } else {
+    // Create a new work item for each result using the next step
+    for await (const result of results) {
```
bilts, line 121: Not for your PR to fix, but it strikes me that for no granule limit, we're probably going to want/need some sort of bulk insert (`INSERT INTO ... SELECT ...`)

bilts: In general it'd be good to start making note of any time we're iterating over all results in prep for that.

app/backends/workflow-orchestration.ts
```diff
@@ -158,25 +186,29 @@ export async function updateWorkItem(req: HarmonyRequest, res: Response): Promis
         workItem.workflowStepIndex + 1,
       );
 
+      const successWorkItemCount = await workItemCountForStep(
+        tx,
+        workItem.jobID,
+        workItem.workflowStepIndex,
+        WorkItemStatus.SUCCESSFUL,
```
bilts, line 193: Another now or later thing, we have a ticket that arrived in the backlog for allowing a "success with warnings". It makes me think that this should really be some sort of function that returns a list of states that constitute success, or it should be all wrapped up in a `successfulWorkItemCount` function or something so we can more freely change it.

You don't need to do it as part of this PR, but it may be an overall time savings to write the function now vs trying to find that we need it here later.

test/workflow-orchestration.ts
indiejames: Done.

app/backends/workflow-orchestration.ts
indiejames: I added a FIXME comment with reference to NO GRANULE LIMIT ticket (HARMONY-276)

app/backends/workflow-orchestration.ts
indiejames: I changed  the optional `status` parameter of `workItemCountForStep` to take either a string or an array of strings to allow us to have multiple 'successful' `WorkItemStatus`s. So for now we pass it `[WorkItemStatus.SUCCESSFUL]`, but later on we might pass it `[WorkItemStatus.SUCCESSFUL, WorkItemStatus.SUCCESSFUL_WITH_WARNINGS]`.

PR #96, opened 2021-11-10
DAS-1272 - Add production HOSS entry in services.yml.

owenlittlejohns:
This PR does two things:

1) Adds an entry for HOSS in the production section of `services.yml`. I've just copied-and-pasted the majority of the UAT entry, minus collection and service concept IDs. (I'm guessing some of the Argo related information will be removed in the very near future)
2) Removes the listed collections for the Harmony services administered by Data Services in `services.yml` (UAT: HOSS, Swath Projector, Variable Subsetter and MaskFill). I've checked that the UMM-C to UMM-S associations exist for all collections that were listed in `services.yml`.


config/services.yml
```diff
@@ -573,10 +593,7 @@ https://cmr.uat.earthdata.nasa.gov:
     umm_s:
       - S1240682712-EEDTEST
       - S1241070355-GES_DISC # GESDISC_HOSS
-    collections:
-      - C1238392622-EEDTEST  # RSSMIF16D test collection UAT
-      - C1222931739-GHRC_CLOUD  # RSSMIF16D UAT
-      - C1238621141-POCLOUD # GHRSST Level 4 MUR Global Foundation Sea Surface Temperature Analysis (v4.1)
```
bilts, line 579: Just to verify: are you responsible for this association, or is PO.DAAC depending on it somehow?

config/services.yml
```diff
@@ -249,6 +249,36 @@ https://cmr.earthdata.nasa.gov:
...
+        <<: *default-argo-params
+        template: var-subsetter
+        template_type: chaining
+        image: !Env ${VAR_SUBSETTER_IMAGE}
+        image_pull_policy: !Env ${VAR_SUBSETTER_IMAGE_PULL_POLICY}
+        parallelism: !Env ${VAR_SUBSETTER_PARALLELISM}
+        env:
+          <<: *default-argo-env
+          STAGING_PATH: public/sds/HOSS
+    # umm_s:
```
chris-durbin, line 267: Has the service record not been created yet?

I'm sure you realize it, but as is we won't be able to use HOSS for any collections in production. I'm wondering if there might be a test collection in production that we'd want to make available - something like we do with harmony-service-example to verify the service is working correctly in production.

config/services.yml
```diff
@@ -499,15 +529,7 @@ https://cmr.uat.earthdata.nasa.gov:
...
-    collections:
-      # DO NOT UNCOMMENT THE FOLLOWING LINE - this collection is added via UMM-S and is left here
-      # to prevent re-adding. This is needed for tests related to HARMONY-875
-      # - C1233860183-EEDTEST
-      - C1233800302-EEDTEST
-      - C1221286860-NSIDC_TS1
-      - C1226288788-EDF_DEV07
-      - C1233147317-EEDTEST
-      - C1233147319-EEDTEST
+    collections: []
```
chris-durbin, line 532: Note to us (you don't need to change this PR) - I'm thinking we may as well change the code to make collections optional and treat it as empty if missing.

config/services.yml
nasajay: I have the same question.

config/services.yml
owenlittlejohns: I couldn't remember who added this association, so I took a quick look at the git history, and it looks like Mike Gangl added this line. I've preserved the association with a UMM-S to UMM-C association in CMR (against the EEDTEST version of the UMM-S record), but maybe PODAAC might want their own copy in the future. (I'm not sure I've completely answered the question there)

The other thing to note is that when I tried a request to test the association was good, it first went to the ASF GDAL Subsetter, as the collection is also registered there, and that service is higher in the `services.yml`.

config/services.yml
owenlittlejohns: Good point, Chris.

To my knowledge a HOSS UMM-S entry does not yet exist in the GES-DISC production CMR provider. There are two objectives associated with getting HOSS in production and supporting the GPM/IMERG collection, one for Data Services and the other for GES-DISC; I think the UMM-S record should fall under the GES-DISC objective. The Data Services team don't have the requisite permissions to make this UMM-S record in the GES-DISC provider, and don't have an EEDTEST provider in production to make a UMM-S record of our own.

I was hoping by popping this comment in the `services.yml` it would make it super easy for a GES-DISC developer (maybe @nasajay in a couple of weeks from now) to pop the UMM-S concept ID here when the record exists.

I think it would be _really_ nice to have some sort of test collection, though! I'd be interested in talking to you guys about how you got around the lack of an EEDTEST provider for that collection for harmony-service-example.

config/services.yml
bilts: Would you mind either re-adding the collection so PO.DAAC can manage it as they see fit or confirming with Mike that it can be removed?

owenlittlejohns: Sure - I'll check with Mike about what he wants done with this collection.

config/services.yml
owenlittlejohns: Yeah - my first guess was to remove `collections` altogether, but that didn't work so well (hence the second commit). That would be a nice change. It definitely felt nice to clean out some lines from this file, though!

config/services.yml
chris-durbin: In production there's a provider with collections and granules that are hidden from normal users via ACLs - XYZ_PROV. 

We have a test collection there that we use with harmony-service-example: C1756916832-XYZ_PROV.

I'm fine with merging this in as is, just making sure you realized there wouldn't be any way to verify HOSS working with harmony in production without at least one collection.

owenlittlejohns: `XYZ_PROV` sounds promising! I'll check within our team about making some test data available from this provider. If that's something we want to pursue (and it sounds great to me) I'll probably then coordinate with you offline about getting the correct access to the provider so we can do that. Thanks, Chris!

In the meantime, though, my inclination is to forge on ahead with this PR, albeit with a production HOSS that is currently non-verifiable until we have  a UMM-S and associated collection.

owenlittlejohns: - [ ] Add `XYZ_PROV` UMM-S record.

bilts (approved):

PR #97, opened 2021-11-15
HARMONY-969: Update services.yml to add turbo steps for all definitions

chris-durbin:
I removed the `podaac/ps3` and `podaac/netcdf-converter` service definitions as part of this change. I'm awaiting confirmation from Frank before merging in. 

All of the services other than maskfill are now configured for turbo (maskfill will be added in a separate PR).

I also updated a bunch of dependencies in this PR. I moved all the types definitions to dev dependencies and tried to make sure production vs. dev dependencies was correct. I'll do a sandbox deployment to confirm no issues.


vinnyinverso (approved):

bilts (approved):
Yay for paying down maintenance!

PR #98, opened 2021-11-15
Harmony 961 aggregate inputs

indiejames:
Creates a STAC catalog that contains all the output 'item' links from the STAC catalogs that were generated by the previous step.


indiejames:
Frank verified output from manual test.

app/backends/workflow-orchestration.ts
```diff
@@ -81,23 +84,91 @@ async function _handleWorkItemResults(
 }
 
 /**
- * Creates a work item that uses all the output of the previous step.
+ * Read a STAC catalog and return the item links. This does not handle sub-catalogs. This function 
+ * makes assumptions based on the Harmony STAC directory layout for services inputs/outputs and
+ * is only intended to be used when aggregating service outputs into a single catalog.
+ * @param catlogPath - the path to the catalog
```
chris-durbin, line 90: catalogPath typo

app/backends/workflow-orchestration.ts
```diff
@@ -81,23 +84,91 @@ async function _handleWorkItemResults(
...
+        const fullPath = path.join(directory, filePath);
+        const newLinks = await getLinksFromCatalog(fullPath);
+        itemLinks.push(...newLinks);
+      }
+      processedItemCount++;
+    }
+    page++;
+  }
+
+  // create a STAC catalog using `catalogLinks`
```
chris-durbin, line 148: I believe the query-cmr task has a bunch of STAC utility functions that include creating and writing out a STAC catalog. It might be worth trying to move them into the main harmony code and refactoring to use them. Probably a separate maintenance ticket rather than a comment to address this PR.

app/backends/workflow-orchestration.ts
```diff
@@ -81,23 +84,91 @@ async function _handleWorkItemResults(
...
+  tx: Transaction, currentWorkItem: WorkItem, nextStep: WorkflowStep,
 ): Promise<void> {
-  // TODO use ALL outputs from the prior step here (HARMONY-961)
-  // just use the first result of the current output for now
+  const itemLinks: StacItemLink[] = [];
+  // get all the previous results
+  const workItemCount = await workItemCountForStep(tx, currentWorkItem.jobID, nextStep.stepIndex - 1);
+  let page = 1;
+  let processedItemCount = 0;
+  while (processedItemCount != workItemCount) {
```
chris-durbin, line 129: Possible infinite loop if this function is called before all of the results are available or if one of the database calls doesn't return all the items we expect, or some bug. We probably want some kind of safe guard to throw an error to prevent the infinite loop.

app/backends/workflow-orchestration.ts
```diff
@@ -81,23 +84,91 @@ async function _handleWorkItemResults(
...
+    const prevStepWorkItems = await getWorkItemsByJobIdAndStepIndex(tx, currentWorkItem.jobID, nextStep.stepIndex - 1, page);
+    for (const workItem of prevStepWorkItems.workItems) {
```
chris-durbin, line 131: Having to read all of the catalogs puts a some kind of upper limit on the number of granules we can reasonably handle for an aggregation step. Something we should keep in mind for the no granule limits epic.

app/models/work-item.ts
```diff
@@ -260,6 +260,37 @@ export async function getWorkItemsByJobId(
...
+ * @param sortOrder - orderBy string (desc or asc)
+ *
+ * @returns A promise with the work items array
+ */
+export async function getWorkItemsByJobIdAndStepIndex(
+  tx: Transaction,
+  jobID: string,
+  workflowStepIndex: number,
+  currentPage = 0,
+  perPage = 10,
```
chris-durbin, line 279: I'm not sure if 10 is our normal default for paging, but since calling this function is internal code rather than some user request, and in the place we're using it now we're going to want to iterate through all of the items I think we'd want to use a larger page size. I'd either increase the default here to something like 100 or 500, or pass in a larger value when we call this function rather than using the default.

env-defaults
```diff
@@ -368,7 +368,8 @@ MAX_POST_FILE_PARTS=100
 
 # The location on the file system to save the work items that will be mounted into the
 # containers for any services to read and save STAC catalogs.
-HOST_VOLUME_PATH=
+# Defaulting to /tmp to allow tests to pass in CI
+HOST_VOLUME_PATH=/tmp
```
chris-durbin, line 372: Using /tmp doesn't work with Mac OS and Docker, and for a new developer setting up harmony it would cause problems that might be difficult to debug. I think that's enough reason to not set it in the defaults, and instead override it in the test env.ts file if it just needs to be set for the tests.

test/helpers/work-items.ts
```diff
@@ -152,3 +155,54 @@ export function getWorkForService(app: Application, serviceID: string): Test {
...
+      },
+    ],
+    description: 'Fake STAC catalog',
+  };
+  await fs.writeFile(path.join(outputDir, 'catalog0.json'), JSON.stringify(catalog, null, 4));
+
+  // create a fake STAC item
+  // NOTE: this is not a valid STAC item because it is missing fields we don't need for our tests
+  const item = {
+    tac_version: '1.0.0-beta.2',
```
chris-durbin, line 197: stac_version

test/workflow-orchestration.ts
```diff
@@ -49,14 +53,14 @@ describe('When a workflow contains an aggregating step', async function () {
     savedWorkItem.status = WorkItemStatus.SUCCESSFUL;
     savedWorkItem.results = [
       'test/resources/worker-response-sample/catalog0.json',
-      'test/resources/worker-response-sample/catalog1.json',
```
chris-durbin, line 52: I'm not quite following these tests - are we only testing with a single work item and catalog? If so we should modify the tests to test with multiple work items and catalogs.

chris-durbin (changes requested):
Looks good - just requesting changes for `env-defaults` and clarifying my question on tests.

app/backends/workflow-orchestration.ts
indiejames: Fixed

app/backends/workflow-orchestration.ts
indiejames: Yes, as I have mentioned before just using STAC to communicate between steps is limiting since JSON must be entirely read into memory to be parsed. We should think about using a simpler format that can be read (and parsed) line by line.

app/models/work-item.ts
indiejames: Increased to 100, but that seems like an arbitrary number to me (as does 10) and will obviously use more memory than 10.

test/workflow-orchestration.ts
indiejames: We have two work items (each producing one catalog for one granule, which is what turbo does) followed by an aggregation step.

env-defaults
indiejames: Fixed

test/helpers/work-items.ts
indiejames: Fixed

app/backends/workflow-orchestration.ts
indiejames: Wrote HARMONY-1024 for this.

app/backends/workflow-orchestration.ts
```diff
@@ -81,23 +85,99 @@ async function _handleWorkItemResults(
...
-  // just use the first result of the current output for now
+  const itemLinks: StacItemLink[] = [];
+  // get all the previous results
+  const workItemCount = await workItemCountForStep(tx, currentWorkItem.jobID, nextStep.stepIndex - 1);
+  let page = 1;
+  let processedItemCount = 0;
+  while (processedItemCount < workItemCount) {
+    const prevStepWorkItems = await getWorkItemsByJobIdAndStepIndex(tx, currentWorkItem.jobID, nextStep.stepIndex - 1, page);
+    // guard against failure case where we cannot retrieve all items - THIS SHOULD NEVER HAPPEN
+    if (prevStepWorkItems.workItems.length < 1) break;
```
chris-durbin, line 133: Throw an exception instead of break? I think at a minimum we should log something if this were to occur.

chris-durbin (approved):

app/backends/workflow-orchestration.ts
indiejames: I break because I added a check below that will throw an exception if the number of processed items is not what we expect. Better to check there to catch any situation where we did not get all the worked items.

indiejames:
I just pushed a fix for a problem with chained aggregations that I saw when preparing for the tech meeting.

app/backends/workflow-orchestration.ts
bilts: I think this is more likely in the future.  We have cases in the pipeline where a variable subset is requested but some data files don't contain the variable, which may result in this case being triggered, and a success-with-warning condition.

app/backends/workflow-orchestration.ts
bilts: Or page-by-page, with one catalog per page

bilts: So on that note, this is 1000% going to come up with the rest of no granule limit, and we've already done the paging parts of that multi-epic.  Given this introduces a paging issue, I think we at least need a ticket to tackle it next PI, if not entirely handling the paging now

bilts (approved):

PR #102, opened 2021-11-19
HARMONY-982: Add example requests to workload for new services. Updated so that all requests can now use turbo.

chris-durbin:
Added new requests to the workload driver for new harmony services and configured all of the requests to be able to use turbo. Ran tests with 10 concurrent requests with 2 m5.xlarge nodes in sandbox and 2 pods configured for each turbo service. No issues observed with CPU or memory. In all cases the turbo requests had better average performance than with Argo. When a service had many concurrent requests the turbo performance did degrade some, but this configuration should be sufficient until we begin autoscaling pods based on the load on a service.


workload/locustfile.py
```diff
@@ -220,15 +255,46 @@ def netcdf_to_zarr_single_granule_argo(self):
     def netcdf_to_zarr_single_granule_turbo(self):
         self._netcdf_to_zarr_single_granule(True)
 
-    @tag('netcdf-to-zarr', 'async', 'zarr', 'argo', 'memory', 'slow')
-    @task(1)
-    def netcdf_to_zarr_large_granule_argo(self):
-        self._netcdf_to_zarr_large_granule()
+    # @tag('new', 'async', 'concise', 'argo')
```
indiejames, line 258: Is this commented out code being left here for a reason, or should it be deleted?

indiejames (approved):

hailiangzhang (approved):

PR #109, opened 2021-12-01
HARMONY=976: Add 'publish' to github actions for service-runner

indiejames:
Simple change to add a publish action to package.json for the service runner so we can publish it to DockerHub


vinnyinverso (approved):

chris-durbin (approved):

chris-durbin:
In the github actions file we have:
```
          # build and publish query-cmr (cmr granule locator)
          VERSION="${VERSION}" lerna run build
          VERSION="${VERSION}" lerna run publish
```
The `lerna run publish` command will run publish for service-runner too, so probably update the comment there to say build and publish all tasks instead of just query-cmr.

PR #112, opened 2021-12-02
Add prod entry for concise and chain

frankinspace:


chris-durbin (approved):
I tested out by pointing my local environment at production. I had to make one change due to a harmony bug to just add an empty collections list to the `podaac/concise` configuration.

config/services.yml
```diff
@@ -258,6 +258,90 @@ https://cmr.earthdata.nasa.gov:
...
+      <<: *default-argo-config
+      params:
+        <<: *default-argo-params
+        image: !Env ${PODAAC_CONCISE_IMAGE}
+        image_pull_policy: !Env ${PODAAC_CONCISE_IMAGE_PULL_POLICY}
+        template_type: chaining
+        env:
+          <<: *default-argo-env
+          STAGING_PATH: public/podaac/l2-subsetter-concise
+    collections:
```
chris-durbin, line 300: Seems like moving this to use `umm_s` as well is a good idea in the future.

vinnyinverso (approved):

config/services.yml
frankinspace: I wasn't sure if that was possible with a chain?

chris-durbin: I thought there was an idea that each chain would be its own UMM-S record, but not sure where that ended up in practice. @bilts or @asteiker would know best.

PR #114, opened 2021-12-02
Harmony 1000 - Set up prometheus so that metrics can be scraped from backend service pods

vinnyinverso:
This PR provides the k8s config setup for Prometheus and shows how metrics can be scraped once HARMONY-999 is finished. There will be further tickets for setting up the custom metrics server and HPAs. (CI-CD code will be added once the other pieces are in place.)

To test out the acceptance criteria:
> When a metric is provided by a manager container in a service pod and I query that metric in Prometheus while specifying a serviceID, it provides the correct value for the metric

Add the below to tasks/service-runner/app/routers/router.ts

```javascript
  let gauge = 0;
  result.get('/metrics', (function (req, res) {
    gauge += 1;
    const prom_metric = 
   `# HELP custom_metric An example of a custom metric, using the gauge type.
    # TYPE custom_metric gauge
    custom_metric{service_id="query-cmr-latest"} ${gauge}`;
    return res.send(prom_metric);
  }));
```
 
Under ‘labels’ (at the same level of indentation), around line 15 in query-cmr-sidecar.yaml, add the following

```yaml
annotations:
  prometheus.io/port: "5000"
  prometheus.io/scrape: "true"
```
 
Pull the prometheus image and run the k8s deployment

```console
docker pull quay.io/prometheus/prometheus:v2.0.0
bin/start-prometheus
```

Copy the pod name to the port-forward command

```console
kubectl get pods -n monitoring
kubectl port-forward pod/prometheus-b8c445754-5lwms 9090:9090 -n monitoring
```

Run Harmony

```console
bin/start-argo
bin/deploy-services
npm run start-dev
 ```

Go to http://localhost:9090/targets to view the scraping targets (should see query-cmr)
Go to http://localhost:9090/graph, then enter`custom_metric{service_id="query-cmr-latest"}` in the expression input, click execute, then click the graph tab.

### CODE REVIEW

File: config/prometheus/monitoring-namespace.yaml [hunk 1]
```diff
@@ -0,0 +1,4 @@
+apiVersion: v1
```
chris-durbin, line 1: I think I'd prefer a single file containing all of the prometheus config (just separate resources with:
```
---
```
See `local-argo-deployment.yml` as an example.

The reason being that I'd always want to deploy all of them or none of them. It also makes it easy for me to delete the resources with one command (`kubectl delete -f ./config/prometheus-resources.yaml`)

bin/start-prometheus
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
```
bilts, line 1: It seems like this script should be either (ideal) rolled into other scripts as part of creating / updating a Harmony environment or otherwise documented as a step in the README.

bilts (commented):
If prometheus is ever meant to be accessed directly locally, it'd be nice to document its endpoint exists.  I believe one of the "start" scripts emits URLs at the end, so my preference would be to add it there.

bin/start-prometheus
```diff
@@ -0,0 +1,6 @@
...
+
+kubectl apply -f ./config/prometheus/monitoring-namespace.yaml
+kubectl apply -f ./config/prometheus/prometheus-config.yaml
+kubectl apply -f ./config/prometheus/prometheus-roles.yaml
```
chris-durbin, line 5: I believe creating the prometheus-roles is throwing this warning:

>Warning: resource serviceaccounts/default is missing the kubectl.kubernetes.io/last-applied-configuration annotation which is required by kubectl apply. kubectl apply should only be used on resources created declaratively by either kubectl create --save-config or kubectl apply. The missing annotation will be patched automatically.

chris-durbin (approved):
Since this is just adding a new script that isn't tied into anything yet I'm good with merging it in.

bin/start-prometheus
indiejames: I'm actively making changes to the deployment scripts to fix some Harmony-in-a-box things regarding turbo, so it would be better for me if these don't get rolled into existing scripts right now.

hailiangzhang:
@vinnyinverso , why are you suggesting step wisely increasing `gauge` for each `metric` call? Or this is just a dummy example?
```
  let gauge = 0;
  result.get('/metrics', (function (req, res) {
    gauge += 1;
    const prom_metric = 
   `# HELP custom_metric An example of a custom metric, using the gauge type.
    # TYPE custom_metric gauge
    custom_metric{service_id="query-cmr-latest"} ${gauge}`;
    return res.send(prom_metric);
  }));
```

vinnyinverso:
> @vinnyinverso , why are you suggesting step wisely increasing `gauge` for each `metric` call? Or this is just a dummy example?
> 
> ```
>   let gauge = 0;
>   result.get('/metrics', (function (req, res) {
>     gauge += 1;
>     const prom_metric = 
>    `# HELP custom_metric An example of a custom metric, using the gauge type.
>     # TYPE custom_metric gauge
>     custom_metric{service_id="query-cmr-latest"} ${gauge}`;
>     return res.send(prom_metric);
>   }));
> ```

Yep -- that's just a dummy example so that one can test out the acceptance criteria.

config/prometheus/prometheus-roles.yaml
```diff
@@ -0,0 +1,39 @@
...
+metadata:
+  name: prometheus
+rules:
+- apiGroups: [""]
+  resources:
+  - nodes
+  - nodes/proxy
+  - services
+  - endpoints
+  - pods
```
hailiangzhang, line 12: If not too hard, maybe we can restrict the permission to the resources that are needed?

config/prometheus/prometheus-roles.yaml
```diff
@@ -0,0 +1,39 @@
...
+  verbs: ["get", "list", "watch"]
+- apiGroups:
+  - extensions
+  resources:
+  - ingresses
```
hailiangzhang, line 17: Same as above, do we need give prometheus the ingresses resources access? (not sure exactly what it means, but seems related to NGINX).

config/prometheus/prometheus-roles.yaml
```diff
@@ -0,0 +1,39 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
```
hailiangzhang, line 2: Sorry if I am getting too picky, but we are only monitoring resources in `argo` namespace right? So should we use `Role` instead of `ClusterRole`?

config/prometheus/prometheus-config.yaml
```diff
@@ -0,0 +1,34 @@
...
+  name: prometheus-config
+  namespace: monitoring
+data:
+  prometheus.yml: |
+    global:
+      scrape_interval:     15s
+    scrape_configs:
+      - job_name: 'kubernetes-pods'
+        kubernetes_sd_configs:
+        - role: pod
```
hailiangzhang, line 13: Great, I only see `pod` in configMap, so I guess we have another reason to narrow the access in `prometheus-roles.yaml`:)

hailiangzhang:
@vinnyinverso , not sure why, but it looks like that my prometheus deployment was able to scrape the pod only once. Basically the UI can always show up, but the target is always down:
image
It did work once (show as "UP") somehow, but it's always shown as "DOWN" now, even after I started everything from scratch.

config/prometheus/prometheus-roles.yaml
vinnyinverso: Good idea.

config/prometheus/prometheus-roles.yaml
vinnyinverso: Yup I'll look into this, thank you.

bin/start-prometheus
vinnyinverso: 👍  I'm gonna leave the scripting for later tickets (once the metrics server and HPA tickets are complete since they'll likely all get deployed together).

config/prometheus/prometheus-roles.yaml
vinnyinverso: I still think this is a good idea, however much of the docs that I'm reading seem to suggest the same setup as is. I think I'll leave it for now unless we can pinpoint exactly what should be changed.

https://github.com/prometheus/prometheus/blob/main/documentation/examples/rbac-setup.yml
https://github.com/bibinwilson/kubernetes-prometheus/blob/master/clusterRole.yaml

config/prometheus/prometheus-roles.yaml
vinnyinverso: No worries. I think we're using the CusterRole so that Prometheus can monitor across namespaces.

hailiangzhang:
@vinnyinverso , I saw you have merged the yaml files into one, but couldn't find `bin/start-prometheus` (which is supposed to be getting changed)... did I miss something?

vinnyinverso:
> @vinnyinverso , I saw you have merged the yaml files into one, but couldn't find `bin/start-prometheus` (which is supposed to be getting changed)... did I miss something?

@hailiangzhang I just deleted that, since we can use `kubectl apply -f config/prometheus.yaml` now that it's in the single file.

hailiangzhang:
> > @vinnyinverso , I saw you have merged the yaml files into one, but couldn't find `bin/start-prometheus` (which is supposed to be getting changed)... did I miss something?
> 
> @hailiangzhang I just deleted that, since we can use `kubectl apply -f config/prometheus.yaml` now that it's in the single file.

Got it!

hailiangzhang (approved):

PR #116, opened 2021-12-07
Harmony 1030

indiejames:
Use argo granule locator instead of turbo for no-op service even if turbo mode is requested. This is necessary because the no-op service cannot handle the scroll-id sent by the new granule locator.


indiejames:
Ordinarily I would have created a test that uses the collection/granule from the bug ticket, but that collection is not publicly available, so I had to use one that was.

chris-durbin (approved):

vinnyinverso (approved):

PR #120, opened 2021-12-13
Harmony 968 default to turbo

indiejames:
Make turbo the default, fix tests broken by same, and update README


chris-durbin:
It looks like a lot of fixtures were added but none removed. I'm thinking we probably want to remove all the fixtures and regenerate only the needed ones. I think there's only one that you need to keep around because it won't be regenerated correctly - 158890249394854233-manually-updated.

indiejames:
> 

Done

chris-durbin (approved):
Deleted my argo namespace and then ran `bin/start-postgres-localstack` and `bin/deploy-services` then requests completed successfully and were sent to turbo by default. I then ran the `bin/start-argo` script and ran requests with `turbo=false` and they completed successfully and were sent to argo.

bin/start-postgres-localstack
```diff
@@ -0,0 +1,111 @@
+#!/bin/bash
+
+# DEPRECATED
```
vinnyinverso, line 3: Might want to remove the #DEPRECATED comment since this file is not deprecated?

vinnyinverso (approved):

README.md
```diff
@@ -137,22 +130,17 @@ If you modify the `services.yml` file Harmony will need to be restarted. You can
...
+1. Build the image for your service
+2. Create a sidecar YAML file in harmony/tasks/service-runner/config by copying the 
+   `netcdf-to-zarr-sidecar.yaml` file and modifying it for your service.
   a. Change `netcdf-to-zarr` everywhere in the file to the name of your service
-  b. Set `image` under `containers` to the Docker image you want to wrap. Do the same for the
-  `HARMONY_SERVICE` environment variable entry
-  c. Set the `WORKING_DIR` environment variable to the directory where in your container that the code should
-  execute. This defaults to `/home`, so you can remove this variable if that works for your service.
+  b. Add environment variables for your service to the `env-defaults` file in the top-level directory
+  of the Harmony repository. See the NETCF_TO_ZARR entries to know what variables must be declared.
```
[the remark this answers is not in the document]
indiejames: Fixed

test/argo-callbacks.ts
```diff
@@ -36,7 +36,7 @@ describe('Argo Callbacks', function () {
   // test progress is updated correctly
   describe('when the result handler sends a batch completion notice', function () {
     hookHttpBackendEach(function () {
-      return wmsRequest(this.frontend, collection, { ...validGetMapQuery, crs: 'ASYNC', layers: collection }).ok(() => true);
+      return wmsRequest(this.frontend, collection, { ...validGetMapQuery, crs: 'ASYNC', layers: collection, turbo: false }).ok(() => true);
```
bilts, line 39: I think this is fine for this PR, but it's racking up a bit of debt, because ultimately we're going to have to fix these tests before removing Argo.  I assume what this is trying to do is force it to not use scrolling?

bilts (approved):

PR #122, opened 2021-12-14
HARMONY-1035: Remove references to shapefile subsetter and comment out ASF GDAL regression cells

vinnyinverso:
- Commented out problematic ASF GDAL Subsetter cells to match what was done in the regression repo
- Removed cells that were using PODAAC's shapefile subsetter (which subsets shapefile data via shapefile), which no longer exists, and does not have a replacement to my knowledge
- ~Wired up C1233800302-EEDTEST and C1234088182-EEDTEST to the ASF GDAL Subsetter (UAT)~

~Item 3 may still not fully work after this PR due to issues in the ASF GDAL Subsetter, but I think this is a step in the right direction. There were no umm_s records for the GDAL Subsetter so I listed the collections directly, but I can go down that route if necessary.~


chris-durbin (commented):
It looks like the changes to services.yml caused some different services to be chosen in tests breaking a few of them.

I see a lot of errors when trying to run the notebooks in calls to `show(response)`. I don't think they are related to the changes here, so I'm good with merging things in once all the tests are passing.

vinnyinverso:
> It looks like the changes to services.yml caused some different services to be chosen in tests breaking a few of them.
> 
> I see a lot of errors when trying to run the notebooks in calls to `show(response)`. I don't think they are related to the changes here, so I'm good with merging things in once all the tests are passing.

Ahhh 💡. So looking at the git history, at the time that those two collections were first used for shapefile subsetting in the notebooks, they were hooked up to _harmony_ gdal -- which is why my change broke so many tests. When Amy said they were originally hooked up to gdal I thought she meant asf gdal. I don't think harmony-gdal/harmony-service-example supports shapefiles anymore though. It must have at one point in time.

Regardless, this makes me skeptical that my services.yml change is the right thing to do, since those collections have never been linked to asf gdal.. Perhaps I'll take a pause on this until everyone is back in Jan.

vinnyinverso:
Per the conversation this morning at standup, item 3 is no longer relevant, but I think 1 and 2 still are:

> - Commented out problematic ASF GDAL Subsetter cells to match what was done in the regression repo
> - Removed cells that were using PODAAC's shapefile subsetter (which subsets shapefile data via shapefile), which no longer exists, and does not have a replacement to my knowledge
> - ~Wired up C1233800302-EEDTEST and C1234088182-EEDTEST to the ASF GDAL Subsetter (UAT)~

bilts (approved):

PR #123, opened 2021-12-22
Harmony 948, 984, 986 - Jobs dashboard enhancements

vinnyinverso:
This PR combines HARMONY-948, 984, 986. 
- display EDL username as column in jobs table
- show dates in local time, including date that job was submitted
- quickly distinguish between different execution modes

For the "mode" column that was added, you'll see 🚀  for turbo=true, 🐙  for turbo=false, 🚫  for not specified/default. I decided to go with 3 signifiers since the meaning of "default" (no turbo flag) could imply either turbo or argo execution mode depending on which jobs you're looking at. Open to suggestions if this is too confusing.

There is also a title element the user can hover over to see what this means if they're not well versed with Harmony, but given that this is mostly used internally at the moment and probably is a temporary column, I didn't want to take up too much space.

Screen Shot 2021-12-22 at 2 04 29 PM
Screen Shot 2021-12-22 at 2 05 25 PM


chris-durbin (approved):
Tested out locally and looks good.

bilts (approved):
Nice!

I'd maybe consider dropping the "mode" column, making the default something that looks less like an error, and/or hiding it on the admin screen.

I suspect few people currently care and in a couple of months, it should be all "default" (= Turbo)

vinnyinverso:
> Nice!
> 
> I'd maybe consider dropping the "mode" column, making the default something that looks less like an error, and/or hiding it on the admin screen.
> 
> I suspect few people currently care and in a couple of months, it should be all "default" (= Turbo)

Got rid of that column.

PR #124, opened 2022-01-03
Harmony 1043

asteiker:
Added notebook from PI 21.4 solution demo with Trajectory subsetter, COG, Concise, and Zarr re-chunking examples. Removed local file from zarr example to ensure it ran correctly, and added requirements and guidance on how to run notebook.


chris-durbin (approved):
I noticed the output from the run is included in the notebook. I think that's fine since this is a demo and the job output links being in there should be ok since this is all public data. I wasn't able to verify running the notebook due to dependencies - I should probably learn how to use conda forge to set up a kernel.

bilts (approved):

PR #125, opened 2022-01-07
Harmony 1045

hailiangzhang:
This PR introduces prometheus adapter deployment via helm.

It also added some service and fixed some issue in prometheus deployment.

You can run the following to deploy prometheus and the adapter:
```
./bin/deploy-prometheus
```

After a minute or so, you can query the raw metric API via:
```
kubectl get --raw /apis/custom.metrics.k8s.io/v1beta1
```
and should get something like this:
```
{"kind":"APIResourceList","apiVersion":"v1","groupVersion":"custom.metrics.k8s.io/v1beta1","resources":[]}
```

Note that a sample custom metric rule is provided in `config/prometheus-adapter-helm-values.yaml`. To test against this rule, download this attachment 
[sample-app.tar.gz](https://github.com/nasa/harmony/files/7830833/sample-app.tar.gz), unzip and run the following:
```
kubectl apply -f sample-app.deploy.yaml  -f sample-app.service.yaml -f sample-app-hpa.yaml
```
Then after another minute or so, you can query for this custom metric:
```
kubectl get --raw /apis/custom.metrics.k8s.io/v1beta1/namespaces/monitoring/pods/*/http_requests_per_second
```
and it should with some non-empty json.
Also, you should be able to see that the `sample-app` pods have been scaled up.


chris-durbin:
I deployed an earlier version of this branch and I've seen the prometheus pod get evicted twice within the past week. I'm wondering if it's logging too much or if we should do some kind of log rotation so the pod can run for longer:
> Message:        Pod ephemeral local storage usage exceeds the total limit of containers 500Mi.

hailiangzhang:
> I deployed an earlier version of this branch and I've seen the prometheus pod get evicted twice within the past week. I'm wondering if it's logging too much or if we should do some kind of log rotation so the pod can run for longer:
> 
> > Message:        Pod ephemeral local storage usage exceeds the total limit of containers 500Mi.

That's a good point ... I will check whether there is a way to rotate the logs for the pod.

vinnyinverso:
Nice! The provided instructions worked for me. I am using this version:

```
Client Version: version.Info{Major:"1", Minor:"21", GitVersion:"v1.21.3", GitCommit:"ca643a4d1f7bfe34773c74f79527be4afd95bf39", GitTreeState:"clean", BuildDate:"2021-07-15T21:04:39Z", GoVersion:"go1.16.6", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"19", GitVersion:"v1.19.7", GitCommit:"1dd5338295409edcfff11505e7bb246f0d325d15", GitTreeState:"clean", BuildDate:"2021-01-13T13:15:20Z", GoVersion:"go1.15.5", Compiler:"gc", Platform:"linux/amd64"}
```

vinnyinverso (approved):

hailiangzhang:
> Nice! The provided instructions worked for me. I am using this version:
> 
> ```
> Client Version: version.Info{Major:"1", Minor:"21", GitVersion:"v1.21.3", GitCommit:"ca643a4d1f7bfe34773c74f79527be4afd95bf39", GitTreeState:"clean", BuildDate:"2021-07-15T21:04:39Z", GoVersion:"go1.16.6", Compiler:"gc", Platform:"darwin/amd64"}
> Server Version: version.Info{Major:"1", Minor:"19", GitVersion:"v1.19.7", GitCommit:"1dd5338295409edcfff11505e7bb246f0d325d15", GitTreeState:"clean", BuildDate:"2021-01-13T13:15:20Z", GoVersion:"go1.15.5", Compiler:"gc", Platform:"linux/amd64"}
> ```

Cool!

hailiangzhang:
As chatted with @chris-durbin , I will add a deletion script for prometheus and adapter.

hailiangzhang:
> As chatted with @chris-durbin , I will add a deletion script for prometheus and adapter.

Done.

hailiangzhang:
> > I deployed an earlier version of this branch and I've seen the prometheus pod get evicted twice within the past week. I'm wondering if it's logging too much or if we should do some kind of log rotation so the pod can run for longer:
> > > Message:        Pod ephemeral local storage usage exceeds the total limit of containers 500Mi.
> 
> That's a good point ... I will check whether there is a way to rotate the logs for the pod.

@chris-durbin Looks like the pod which used more ephemeral memory is from the prometheus service instead of the adapter. Besides increasing the ephemeral memory in configuration, there is a `storage.tsdb.retention` parameter (which defines `How long to retain samples in the storage.`) from the entrypoint of this pod, which is set to be 6 hours but we could reduce it to 1 hour (maybe).

Anyway, this seems to be a separate story.

chris-durbin (approved):
I tested everything out locally successfully. The new delete script returned in under a minute and deleted everything that was installed.

chris-durbin:
> Looks like the pod which used more ephemeral memory is from the prometheus service instead of the adapter. Besides increasing the ephemeral memory in configuration, there is a storage.tsdb.retention parameter (which defines How long to retain samples in the storage.) from the entrypoint of this pod, which is set to be 6 hours but we could reduce it to 1 hour (maybe).
>
> Anyway, this seems to be a separate story.

Agreed we can make these changes as a separate ticket.

PR #126, opened 2022-01-13
Harmony 935 service yml files

indiejames:
Changes to use a single parameterized k8s template for locally deployed services.


chris-durbin (approved):
My local environment is not in a good position to test things out, but the changes look good to me.

bilts (approved):

PR #128, opened 2022-01-20
HARMONY-977: Remove argo

chris-durbin:
To test this out locally you will first need to remove your argo namespace - all resources are now created in the harmony namespace:

```
kubectl delete ns argo
```

Due to the namespace change you will also need to build the service-runner image:
```
cd tasks/service-runner
npm run build
```

Once that completes you can run:
```
bin/start-postgres-localstack
bin/deploy-services
```

Check the running pods to be sure there are no pods in the argo namespace and the expected pods are running in the harmony namespace:
```
kubectl get pods -A
```

You should now be able to use harmony the same as before.

Make sure to merge this PR in together with the harmony-ci-cd PR for HARMONY-977.


indiejames (approved):
Tested on my machine and worked on local deployment

PR #131, opened 2022-01-26
Harmony 1060

hailiangzhang:
This PR introduces variable-based services in Harmony for Giovanni implementation.
Currently the Giovanni service is not implemented in Harmony, so the added Giovanni services in `services.yaml` is still like a place holder.

### DIFF
--- a/app/models/services/index.ts
+++ b/app/models/services/index.ts
@@ -146,10 +146,8 @@ function isCollectionMatch(
     let matchedCollection;
     if ( serviceConfig.collections.includes(source.collection) ) return true;
     else if ( matchedCollection = serviceConfig.collections.find(x => x[source.collection]) ) {
-      let b = source.variables.every(x => matchedCollection[source.collection].includes(x.id));
-      return b;
-    }
-    else return false;
+      return source.variables.every(x => matchedCollection[source.collection].includes(x.id));
+    } else return false;
   });
 }
 
--- a/app/models/services/index.ts
+++ b/app/models/services/index.ts
@@ -142,7 +142,15 @@ function isCollectionMatch(
   operation: DataOperation,
   serviceConfig: ServiceConfig<unknown>,
 ): boolean {
-  return operation.sources.every((source) => serviceConfig.collections.includes(source.collection));
+  return operation.sources.every((source) => {
+    let matchedCollection;
+    if ( serviceConfig.collections.includes(source.collection) ) return true;
+    else if ( matchedCollection = serviceConfig.collections.find(x => x[source.collection]) ) {
+      let b = source.variables.every(x => matchedCollection[source.collection].includes(x.id));
+      return b;
+    }
+    else return false;
+  });
 }
 
 /**
--- a/config/services.yml
+++ b/config/services.yml
@@ -79,7 +79,7 @@ https://cmr.earthdata.nasa.gov:
           <<: *default-argo-env
           STAGING_PATH: public/gesdisc/giovanni
     collections:
-      - C1225808238-GES_DISC
+      - C1225808238-GES_DISC:
         - V2144697547-GES_DISC
     capabilities:
       subsetting:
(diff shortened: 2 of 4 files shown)


app/models/services/index.ts
```diff
@@ -142,7 +142,13 @@ function isCollectionMatch(
   operation: DataOperation,
   serviceConfig: ServiceConfig<unknown>,
 ): boolean {
-  return operation.sources.every((source) => serviceConfig.collections.includes(source.collection));
+  return operation.sources.every((source) => {
+    let matchedCollection;
+    if ( serviceConfig.collections.includes(source.collection) ) return true;
+    else if ( matchedCollection = serviceConfig.collections.find(x => x[source.collection]) ) {
```
indiejames, line 148: nit: you don't really need the `else`s here since you have returns inside your `if` blocks.

indiejames (approved):

app/models/services/index.ts
hailiangzhang: Sorry James if I missed something, but I also need to return the expected boolean in `else`s when the condition inside `if` is not true, right?

app/models/services/index.ts
```diff
@@ -142,7 +142,13 @@ function isCollectionMatch(
...
+      return source.variables.every(x => matchedCollection[source.collection].includes(x.id));
```
chris-durbin, line 149: What about requests that do not request variable subsetting (ones that specify 'all')? Can we pass all to Giovanni? Will Giovanni just return the variables it allows?

test/models/services.ts
```diff
@@ -425,6 +425,45 @@ describe('services.chooseServiceConfig and services.buildService', function () {
       });
     });
   });
+
+  describe('when the collection match includes variables for variable-based service', function () {
```
chris-durbin, line 429: Add a test for requesting 'all' variables to document whether we return an error or not.

Add a test for specifying a variable that exists, but is not configured in the variable list within services.yml.

chris-durbin (changes requested):
Had a couple comments, also was thinking we should update the adapting-new-services.md documentation to explain how collection and variables can be configured in services.yml.

app/models/services/index.ts
hailiangzhang: Yes, Giovanni is variable-based service, and we have to specify the variables for the request.

test/models/services.ts
hailiangzhang: So, if my [above reply](https://github.com/nasa/harmony/pull/131#discussion_r793050209) makes sense, do we still need a test for `all`?

app/models/services/index.ts
hailiangzhang: After a walkthrough, we found that when `all` is specified from the request, `source.variables` will be `undef` and an exception will be raised.
This is fixed from more recent commits (along with test added).

config/services.yml
```diff
@@ -413,6 +446,39 @@ https://cmr.uat.earthdata.nasa.gov:
...
+    capabilities:
+      subsetting:
+        bbox: true
+        variable: true
+        multiple_variable: false
+      output_formats:
+        - text/csv
+    steps:
+      - image: !Env ${QUERY_CMR_IMAGE}
+      - image: !Env ${GESDISC_GIOVANNI_IMAGE}
```
chris-durbin, line 480: You should add the new variable to `env-defaults`.

test/models/services.ts
```diff
@@ -425,6 +425,62 @@ describe('services.chooseServiceConfig and services.buildService', function () {
...
+          collections: [
+            {
+              [collectionId]: [ variableId ],
+            },
+          ],
+        },
+      ];
+    });
+
+    describe('requesting service with variable subsetting', function () {
```
chris-durbin, line 450: You don't need to do this in the current PR, but I have some notes for all the tests I'd like to see when we fully integrate the giovanni service.

1. One variable requested and matches (done)
2. Request for 'all' - aka no variable subsetting requested (done)
3. Two variables provided and both match
4. Two variables provided and one of the two match
5. One variable provided and doesn't match

chris-durbin (approved):

test/models/services.ts
hailiangzhang: Sure, I will add them in some following giovanni stories ... thanks!

hailiangzhang: Done in story 1061.

PR #132, opened 2022-01-26
Harmony 1054 pod counts

indiejames:
This PR adds the deployment of the `kube-state-metrics` service to the `deploy-prometheus` script to provide a lot of metrics convenient to monitoring the cluster/pods. An OPERATIONS.md file provides a starting point for documenting
general operating procedures not specific to our Harmony deployments. 

Specifically, this PR makes it possible to determine the number of running pods for each service (AC for HARMONY-1054).
This can be verified as follows:

1. deploy this branch to the sandbox
2. get the pod identifier for Prometheus
```
kubectl get pods -n monitoring [--kubeconfig <config file>]
```
3. forward port `9090` from the pod to your local machine
```
kubectl port-forward <Prometheus Pod ID> 9090:9090 -n monitoring [--kubeconfig <config file>]
```
4. Execute the following promQL in the UI
```
sum(
count(kube_pod_status_phase{phase="Running", namespace="harmony"}) by (pod)
*
on (pod) group_left(label_name) sum (kube_pod_labels{namespace="harmony",label_name=~".*"}) without (namespace)
) by (label_name)
```

You can replace ".*" in the query with a service name to get the metrics for a single service.

### DIFF
--- a/bin/delete-prometheus
+++ b/bin/delete-prometheus
@@ -8,10 +8,7 @@ set +a
 eval "$env_save"
 
 # Delete kube-state-metrics service
-kubectl delete deployment kube-state-metrics -n monitoring
-kubectl delete service kube-state-metrics -n monitoring
-kubectl delete clusterrolebinding kube-state-metrics
-kubectl delete clusterrole kube-state-metrics
+kubectl delete -f ./config/kube-state-metrics
 
 # Delete prometheus adapter
 PROMETHEUS_NAMESPACE=monitoring
--- a/bin/delete-prometheus
+++ b/bin/delete-prometheus
@@ -7,6 +7,12 @@ source ".env"
 set +a
 eval "$env_save"
 
+# Delete kube-state-metrics service
+kubectl delete deployment kube-state-metrics -n monitoring
+kubectl delete service kube-state-metrics -n monitoring
+kubectl delete clusterrolebinding kube-state-metrics
+kubectl delete clusterrole kube-state-metrics
+
 # Delete prometheus adapter
 PROMETHEUS_NAMESPACE=monitoring
 PROMETHEUS_ADAPTER_HELM_NAME=harmony-prometheus-adapter
@@ -14,3 +20,4 @@ helm delete ${PROMETHEUS_ADAPTER_HELM_NAME} -n ${PROMETHEUS_NAMESPACE}
 
 # Delete prometheus
 envsubst < "config/prometheus.yaml" | kubectl delete -f -
+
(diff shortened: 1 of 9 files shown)


bin/delete-prometheus
```diff
@@ -7,10 +7,17 @@ source ".env"
 set +a
 eval "$env_save"
 
+# Delete kube-state-metrics service
+kubectl delete deployment kube-state-metrics -n monitoring
```
chris-durbin, line 11: Can we instead run:

```
kubectl delete -f ./config/kube-state-metrics
```

bin/deploy-prometheus
```diff
@@ -38,6 +38,9 @@ else
                     -n ${PROMETHEUS_NAMESPACE} \
                     ${PROMETHEUS_ADAPTER_HELM_NAME} \
                     prometheus-community/prometheus-adapter
-
 fi
 
+# Deploy kube-state-metrics service to support monitoring
+echo "deploying kube-state-metrics service"
+kubectl apply -f ./config/kube-state-metrics
```
chris-durbin, line 45: Oh I didn't realize you could deploy a directory of YAML files with a single statement. That's useful.

chris-durbin (approved):
Tested out locally successfully. Will kube-state-metrics interfere with the metrics-server that is deployed to the kube-system namespace in AWS environments? You'll want to sanity check with a sandbox deployment since the deploy-prometheus script is used there as well.

vinnyinverso (approved):

bin/delete-prometheus
indiejames: done

bin/deploy-prometheus
indiejames: yeah - handy

indiejames:
I removed references to `num_ready_work_items_avg`.

> May want to change this:
> 
> Additionally, Harmony deploys the [Prometheus Adapter](https://github.com/kubernetes-sigs/prometheus-adapter) to generate a custom metric based on the average value of `num_ready_work_items` over a two minute window. This metric, `num_ready_work_items_avg`, is used by the [Horizontal Pod Autoscaler (HPA)](https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale/) to scale the number of pods for a given service up or down in response to load.
> 
> to reflect that we're not using the _avg at this time
> 
> Additionally, Harmony deploys the [Prometheus Adapter](https://github.com/kubernetes-sigs/prometheus-adapter) to generate a custom metric `num_ready_work_items`. This metric, `num_ready_work_items`, is used by the [Horizontal Pod Autoscaler (HPA)](https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale/) to scale the number of pods for a given service up or down in response to load.

PR #135, opened 2022-02-09
DAS-1371 - Use error message from error.json in service-runner, if present.

owenlittlejohns:
This PR follows on from work being done within HOSS to ensure error messages are well handled. In that work, we found that Harmony has a bit of an issue with logging messages that include curly braces. This is a particular issue for OPeNDAP error messages:

* OPeNDAP returns error messages that include curly braces, (e.g.: `{\n            font-size: 24px;\n            font-weight: bold;\n        }`)
* `harmony-service-lib-py` [logs all error messages to STDOUT](https://github.com/nasa/harmony-service-lib-py/blob/main/harmony/http.py#L361).
* `harmony-service-lib-py` then swallows the OPeNDAP error message, so the backend service can't access it.
* When the backend service fails, `harmony-service-lib-py` [creates `error.json` in the `WorkItem` directory](https://github.com/nasa/harmony-service-lib-py/blob/main/harmony/cli.py#L132).
* The Harmony service-runner currently [searches through STDOUT for matches to the regular expression `\{.*\}`](https://github.com/nasa/harmony/blob/main/tasks/service-runner/app/service/service-runner.ts#L57). This will include things like the part of the OPeNDAP message quoted above.
* The service-runner tries to use `JSON.parse()` on the match to the regular expression that is within the OPeNDAP error message.
* This causes a JSON parsing error. In this case it would be `Unexpected token \ in JSON at position 1`.

In this PR, the `_getErrorMessage` function is updated to first check for the `error.json` output from `harmony-service-lib-py`. This should contain a single JSON log entry. If this file is present, the contents will be used to generate the final output error message. If the file is absent, the previous logic of scanning the STDOUT log messages is used as before.

### DIFF
--- a/tasks/service-runner/app/service/service-runner.ts
+++ b/tasks/service-runner/app/service/service-runner.ts
@@ -1,5 +1,5 @@
 import * as k8s from '@kubernetes/client-node';
-import { readdirSync } from 'fs';
+import { existsSync, readdirSync, readFileSync } from 'fs';
 import stream from 'stream';
 import { sanitizeImage } from '../../../../app/util/string';
 import env from '../util/env';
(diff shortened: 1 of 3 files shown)


owenlittlejohns:
I've tested this change locally, and an OPeNDAP failure is successfully handled by the new code.

chris-durbin (approved):
Fantastic. I didn't notice a harmony-service-lib-py PR - are you also planning to add code to the service library to generate error.json?

owenlittlejohns:
Hi, Chris. I think `harmony-service-lib-py` already generates `error.json` using the [`_write_error` function](https://github.com/nasa/harmony-service-lib-py/blob/main/harmony/cli.py#L132). It looks like that should be called by `_invoke` in the case of a `HarmonyException` or `BaseException` being raised. (To be honest, it's where I got the idea for this change - so thanks to Patrick and HARONY-556 for that!)

bilts (approved):

PR #137, opened 2022-02-11
HARMONY-1068: netcdf-to-zarr now concatenates by default

chris-durbin:
Added a new data operation schema with a new concatenate parameter to send to backend services. For netcdf-to-zarr requests we will send concatenate: true by default. Note that the user can override this behavior by specifying concatenate=false in their request.

There's also a change to the harmony-service-library to parse the concatenate parameter.

The netcdf-to-zarr service doesn't make use of the parameter yet, so the only thing we can verify is that harmony is sending "concatenate=true" by default for netcdf-to-zarr and "concatenate=false" by default for other services.

I regression tested that PODAAC concise concatenation continues to work the same as before.


app/middleware/concatenation.ts
```diff
@@ -0,0 +1,66 @@
...
+ * to choosing the service that will be used.
+ *
+ * @param req - The client request, containing an operation
+ * @param res - The client response
+ * @param next - The next function in the middleware chain
+ */
+export function preServiceConcatenationHandler(
+  req: HarmonyRequest, _res: Response, next: NextFunction,
+): void {
+  const query = keysToLowerCase(req.query);
```
bilts, line 18: Realizing we do this lowercasing everywhere. It's probably begging for a move to a middleware or something unless we ever care about keys in a case-sensitive way.

bilts: (Not necessary for this PR)

bilts (approved):

vinnyinverso (approved):

PR #139, opened 2022-02-14
Harmony 670 - EKS monitoring and alerting

vinnyinverso:
A PR for CICD code is forthcoming. You may want to test this out there instead if you'd like to see the alert emails get sent to your inbox. (AlertManager will only get deployed in AWS.)

The overall architecture of these changes is:
Prometheus uses its "rules" (PromQL queries) to determine which alerts are "firing" ==> firing alerts are sent via POST request to the Alert Manager ==> Alert Manager routes the alerts to a notification service like AWS SNS

To test this out locally:
- .env changes
  - Set `PROMETHEUS_POD_NOT_HEALTHY_DURATION` and `PROMETHEUS_NODE_NOT_READY_DURATION` to **1m** in your .env file if you want to see the alert sooner than what is set in env-default.
- Alert triggering
  - To test the KubernetesPodNotHealthy alert, before deploying harmony services, delete a service image or two and set the image location(s) in your .env to a nonexistent location (so that when the pod is spun up it can't get to a ready state).
  - To test the KubernetesNodeNotReady alert, tweak the prometheus.yaml where it says `== 0` to say `==1`. I couldn't find a good way of making a node unhealthy, but I think testing the opposite condition should be sufficient to give an idea of what would happen if the node actually went down. (Try pasting variations of `kube_node_status_condition{condition="Ready",status="true"} == $zero_or_one` in the prometheus charting UI if you're not convinced.)

`bin/deploy-prometheus`

`kubectl port-forward {prometheus-pod} 9090:9090 -n monitoring`

Navigate to the prometheus UI 'alerts' tab to see the alerts firing. Note that this is reflecting the changes in prometheus.yaml which is where you configure the alerting rules (not the changes in alert-manager.yaml where the alerts are routed to SNS in deployed environments).

Example email:

Screen Shot 2022-02-14 at 9 38 46 AM


config/prometheus.yaml
```diff
@@ -160,3 +187,102 @@ spec:
...
+      repeat_interval: 2h
+      routes:
+    receivers:
+      - name: 'default'
+        sns_configs:
+        - topic_arn: $PROMETHEUS_ALERTS_ARN
+          subject: '[Alert Manager $CLIENT_ID] {{ .GroupLabels.alertname }}'
+          sigv4:
+            region: $AWS_DEFAULT_REGION
+            access_key: $AWS_ACCESS_KEY_ID
```
chris-durbin, line 215: https://prometheus.io/docs/alerting/latest/configuration/#sigv4_config

We should set up an IAM role and use that role instead of access keys.

env-defaults
```diff
@@ -338,9 +341,24 @@ PROMETHEUS_REQUESTS_EPHEMERAL_STORAGE=300Mi
...
+PROMETHEUS_ALERT_MANAGER_LIMITS_MEMORY=1Gi
+
+
 PROMETHEUS_PROMETHEUS_SCRAPE_INTERVAL=15s
 PROMETHEUS_POD_MANAGER_SCRAPE_INTERVAL=15s
 
+# Set this to the ARN of the prometheus-alerts-local AWS SNS topic from harmony sandbox
+# (in your .env file) if you plan to deploy Prometheus for local development.
+# Your email address must also be subscribed to the topic in order to receive alerts.
+PROMETHEUS_ALERTS_ARN=
```
chris-durbin, line 356: This one is tricky - I guess we want to allow people to deploy Prometheus to their docker-desktop cluster, but then use sandbox resources for alerting?

If the goal is to make that possible I guess we should also add a PROMETHEUS_ALERTS_ROLE_ARN and use that role ARN when deploying Prometheus.

vinnyinverso: My hope was that people could see actual emails when developing locally.

env-defaults
```diff
@@ -289,8 +289,11 @@ BUILT_IN_TASK_PREFIX=
 # Version to use for all built-in tasks.  "latest" (local / sandbox), "sit", "uat", or "prod"
 BUILT_IN_TASK_VERSION=latest
 
-# AWS region where harmony is deployed, irrelevant for local development
+# Set these to your harmony sandbox credentials (in your .env file) if you plan to deploy Prometheus for local development.
+# AWS region where harmony is deployed.
 AWS_DEFAULT_REGION=us-west-2
+AWS_ACCESS_KEY_ID=
```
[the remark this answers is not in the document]
vinnyinverso: I'm going to set up a role instead like Chris suggested.

vinnyinverso:
I updated the testing instructions in the description of the PR to reflect the recent changes. I also should probably change the alertmanager image to something other than latest.

bilts (approved):

chris-durbin (approved):
Verified along with the harmony-ci-cd change by deploying to sandbox and received an alert to my email.

PR #141, opened 2022-02-16
HARMONY-451: Fix fair queueing queries for postgres

indiejames:
Fixes an issue with the queries not working on postgres

### DIFF
--- a/tasks/service-runner/config/service-template.yaml
+++ b/tasks/service-runner/config/service-template.yaml
@@ -5,7 +5,7 @@ metadata:
   labels:
     name: $SERVICE_NAME
 spec:
-  replicas: 5
+  replicas: 1
   selector:
     matchLabels:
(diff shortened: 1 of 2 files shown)


tasks/service-runner/config/service-template.yaml
```diff
@@ -5,7 +5,7 @@ metadata:
   labels:
     name: $SERVICE_NAME
 spec:
-  replicas: 1
+  replicas: 5
```
chris-durbin, line 8: revert this change

chris-durbin (approved):
Paired with James while testing this against postgres - looks good.

tasks/service-runner/config/service-template.yaml
indiejames: done

bilts (approved):

PR #143, opened 2022-02-22
Harmony 795 granule limits

indiejames:
This PR adds configurable granule limits for collections associated with a service.


chris-durbin (approved):
Just have the one question about whether we need to consider how to handle different limits for the same collection for different services. I think at a minimum the warning message should include some kind of service identifier (name from services.yml?).

app/middleware/cmr-granule-locator.ts
```diff
@@ -37,48 +47,73 @@ function getBbox(collection: cmr.CmrCollection, granule: cmr.CmrGranule): Boundi
...
+ * Get the maximum number of granules that should be used from the CMR results
+ * 
+ * @param req - The client request, containing an operation
+ * @param collection - The id of the collection to which the granules belong
+ * @returns a tuple containing the maximum number of granules to return from the CMR and the
+ * reason why it is being limited
  */
-function getMaxGranules(req: HarmonyRequest): number {
-  const query = keysToLowerCase(req.query);
+export function getMaxGranules(req: HarmonyRequest, collection: string): [number, GranuleLimitReason] {
```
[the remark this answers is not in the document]
indiejames: No - fixed

app/middleware/cmr-granule-locator.ts
```diff
@@ -37,48 +47,73 @@ function getBbox(collection: cmr.CmrCollection, granule: cmr.CmrGranule): Boundi
...
-    } else {
-      message += ' because of system constraints.';
+
+    switch (reason) {
+      case GranuleLimitReason.MaxResults:
+        message += ` because you requested ${operation.maxResults} maxResults.`;
+        break;
+
+      case GranuleLimitReason.Collection:
+        message += ` because collection ${collection} is limited to ${numGranules} for the ${req.context.serviceConfig.name} service.`;
```
bilts, line 111: I'm not sure I'd put the service name in there, as it feels like sausage making. Maybe just "for your requested operations" or something like that.

app/middleware/cmr-granule-locator.ts
```diff
@@ -124,15 +159,15 @@ async function cmrGranuleLocatorTurbo(
     const queries = sources.map(async (source) => {
       logger.info(`Querying granules for ${source.collection}`, { cmrQuery, collection: source.collection });
       const startTime = new Date().getTime();
-      const maxResults = getMaxGranules(req);
+      const [maxResults, _reason] = getMaxGranules(req, source.collection);
```
bilts, line 162: Unpacking a tuple only to discard it twice in this method and probably elsewhere maybe suggests using an object instead of an array.

bilts (approved):

app/middleware/cmr-granule-locator.ts
indiejames: Yeah, I'm on the fence about that.

app/middleware/cmr-granule-locator.ts
indiejames: I think based on the reasons Chris mentioned above we need to add the service name - the same collection might have different limits defined for different services so we can just say "limited for the collection" as it would be confusing if two different limits are mentioned for the same collection without distinguishing by service as well.

app/middleware/cmr-granule-locator.ts
indiejames: Done.

PR #144, opened 2022-02-23
HARMONY-670: Wait before firing alert so that pod restarts don't trigger POD_NOT_HEALTHY alert

vinnyinverso:
I added a configurable env variable so that we can tweak how long the 'pod not healthy' alert must be firing before the alert is sent to the alert manager. 

During deployments and when HPAs scale up/down, depending on the timing of the pod startup, a pod not healthy alert is incorrectly fired. This is due to metrics that are missing from the time before the pod was created. For a very brief period of time it appears that the pod was in a pending state for the past +∞ minutes due to the missing metrics (the brief "pending" state is extrapolated into the past).

By waiting PROMETHEUS_POD_NOT_HEALTHY_WAIT_FOR (defaulted to 5 minutes), Prometheus will have more than enough time to collect a subsequent sample for the kube_pod_status_phase metric which should indicate that the pod is healthy, assuming it was restarted. This subsequent sample should prevent the undesired alert from firing.


env-defaults
```diff
@@ -351,9 +351,12 @@ PROMETHEUS_RETENTION_TIME=180d
 # pod may be evicted due to running out of ephemeral space.
 PROMETHEUS_RETENTION_SIZE=1750MB
 
-# how long before the specified alert is triggered/firing
+# how many consecutive minutes to sample in the PromQL query for KubernetesPodNotHealthy
```
indiejames, line 354: can you add something in the comments here about what units these are. I still get confused by wheat 10m means.

vinnyinverso:
I'm going to test this out a little bit more locally.

chris-durbin (approved):

PR #145, opened 2022-02-23
Harmony 1063

hailiangzhang:
This PR introduces spatial point to the query parameter.


app/frontends/ogc-coverages/get-coverage-rangeset.ts
```diff
@@ -70,6 +70,11 @@ export default function getCoverageRangeset(
     if (bbox) {
       operation.boundingRectangle = bbox;
     }
+    const point = parsePointParam(query.point);
+    if (point) {
+      if (bbox) throw new RequestValidationError('bounding_box and point query parameters should not co-exist');
```
chris-durbin, line 75: What about point and shapefile? Should we throw a validation error for that as well?

app/models/data-operation.ts
```diff
@@ -520,6 +532,26 @@ export default class DataOperation {
...
   }
 
+  /**
+   * Sets the spatial point to be used for spatial subsetting, an array of 2 coordinates:
+   *   [ Longitude, Latitude ]
+   *
+   * @param point - The spatial point in form of [ Longitude, Latitude ]
+   */
+  set spatialPoint(point: Array<number>) {
+    this.model.point = point;
```
chris-durbin, line 542: I'm thinking that point should be under subset (`this.model.subset.point`). That will affect the schema in addition to the model changes in this file.

app/schemas/ogc-api-coverages/1.0.0/ogc-api-coverages-v1.0.0.yml
```diff
@@ -666,6 +668,22 @@ components:
         type: array
         items:
           type: number
+    point:
+      name: point
+      in: query
+      description: |
+        Only collections that have a geometry that contains a spatial point are selected. The spatial point is provided as two numbers:
+        * Longitude, coordinate axis 1 * Latitude, coordinate axis 2
+        The coordinate reference system of the values is WGS84 longitude/latitude (http://www.opengis.net/def/crs/OGC/1.3/CRS84).
```
chris-durbin, line 677: I sort of remember while we were pairing you were reversing the points returned by CMR metadata. I could see that being confusing for end users - should we use latitude/longitude like the CMR?

chris-durbin: I forgot OGC might have a way for specifying the lat/lon order of points. You can disregard my comment about CMR and make sure we are matching the OGC specification.

hailiangzhang (commented):
Have moved `point` to `subset` in data operation schema; also make sure `point` and `shapefile` doesn't co-exist.

app/frontends/ogc-coverages/get-coverage-rangeset.ts
```diff
@@ -70,6 +70,12 @@ export default function getCoverageRangeset(
     if (bbox) {
       operation.boundingRectangle = bbox;
     }
+    const point = parsePointParam(query.point);
+    if (point) {
+      if (bbox || req.rawHeaders.filter((x) => new RegExp('shapefile=*').test(x)).length)
```
chris-durbin, line 75: Ideally we should be able to check for `req.context.shapefile`, and if not that `req.operation.geojson` instead of a regex on the headers. If neither of those work I'm ok with merging this in.

app/schemas/data-operation/CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+## [0.14.0] - 2022-02-19
+### Added
+- DataOperation.point - Spatial point provided by the CMR
```
chris-durbin, line 3: I believe this is the spatial point specified by the user. Also it is now under `subset.point`.

chris-durbin (approved):

app/frontends/ogc-coverages/util/parameter-parsing.ts
```diff
@@ -160,6 +160,44 @@ function _getDimensionName(value: string): string {
...
+ */
+export function parsePointParam(
+  values: string,
+  dimConfig: DimensionConfig = dimensionConfig,
+): number[] {
+  let results;
+  let coordinate, coordinates;
+  if (values !== undefined){
+    coordinates = values;
+    if ( coordinates.length !== 2 )
```
indiejames, line 183: is this correct? `values` is a string - should it be length === 2?

app/frontends/ogc-coverages/util/parameter-parsing.ts
```diff
@@ -160,6 +160,44 @@ function _getDimensionName(value: string): string {
...
+      throw new ParameterParseError(`wrong number of spatial coordinates provided in "${values}"`);
```
indiejames, line 184: You should point out both the expected number of coordinates and the received number in the error message.

app/frontends/ogc-coverages/util/parameter-parsing.ts
```diff
@@ -160,6 +160,44 @@ function _getDimensionName(value: string): string {
   }
 }
 
+/**
+ * Parses the provided point parameters and ensures they are valid, throwing an error message
+ * if not
+ *
+ * @param values - An array of all the specified point parameters from the request
+ * @param dimConfig - A mapping of dimension names to min, max, and data type values,
+ *   see `dimensionInfo` (the default value) in this file.  Usually should not be specified,
```
indiejames, line 169: should that be `dimensionConfig` instead of `dimensionInfo`?

app/frontends/ogc-coverages/util/parameter-parsing.ts
```diff
@@ -160,6 +160,44 @@ function _getDimensionName(value: string): string {
...
+  let results;
+  let coordinate, coordinates;
+  if (values !== undefined){
+    coordinates = values;
+    if ( coordinates.length !== 2 )
+      throw new ParameterParseError(`wrong number of spatial coordinates provided in "${values}"`);
+
+    results = ['lon', 'lat'].map( (dimName, idx): number => {
+      const dim = dimConfig[dimName];
+      coordinate = coordinates[idx];
```
indiejames, line 188: How does this work? It seems like you took a string parameter, `values`, and tried to coerce it to an array somehow?

chris-durbin (changes requested):
I just realized the validation tests are missing. Can you add error tests for parsing points to this file: `test/ogc-api-coverages/util-parameter-parsing.ts`

app/frontends/ogc-coverages/util/parameter-parsing.ts
```diff
@@ -160,6 +160,44 @@ function _getDimensionName(value: string): string {
...
+ * @param values - An array of all the specified point parameters from the request
+ * @param dimConfig - A mapping of dimension names to min, max, and data type values,
+ *   see `dimensionInfo` (the default value) in this file.  Usually should not be specified,
+ *   except for testing.
+ * @returns An array with two elements corresponding to [longitude, latitude]
+ * @throws ParameterParseError - if a subset parameter cannot be parsed, has unrecognized
+ *   axis names, or is otherwise invalid
+ */
+export function parsePointParam(
+  values: string,
```
chris-durbin, line 176: This type is incorrect - values is an Array of numbers as defined in the OGC schema for point.

indiejames, line 176: I see the problem now - `values` should be type `string[]`

app/frontends/ogc-coverages/get-coverage-rangeset.ts
hailiangzhang: Done.

app/frontends/ogc-coverages/get-coverage-rangeset.ts
hailiangzhang: Both `req.context.shapefile` and `req.operation.geojson` is not generated at this point.

app/schemas/data-operation/CHANGELOG.md
hailiangzhang: Ah yes ... good catch!

hailiangzhang: Done.

app/frontends/ogc-coverages/util/parameter-parsing.ts
hailiangzhang: Good catch ... actually `values` is supported to be an array.
Fixed.

app/frontends/ogc-coverages/util/parameter-parsing.ts
hailiangzhang: Good point ... Done.

app/frontends/ogc-coverages/util/parameter-parsing.ts
hailiangzhang: Yes (and fixed).

app/frontends/ogc-coverages/util/parameter-parsing.ts
hailiangzhang: Fixed by correcting the parameter type.

hailiangzhang:
> I just realized the validation tests are missing. Can you add error tests for parsing points to this file: `test/ogc-api-coverages/util-parameter-parsing.ts`

Done.

chris-durbin (approved):

indiejames (approved):

PR #148, opened 2022-03-02
HARMONY-1076: Add migration to add foreign key constraint between work_items and workflow_steps

indiejames:
Tested locally with postgres and in a sandbox deployment. I also ran the following query to verify that we don't already have any work_items rows violating the new foreign key constraint in SIT/UAT/Prod.

```sql
select * from work_items w left join workflow_steps ws on "w"."jobID" = "ws"."jobID" and "w"."workflowStepIndex" = "ws"."stepIndex" where ws.id is null;
```

If you want to test this locally, check out the branch and the migration using `knex`. Then for an existing job ID try to insert a row into work_items that does not have a matching workflow_step row:

```sql
insert into work_items values(<some unused work_item_id>,<some real job id>,1000,'','foo','ready','',now(),now());
```

1000 above is just a really high workflowStepIndex that should not exist.

### DIFF
--- a/db/migrations/20220301141810_add_work_item_foreign_keys.js
+++ b/db/migrations/20220301141810_add_work_item_foreign_keys.js
@@ -1,0 +1,14 @@
+
+exports.up = function(knex) {
+  return knex.schema.table('work_items', (t) => {
+    t.foreign(['jobID','workflowStepIndex'])
+      .references(['jobID','stepIndex'])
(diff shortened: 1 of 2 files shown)


chris-durbin (approved):
Tested both the up and down migrations and verified the foreign key constraints were added and dropped correctly. Sanity checked requests still worked fine.

vinnyinverso (approved):

PR #149, opened 2022-03-03
HARMONY-1086: Fix URL formation error handling

vinnyinverso:
I verified manually that this doesn't break anything, but wasn't able to add a test for this because there is validation logic that prevents me from adding a job with a malformed request url.

### DIFF
--- a/app/frontends/workflow-ui.ts
+++ b/app/frontends/workflow-ui.ts
@@ -49,6 +49,8 @@ export async function getJobs(
           const path = url.pathname + url.search;
           return path;
         } catch (e) {
+          req.context.logger.error(`Could not form a valid URL from job.request: ${this.request}`);
+          req.context.logger.error(e);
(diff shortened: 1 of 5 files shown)


app/frontends/workflow-ui.ts
```diff
@@ -44,9 +44,13 @@ export async function getJobs(
...
       jobBadge() { return badgeClasses[this.status]; },
       jobUrl() {
-        const url = new URL(this.request);
-        const path = url.pathname + url.search;
-        return path;
+        try {
+          const url = new URL(this.request);
+          const path = url.pathname + url.search;
+          return path;
+        } catch (e) {
```
chris-durbin, line 51: We should also log the exception along with the request in this case.

chris-durbin (approved):

indiejames (approved):

PR #152, opened 2022-03-07
HARMONY-1082: Add more specific routing to return 404s for bad URLs

indiejames:
I made the acceptable routes a bit more specific to prevent bad URLs from being processed as service requests. They now
return a 404. This can be tested with the following (missing variable):

```
curl -Ln -bj "http://localhost:3000/C1234208436-POCLOUD/ogc-api-coverages/1.0.0/collections/coverage/rangeset?forceAsync=true&subset=lat(-20:0)&subset=lon(-80:0)&maxResults=1&format=application/x-zarr"
```


chris-durbin (approved):
Tested with the same issue that I wrote the ticket for (not including any 'collections' aka variables in the URL). Response returned was a 404 with a JSON body:
```
{
    code: "harmony.NotFoundError",
    description: "Error: The requested page was not found."
}
```

PR #155, opened 2022-03-09
Harmony 1086 - Workflow UI jobs page uses env max page size when page limit exceeded

vinnyinverso:
Updated this according to Patrick's comment https://bugs.earthdata.nasa.gov/browse/HARMONY-1086


test/workflow-ui/jobs.ts
```diff
@@ -131,6 +132,14 @@ describe('Workflow UI jobs route', function () {
...
     });
 
+    describe('who asks for more than env.maxPageSize jobs', function () {
+      hookWorkflowUIJobs({ username: 'woody', limit: env.maxPageSize + 999 });
+      it('returns all of the users jobs without error', function () {
+        const listing = this.res.text;
+        expect((listing.match(/job-table-row/g) || []).length).to.equal(3);
+      });
+    });
+
```
chris-durbin, line 142: Is there a scenario we can test where useMaxWhenExceeded is false? Maybe there's already a test case that validates that?

chris-durbin (approved):

indiejames (approved):

test/workflow-ui/jobs.ts
vinnyinverso: I believe `Jobs listing route => pagination` exercises those cases.

PR #159, opened 2022-03-11
HARMONY-1091: Fixes cases where fair queueing is choosing a user with a canceled job as waiting for work for a service

chris-durbin:
Do not consider a user as waiting for work on a service if the job associated with any 'ready' work items is in a terminal state. Fixes an issue when jobs are canceled, but not all work items were marked as canceled.

During our regression tests we perform a cancel, and for some reason when we cancel a request it does not always mark all of the work items for that job as canceled (separate bug). This makes sure that if a job is canceled we ignore any work items for that job regardless of their status.


indiejames (approved):

indiejames:
Tested locally

vinnyinverso (approved):

PR #160, opened 2022-03-14
Harmony 1062

hailiangzhang:
This PR adds giovanni-adapter service (which replaces query-cmr).


tasks/giovanni-adapter/app/stac/catalog.ts
```diff
@@ -0,0 +1,69 @@
...
+import { v4 as uuid } from 'uuid';
+import _ from 'lodash';
+import StacItem from './item';
+import { StacCatalog, StacLink } from './types';
+
+/**
+ * Implementation of the StacCatalog type with constructor, write function, and ability
+ * to add children
+ */
+export default class Catalog implements StacCatalog {
```
chris-durbin, line 13: It feels like we should just define the Stac data structures and functions in a single library and use them throughout. You don't need to do that for this PR, but we should have a ticket for it.

test/ogc-api-coverages/get-coverage-rangeset.ts
```diff
@@ -390,7 +390,7 @@ describe('OGC API Coverages - getCoverageRangeset', function () {
           name: 'turbo',
         },
         steps: [{
-          image: 'fake-internal.earthdata.nasa.gov/nexus-service/foo:uat',
```
chris-durbin, line 393: I believe this was testing something with an internal earthdata.nasa.gov URL. Can we put this back?

chris-durbin (changes requested):
I only had a few comments. The big one that needs to be fixed prior to merge is the matching to see if the step is for the query-cmr image.

You'll also need to merge main into your branch and then fix the failing build.

tasks/giovanni-adapter/app/stac/types.ts
```diff
@@ -0,0 +1,57 @@
+/**
+ * A link within a STAC item or catalog
+ * https://github.com/radiantearth/stac-spec/blob/master/catalog-spec/catalog-spec.md#link-object
+ * https://github.com/radiantearth/stac-spec/blob/master/item-spec/item-spec.md#link-object
+ */
+export interface StacLink {
+  rel: 'self' | 'root' | 'parent' | 'child' | 'item' | 'collection' | 'derived_from' | 'alternate' | string;
```
indiejames, line 7: I think this can just be `rel: string` since it won't enforce the other entries since they are a subset of `string`

tasks/giovanni-adapter/app/stac/types.ts
```diff
@@ -0,0 +1,57 @@
...
+/**
+ * An asset within a STAC item
+ * https://github.com/radiantearth/stac-spec/blob/master/item-spec/item-spec.md#asset-object
+ */
+export interface StacAsset {
+  href: string;
+  title?: string;
+  description?: string;
+  type?: string;
+  roles?: ('thumbnail' | 'overview' | 'data' | 'metadata' | string)[];
```
indiejames, line 22: same comment as line 7 above

indiejames (approved):
Approved with minor comments once you address Chris's comments

env-defaults
```diff
@@ -269,6 +269,13 @@ QUERY_CMR_REQUESTS_MEMORY=128Mi
 QUERY_CMR_LIMITS_CPU=128m
 QUERY_CMR_LIMITS_MEMORY=512Mi
 
+GIOVANNI_ADAPTER_IMAGE=harmonyservices/giovanni-adapter:latest
```
chris-durbin, line 272: Note that we're going to need to add the Giovanni service setup in harmony-ci-cd in order to start using it in AWS deployed environments. Are you doing this with a separate ticket?

app/middleware/cmr-granule-locator.ts
```diff
@@ -169,20 +169,23 @@ async function cmrGranuleLocatorTurbo(
...
-        req.accessToken,
-        maxGranules,
-      );
-      if (hits === 0) {
-        throw new RequestValidationError('No matching granules found.');
-      }
-      const msTaken = new Date().getTime() - startTime;
-      logger.info('timing.cmr-initiate-granule-scroll.end', { durationMs: msTaken, hits });
+      // Only perform CMR granule query when needed by the first step
+      if ( req.context.serviceConfig.steps[0].image.match('^harmonyservices/query-cmr:.*') ) {
```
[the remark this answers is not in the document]
hailiangzhang: Done (removed `^`)

tasks/giovanni-adapter/app/cli.ts
```diff
@@ -0,0 +1,129 @@
...
+import DataOperation from '../../../app/models/data-operation';
+import { createEncrypter, createDecrypter } from '../../../app/util/crypto';
+import logger from '../../../app/util/log';
+import Catalog from './stac/catalog';
+import StacItem from './stac/item';
+import { BoundingBox } from '../../../app/util/bounding-box';
+
+// giovanni globals
+import giovanni_datafield_config from '../config/giovanni-datafield.json';
+const giovanni_base_url = 'https://api.giovanni.earthdata.nasa.gov/';
```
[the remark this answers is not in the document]
hailiangzhang: Yes, I thought about it at some point (different endpoint for different environment), but forgot to confirm with Christine. Let me do that.
BTW, just make sure, we basically needs 2 endpoint: UAT and production, right?

tasks/giovanni-adapter/app/stac/catalog.ts
hailiangzhang: Good point ... story created:
https://bugs.earthdata.nasa.gov/browse/HARMONY-1101

app/middleware/cmr-granule-locator.ts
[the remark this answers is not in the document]
hailiangzhang: Done.

env-defaults
```diff
@@ -278,7 +285,7 @@ QUERY_CMR_LIMITS_MEMORY=512Mi
 # so it will always be deployed, and does not need to be specified. Make sure if the image for the
 # service is not publicly available that you have built the docker image locally, otherwise the
 # service will fail to start.
-LOCALLY_DEPLOYED_SERVICES=harmony-service-example,harmony-netcdf-to-zarr
+LOCALLY_DEPLOYED_SERVICES=harmony-service-example,harmony-netcdf-to-zarr,giovanni-adapter
```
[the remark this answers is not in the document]
hailiangzhang: Done.

test/ogc-api-coverages/get-coverage-rangeset.ts
hailiangzhang: Yea, this is the thing, if the first step is not query-cmr, things behave differently and the related tests will be broken.
As a matter of fact, I added another test suite [when the first step is giovanni-adapter instead of query-cmr](https://github.com/nasa/harmony/pull/160/files#diff-4265cd083e74e99482c60ed86e35eb2cbddf4db892fd847bb400ebf2fe59192fR480). Maybe we can change the image name there to be `'fake-internal.earthdata.nasa.gov/nexus-service/foo:uat'`?

env-defaults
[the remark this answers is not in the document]
hailiangzhang: Yea, Chris has the same question and it's been removed ... thanks!

tasks/giovanni-adapter/app/stac/types.ts
hailiangzhang: Good point:)
But since Chris is suggesting moving the stac stuff in a separate library in the future, do you think we should defer this change until then?

tasks/giovanni-adapter/app/stac/types.ts
hailiangzhang: Same response as line 7 above:-)

env-defaults
hailiangzhang: Yes, I remember Patrick suggesting doing it later ... can you confirm @bilts ?

tasks/giovanni-adapter/app/cli.ts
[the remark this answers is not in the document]
chris-durbin: It would be analogous to CMR_ENDPOINT (defined in env-defaults) - take a look at how that is used as a configuration parameter.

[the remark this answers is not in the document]
hailiangzhang: Hmm, actually how do I get the environment from here? I know there are stuff in `app/util/env.ts` that reads `.env` and `env-defaults`, but in general they are not visible inside task runner, right?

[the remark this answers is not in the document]
chris-durbin: When we start a docker container we pass in environment variables that are specified in the harmony-env ConfigMap.

You'll need to add the environment variable to env-defaults and then add to the ConfigMap in this script:
bin/create-k8s-config-maps-and-secrets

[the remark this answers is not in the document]
hailiangzhang: Done ... thanks!

test/ogc-api-coverages/get-coverage-rangeset.ts
hailiangzhang: Done ... thanks!

tasks/giovanni-adapter/app/stac/types.ts
hailiangzhang: Got it ... fixed.

chris-durbin:
> "numInputGranules": 0,

You can sanitize what gets displayed in `app/frontend/jobs`, but I'm also thinking that number gets used elsewhere for workflow orchestration to know when a job is complete. There might be some logic that needs to change there.

PR #163, opened 2022-03-17
Harmony 1092 - Add exponential backoff retry behavior to service manager container work polling

vinnyinverso:
## Jira Issue ID
1092

## Description
I added the axios-retry library to allow axios to retry a specified number of times with exponential backoff before failing and throwing an error. 

Axios client timeouts [return an error code of ECONNABORTED](https://github.com/axios/axios/blob/e52e4dbb575fc8bd9cb7d2f5306f30ee82b40b4d/test/unit/adapters/http.js#L84) (the issue encountered in this ticket), which is not included as a retryable error in the default retry function provided by axios-retry. So I added a manual check for this condition in the retryCondition function (`error.code === 'ECONNABORTED'`).

I added an offset of 4 to the retryDelay function (`exponentialDelay(retryNumber + exponentialDelayOffest)`) so that the initial retry delay is sufficiently large (~1.5 seconds).

## Local Test Steps
1. Start services and postgres-localstack without starting Harmony
2. Check the logs of any manager container and notice that the retry/backoff behavior is being logged
3. Start Harmony (now the managers should start getting successful poll responses)
4. Run a successful request
5. Stop Harmony
6. Check the logs of any manager container and notice that the retry/backoff behavior has resumed

## PR Acceptance Checklist
* ✅ Acceptance criteria met
* ✅ Tests added/updated (if needed) and passing
* ❌ Documentation updated (if needed)


tasks/service-runner/app/workers/pull-worker.ts
```diff
@@ -25,6 +28,23 @@ const LOCKFILE_DIR = '/tmp';
...
 const maxPrimeRetries = process.env.NODE_ENV === 'test' ? 2 : 1_200;
 
+// Exponential back-off retry delay between failed requests
+axiosRetry(axios, { 
+  // delay in milliseconds = (2^retryNumber) * 100
+  retryDelay: (retryNumber) => exponentialDelay(retryNumber + exponentialDelayOffest),
+  retryCondition: 
+    (error) => {
+      if (isNetworkOrIdempotentRequestError(error) || error.code === 'ECONNABORTED') {
+        logger.error('Axios retry condition has been met.',
```
indiejames, line 38: Do we have a 'warn' log level? Might be better than 'error' here.

tasks/service-runner/app/workers/pull-worker.ts
```diff
@@ -130,7 +150,7 @@ async function _pullAndDoWork(repeat = true): Promise<void> {
         logger.debug(`Sending response to Harmony for results of work item with id ${workItem.id} for job id ${workItem.jobID}`);
         let tries = 0;
         let complete = false;
-        while (tries < maxRetries && !complete) {
+        while (tries < maxItemUpdateRetries && !complete) {
```
indiejames, line 153: Why are you still using the manual retry logic here instead of relying on `axiosRetry`?

tasks/service-runner/app/workers/pull-worker.ts
```diff
@@ -150,7 +170,7 @@ async function _pullAndDoWork(repeat = true): Promise<void> {
             }
           }
           if (!complete) {
-            if (tries < maxRetries) {
+            if (tries < maxItemUpdateRetries) {
               logger.info(`Retrying failure to update work item with id ${workItem.id} for job id ${workItem.jobID}`);
               await sleep(1000);
```
indiejames, line 175: Shouldn't we rely on the `axiosRetry` exponential backoff here instead of using `sleep`?

indiejames (changes requested):
I have some questions about why we are not using the `axios-retry` exponential backoff for work item updates as well.

tasks/service-runner/app/workers/pull-worker.ts
```diff
@@ -25,6 +28,23 @@ const LOCKFILE_DIR = '/tmp';
 // retry twice for tests and 1200 (2 minutes) for real
 const maxPrimeRetries = process.env.NODE_ENV === 'test' ? 2 : 1_200;
 
+// Exponential back-off retry delay between failed requests
+axiosRetry(axios, {
```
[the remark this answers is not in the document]
vinnyinverso: Yea we can make a custom function to calculate the delay.

tasks/service-runner/app/workers/pull-worker.ts
vinnyinverso: Sure, I'll change that.

tasks/service-runner/app/workers/pull-worker.ts
vinnyinverso: The ticket was for polling specifically so I didn't want to bite too much off.

tasks/service-runner/app/workers/pull-worker.ts
vinnyinverso: Same comment as above https://github.com/nasa/harmony/pull/163/files#r829200989 but if you think it's essential for this ticket I can add that.

vinnyinverso: Or I can also file a new ticket. Up to you.

vinnyinverso: Actually axisoRetry wraps every call, so I should probably refactor this regardless.

vinnyinverso:
@chris-durbin and @indiejames I have addressed the comments with some updates. 

The work item PUT requests now use axios-retry (in addition to work item GET requests) and each request type has its own axios instance so that it can have its own retry configuration. 

Tests are still passing, but there's now a vulnerability to look into.

indiejames (approved):

chris-durbin (approved):
Tested everything out locally. The retry interval never exceeded 90 seconds, and I was able to successfully submit a request after starting up harmony. After the 90 second retry it picked up my request and finished successfully.

PR #167, opened 2022-03-21
Harmony 1106

hailiangzhang:
## Jira Issue ID
1106

## Description
This PR fixed a minor issue in `giovanni-adapter` (adding the time properties in the stac item) which was found when testing it in sandbox.

## Local Test Steps
A giovanni request (e.g. the following) should return the giovanni URL successfully.
```
curl -Ln -bj "http://localhost:3000/C1225808238-GES_DISC/ogc-api-coverages/1.0.0/collections/%2FGrid%2FIRprecipitation/coverage/rangeset?format=text%2Fcsv&point=0.76,-3.8&subset=time%28%222020-01-01T00%3A00%3A00Z%22%3A%222020-01-01T03%3A00%3A00Z%22%29&forceAsync=true"
```

## PR Acceptance Checklist
* [X] Acceptance criteria met

### DIFF
--- a/tasks/giovanni-adapter/app/cli.ts
+++ b/tasks/giovanni-adapter/app/cli.ts
@@ -101,7 +101,7 @@ export default async function main(args: string[]): Promise<void> {
   const stacItemFilename = path.join(options.harmonyMetadataDir, stacItemRelativeFilename);
   const time_start = operation.temporal.start;
   const time_end = operation.temporal.end;
-  const properties = { start_datetime: time_start, end_datetime: time_end, datetime: null };
+  const properties = { start_datetime: time_start, end_datetime: time_end };
   const [lon, lat] = operation.spatialPoint;
   const bbox: BoundingBox = [lon, lat, lon, lat];
   const assets = {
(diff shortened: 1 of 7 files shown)


tasks/giovanni-adapter/app/cli.ts
```diff
@@ -99,6 +99,9 @@ export default async function main(args: string[]): Promise<void> {
     'title': 'giovanni stac item',
   });
   const stacItemFilename = path.join(options.harmonyMetadataDir, stacItemRelativeFilename);
+  const time_start = operation.temporal.start;
+  const time_end = operation.temporal.end;
+  const properties = { start_datetime: time_start, end_datetime: time_end, datetime: null };
```
chris-durbin, line 104: Why do we need to set `datetime` to null?

chris-durbin (commented):
It seems like the missing temporal items should have been caught by a test. Can you add a test that verifies the correct information is added to the catalog items?

tasks/giovanni-adapter/app/cli.ts
hailiangzhang: I was just trying to follow the results from other services like `netcdf-to-zarr` and honestly I don't why it was null either, but I just left it there for consistency and in case some failures that was not caught by my local deployment but showed in in sandbox.

hailiangzhang: Actually I think we may just remove `datatime` field since it's not defined in the stacItem interface. Let me try to remove it and give a try.

hailiangzhang: Just removed the `datetime` field and everything is still working. I have pushed the changed, and not sure why `netcdf-to-zarr` introduced that field.

mamadeesangaraySOS:
rescue me please

PR #169, opened 2022-03-23
Harmony 1093 handle cmr errors query cmr

indiejames:
## Jira Issue ID
HARMONY-1093

## Description
This PR changes the worker for the query-cmr service to run as a standalone service behind an http server (express) inside its container. The services manager uses http requests (only for query-cmr right now) to invoke the worker instead of a k8s `exec` request.

This PR also fixes a bug where CMR errors were not causing the work-item to be marked as failed, returning no granule instead.

## Local Test Steps
Run and build the images (npm run build) in the service-runner and query-cmr tasks directories. Then run `bin/restart-services` to restart your pods with the new images. Verify that you can make a request and it gets processed correctly.

Then delete the query-cmr deployment/service, issue a harmony query, and wait ten minutes (alternatively update the `scrollID` field in the corresponding row in the `work_items` table to something like '1234'). Then restart the query-cmr service and the query should complete with an error about a non-existent scroll-id.

## PR Acceptance Checklist
* [ x] Acceptance criteria met
* [ x] Tests added/updated (if needed) and passing
* [ x] Documentation updated (if needed)


app/models/work-item.ts
```diff
@@ -146,6 +146,18 @@ export async function getNextWorkItem(
 
     await jobQuery;
 
+    // const uquery = tx(Job.table)
```
chris-durbin, line 149: Not sure if this was added for debugging, but looks like it can be removed.

chris-durbin, line 149: Not sure if this was added for debugging, but looks like it can be removed.

env-defaults
```diff
@@ -39,6 +39,9 @@ PORT=3000
 # callbacks
 BACKEND_PORT=3001
 
+# The port on which a worker container listens for work from the manager container
+WORKER_PORT=5001
```
chris-durbin, line 43: Is this used by the main harmony application or this env needed elsewhere? Just wondering if we're putting things in env-defaults that aren't related to the main harmony app and we should split them into a different place.

chris-durbin (commented):
Just leaving a few comments, will test this out locally tomorrow.

app/models/work-item.ts
indiejames: Yeah - leftover debugging. Removed.

env-defaults
indiejames: I have this here because this is the file we use when we run `bin/deploy-services`/`bin/restart-services`. Not sure where else I can put env vars that will get picked up by those scripts.

vinnyinverso (approved):

PR #171, opened 2022-03-28
Harmony 1085 - Allow admins to set harmony log level via request

vinnyinverso:
## Jira Issue ID
HARMONY-1085

## Description
This allows you to set the log level on the fly for the Harmony frontend or backend. https://github.com/winstonjs/winston#logging-levels

~I thought adding tests for this would be overkill but if you disagree let me know and I'll add some.~
Adding tests per Chris' comment.

I added a ticket for adding the same functionality for Harmony services https://bugs.earthdata.nasa.gov/browse/HARMONY-1112

## Local Test Steps
1. Set the log level: http://localhost:3000/admin/configuration/log-level?level=error
2. Check that the logs have been filtered accordingly

> winston allows you to define a level property on each transport which specifies the maximum level of messages that a transport should log. For example, using the syslog levels you could log only error messages to the console and everything info and below to a file (which includes error messages)

## PR Acceptance Checklist
* [x]  Acceptance criteria met
* [x]  Tests added/updated (if needed) and passing
* ❌  Documentation updated (if needed)

### DIFF
--- a/app/routers/router.ts
+++ b/app/routers/router.ts
@@ -216,12 +216,12 @@ export default function router({ skipEarthdataLogin = 'false' }: RouterConfig): 
   result.get('/jobs/:jobID/cancel', cancelJob);
   result.get('/admin/jobs/:jobID/cancel', cancelJob);
 
+  result.get('/admin/configuration/log-level', setLogLevel);
+
   result.get('/cloud-access', cloudAccessJson);
   result.get('/cloud-access.sh', cloudAccessSh);
   result.get('/stac/:jobId', getStacCatalog);
   result.get('/stac/:jobId/:itemIndex', getStacItem);
-
-  result.get('/admin/configuration/log-level', setLogLevel);
 
   result.get('/*', () => { throw new NotFoundError('The requested page was not found.'); });
   result.post('/*', () => { throw new NotFoundError('The requested POST page was not found.'); });
--- a/test/helpers/configuration.ts
+++ b/test/helpers/configuration.ts
@@ -15,7 +15,7 @@ export function configureLogLevel(
   options: { username?: string; query?: object },
 ): Test {
   const { query } = options;
-  return request(app).get('/admin/configuration/log-level').query(query || {});
+  return request(app).get('/admin/configuration/log-level').query(query);
 }
 
 export const hookConfigureLogLevel = hookRequest.bind(this, configureLogLevel);
--- a/app/frontends/configuration.ts
+++ b/app/frontends/configuration.ts
@@ -26,6 +26,7 @@ export async function setLogLevel(
       throw new RequestValidationError('Must set log level using a single query parameter (level).');
     }
     const result = configureLogLevel(query.level.toLowerCase());
+    req.context.logger.info(result);
     res.json({ result });
   } catch (e) {
     req.context.logger.error(e);


chris-durbin:
I think we should add tests that the endpoint works (not necessarily that the log levels are updated).

Tests I'm thinking:

It should return a 2xx status code for an admin and accept the level parameter (and only that parameter) with only valid values.
It should return a 400 code for an an admin passing in a bogus parameter.
It should return a 400 code for an admin passing in an invalid value for the level parameter.
It should allow case insensitive log level values.
It should return a 4xx code for a non admin user attempting to change the log levels.

vinnyinverso:
I added the tests, but also got a new fixture which I'm not sure if I should add?

The request was:
`GET /access-control/groups/AG1222736883-EEDTEST/members?
`
The response was:
`+{"errors":["You do not have permission to read access control group [EEDTEST Administrator Group] in provider [EEDTEST]."]}`

chris-durbin:
> I added the tests (thanks for outlining them), but also got a new fixture which I'm not sure if I should add?

Yes, that fixture makes sense. You should add it.

app/routers/router.ts
```diff
@@ -218,6 +220,9 @@ export default function router({ skipEarthdataLogin = 'false' }: RouterConfig):
   result.get('/cloud-access.sh', cloudAccessSh);
   result.get('/stac/:jobId', getStacCatalog);
   result.get('/stac/:jobId/:itemIndex', getStacItem);
+
+  result.get('/admin/configuration/log-level', setLogLevel);
```
chris-durbin, line 224: Nit: I'd probably move the route up above with the other admin routes.

app/util/log.ts
```diff
@@ -76,6 +78,32 @@ export function createTextLogger(transports: winston.transport[]): winston.Logge
...
+ * Change the log level on the logger transport.
+ * @param level - The new log level that'll be used.
+ * @returns a string indicating the action performed
+ * @throws RequestValidationError - if the log level is not one of winston.config.npm.levels
+ */
+export function configureLogLevel(level: string): string {
+  validateLogLevel(level);
+  const currentLevel = transport.level;
+  transport.level = level;
+  return `Log level was changed from ${currentLevel} to ${level}`;
```
chris-durbin, line 104: I think we should log this message (at either warn or info?) in addition to returning it to the end user. Log it here or in the function that calls this one.

test/configuration.ts
```diff
@@ -0,0 +1,53 @@
...
+          expect(this.res.statusCode).to.equal(400);
+          expect(error).to.eql({
+            'code': 'harmony.RequestValidationError',
+            'description': 'Error: Requested to configure log level with invalid level (hi). Valid levels are: error, warn, info, http, verbose, debug, and silly.',
+          });
+        });
+      });
+    });  
+    describe('when the user is not part of the admin group', function () {
+      hookConfigureLogLevel({ username: 'tim' });
```
chris-durbin, line 47: You have to watch out for Tim.

test/helpers/configuration.ts
```diff
@@ -0,0 +1,21 @@
...
+ * @param options - Mapping object. Includes (optional) query param object (which maps query
+ * param names to values) and (optional) username.
+ * e.g. \{username: 'billy', query: \{...\}\}
+ */
+export function configureLogLevel(
+  app: Express.Application,
+  options: { username?: string; query?: object },
+): Test {
+  const { query } = options;
+  return request(app).get('/admin/configuration/log-level').query(query || {});
```
chris-durbin, line 18: I think the more idiomatic way is to set the default in the parameters - ie:

```
options: { username?: string; query?: object = {} },
```

chris-durbin (approved):
Just some minor comments.

test/helpers/configuration.ts
vinnyinverso: There was some error about an object literal not being allowed to have an initializer but I got rid of the `|| {}` as it's not really necessary anyway.

PR #176, opened 2022-03-31
Harmony 1113 - Add status and user filter to workflow UI jobs table

vinnyinverso:
## Jira Issue ID
HARMONY-1113

## Description
Added job status and job username filters to the workflow UI jobs table.

Edit: Updated based on feedback.
Screen Shot 2022-04-04 at 10 32 56 AM

## Local Test Steps
Go to http://localhost:3000/workflow-ui and play around with the filters. Prefix the route with /admin in order to enable the username filter.

## PR Acceptance Checklist
* [x] Acceptance criteria met


vinnyinverso:
@bilts Chris and I were hoping to get UI/UX feedback from you on this feature (if you have any/have the time).

Chris had also mentioned it'd be nice to see the currently selected filters (like in Kibana).

vinnyinverso:
I'm looking into using this: https://yaireo.github.io/tagify. My go-to is select2 but I'd rather not add JQuery into the mix if possible.

vinnyinverso:
Based on the feedback so far I'm aiming for this as of right now. It has autocomplete but also can allow for custom user defined tags (with optional validation) which may be useful in other cases. This is closer to what Kibana has and hopefully closer to what you were suggesting.

Screen Shot 2022-04-01 at 11 38 31 AM

indiejames:
I just tested this again and the new filter box on the side looks great. The only thing is that it seems to use the opposite of what the filter should be for me. Like this
image

vinnyinverso:
> I just tested this again and the new filter box on the side looks great. The only thing is that it seems to use the opposite of what the filter should be for me. 

Huh that's really odd. Did you click 'apply' after checking off the 'negate statuses' box? Does the query string in the url have `disallowStatus=on` after you hit 'apply'?

vinnyinverso:
I'm going to add the 'user' filter as well so I'll convert this back to draft until that's done.

chris-durbin (approved):
I think it would be good to hide the "Negate users" checkbox on the non-admin route since they can't filter on users, otherwise everything looks good.

indiejames (approved):

PR #181, opened 2022-04-12
Harmony 1131 pause job state

indiejames:
## Jira Issue ID
HARMONY-1131

## Description
This PR adds 'paused' to the `job` `status` enum. It does not add an endpoint to pause or resume jobs - that is handled in a subsequent ticket.

## Local Test Steps
There is no good way to test this locally beyond running the db migrations. There is no endpoint to pause a job yet - this will be provided in a subsequent ticket PR. There are, however, some automated tests added to our test suite.

## PR Acceptance Checklist
* [x] Acceptance criteria met


app/util/job.ts
```diff
@@ -22,6 +22,35 @@ async function cleanupWorkItemsForJobID(jobID: string, logger: Logger): Promise<
...
+ */
+export async function pauseAndSaveJob(
+  tx: Transaction,
+  job: Job,
+  finalStatus: JobStatus,
+  logger: Logger,
+  message?,
+): Promise<void> {
+  try {
+    if (job.status != JobStatus.RUNNING) {
```
chris-durbin, line 42: `pause` which is called on line 45 already does this validation - can remove it from here.

app/util/job.ts
```diff
@@ -22,6 +22,35 @@ async function cleanupWorkItemsForJobID(jobID: string, logger: Logger): Promise<
...
+      throw new ConflictError(`Cannot pause a job in the ${job.status} state.`);
+    }
+    job.pause(message);
+    await job.save(tx);
+  } catch (e) {
+    logger.error(`Error encountered for job ${job.jobID} while attempting to pause`);
```
chris-durbin, line 48: I see we are rethrowing the error - is it not being logged without this? Ideally we could get rid of this catch and throw.

app/models/job.ts
```diff
@@ -383,6 +385,43 @@ export class Job extends Record implements JobRecord {
...
+  }
+
+  /**
+   * Updates the status of a paused job to running.
+   *
+   * @throws An error if the job is not currently in the PAUSED state
+   */
+  resume(): void {
+    if (this.status != JobStatus.PAUSED) {
+      throw new ConflictError(`Job is status is ${this.status} - only paused jobs can be resumed.`);
```
chris-durbin, line 420: Extra `is`

test/jobs/pause-jobs.ts
```diff
@@ -0,0 +1,59 @@
...
+      requestId: requestid,
+      status: JobStatus.RUNNING,
+      request: 'foo',
+      numInputGranules: 10,
+      collectionIds: ['C123'],
+    });
+
+    describe('and it is resumed before pausing', function () {
+      it('throws an error', function () {
+        expect(runningJob.resume.bind(runningJob)).to.throw('Job is status is running - only paused jobs can be resumed.');
```
chris-durbin, line 22: Extra is - should be fixed in the code and then updated here.

chris-durbin (approved):
Verified the Postgres migration up and down worked fine.

app/util/job.ts
indiejames: done

app/util/job.ts
indiejames: done

app/models/job.ts
indiejames: fixed

test/jobs/pause-jobs.ts
indiejames: fixed

app/util/job.ts
```diff
@@ -22,6 +22,32 @@ async function cleanupWorkItemsForJobID(jobID: string, logger: Logger): Promise<
...
+   * @throws {@link ConflictError} if the job is already in a terminal state.
+ */
+export async function pauseAndSaveJob(
+  tx: Transaction,
+  job: Job,
+  finalStatus: JobStatus,
+  logger: Logger,
+  message?,
+): Promise<void> {
+  job.pause(message);
```
vinnyinverso, line 41: `job.pause` looks like it can throw an exception as well but is not in the try block. If that was intentional just ignore my comment.

vinnyinverso (approved):

app/util/job.ts
indiejames: Yes it was intentional, based on Chris's earlier comment.

PR #182, opened 2022-04-14
Harmony 1130 unpause

indiejames:
## Jira Issue ID
HARMONY-1130

## Description
Adds a resume endpoint for normal users and admin users. 

## Local Test Steps
* Start a long-running (async) job
* Change the state of the job to 'paused' in the database
* Connect to the job status page and verify that the job is paused
* Resume the job using the `/jobs/:jobID/resume` endpoint
* Connect to the job status page and verify that the job is running

## PR Acceptance Checklist
* [x] Acceptance criteria met
* [ ] ~Documentation updated (if needed)~ - will need to wait until auto-pause is added as there is currently no way for a user to pause their own job, therefore no way for a user to make use of the resume endpoint right now

### DIFF
--- a/app/frontends/jobs.ts
+++ b/app/frontends/jobs.ts
@@ -180,7 +180,7 @@ export async function getJobStatus(
 }
 
 /**
- * Express.js handler that cancels a single job `(POST /jobs/{jobID})`. A user can cancel their own
+ * Express.js handler that cancels a single job `(POST /jobs/{jobID}/cancel)`. A user can cancel their own
  * request. An admin can cancel any user's request.
  *
  * @param req - The request sent by the client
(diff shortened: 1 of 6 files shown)


app/frontends/jobs.ts
```diff
@@ -221,3 +221,42 @@ export async function cancelJob(
     }
   }
 }
+
+/**
+ * Express.js handler that resumes a single job `(POST /jobs/{jobID})`. A user can resume their own
```
chris-durbin, line 226: Looks like the route is /jobs/:jobID/resume

chris-durbin (approved):

app/frontends/jobs.ts
indiejames: Good catch. Copied from the `cancelJob` function above. Fixed here and there was well.

vinnyinverso (approved):

PR #183, opened 2022-04-14
HARMONY-842: Update Node to 16.13

chris-durbin:
## Jira Issue ID
HARMONY-842

## Description
Updates Node to version 16.13. I verified everything worked in my sandbox account and there is a corresponding harmony-ci-cd PR. Note that the latest version - node 16.14 was causing failures with the shapefile tests so we'll need to look into that, but for now I've locked our version to 16.13.

## Local Test Steps
```
rm -rf node_modules
nvm use
npm install
npm test
```
You can also build all of the docker images and then test locally and verify that your requests still work.

## PR Acceptance Checklist
* [X] Acceptance criteria met
* [X] Tests added/updated (if needed) and passing
* [X] Documentation updated (if needed)

PR #184, opened 2022-04-19
Harmony 481 - Increase system granule limit

vinnyinverso:
## Jira Issue ID
HARMONY-481

## Description
This increases the system granule limit (`maxGranuleLimit`).

This PR also handles requests that span multiple CMR pages (requests with over 2k granules).

Batch insertion of DB records has also been added so that work items can be inserted in bulk https://github.com/nasa/harmony/blob/harmony-481/app/models/record.ts#L120

## Local Test Steps
This branch is up to date with main so please see https://github.com/nasa/harmony/pull/183 to update your local environment to node16 if you haven't already and wish to test this PR locally.

You can test this with: 

http://localhost:3000/C1234208438-POCLOUD/ogc-api-coverages/1.0.0/collections/alt_echo_type/coverage/rangeset?maxResults=2001

If you're testing the above request, you should also run `bin/deploy-hpa` and make sure that Docker Desktop has sufficient memory (I increased mine to 4GB). You'll also need to deploy podaac-l2-subsetter (docker pull ghcr.io/podaac/l2ss-py:sit). You may also want to limit the variety of currently deployed services to free up resources for scaling of the l2-subsetter.

To only test the modified query-cmr behavior you can simply set `CMR_MAX_PAGE_SIZE` to something very low, like 5, and issue a request with `maxResults=8`, e.g. http://localhost:3000/C1233800302-EEDTEST/ogc-api-coverages/1.0.0/collections/all/coverage/rangeset?maxResults=8. In the workflow UI you should see 2 query-cmr tasks and 8 harmony-service-example tasks. 

## PR Acceptance Checklist
* [x] Acceptance criteria met


vinnyinverso:
Maybe in the getWork JSON response I should rename `res.send({ workItem, maxCmrGranules });` to something more generic like `res.send({ workItem, maxOutput });` in case future services other than the query-cmr want to take advantage of this variable?

app/backends/workflow-orchestration.ts
```diff
@@ -21,6 +22,34 @@ const RETRY_DELAY = 1000;
...
+    const nextStep = await getWorkflowStepByJobIdStepIndex(
+      tx,
+      workItem.jobID,
+      workItem.workflowStepIndex + 1,
+    );
+    if (!nextStep) {
+      logger.warn('Could not find next step for query CMR work item');
+      return;
+    }
+    const numItemsGenerated = await workItemCountForStep(tx, workItem.jobID, nextStep.stepIndex);
```
chris-durbin, line 46: Double-check this function works correctly for aggregated steps. For example if the next step is netcdf-to-zarr by default it will concatenate and so for a 5000 granule request we won't have any work items created for the next step yet, but we'll want to make sure the first query-cmr task gets 2000, the second gets 2000, and the third gets 1000. After that third request comes back, and we have all 5000 granules then we want to create a work item to send all 5000 to netcdf-to-zarr.

app/util/db.ts
```diff
@@ -6,6 +6,11 @@ import { attachPaginate } from 'knex-paginate';
 import env from './env';
 import logger from './log';
 
+/**
+ * Batch size -- to avoid overly large SQL statements.
+ */
+export const batchSize = env.nodeEnv === 'development' ? 100 : 2000;
```
chris-durbin, line 12: Ok good to see with Postgres that we could insert 2000 at a time. I assume you tested both sqlite and postgres.

app/models/record.ts
```diff
@@ -89,4 +105,41 @@ export default abstract class Record {
...
+   * record is not valid.  Records will be inserted and have their id, createdAt, and
+   * updatedAt fields set. If running SQLite, the id field will not be set as it only returns
+   * the last id.
+   *
+   * @param transaction - The transaction to use for saving the records
+   * @param records - The records to save
+   * @param fieldsList - The fields to save to the database
+   * @throws Error - if the record is invalid
+   */
+  static async insertBatch(transaction: Transaction, records: Record[], fieldsList: Partial<Record>[] = records): Promise<void> {
```
chris-durbin, line 120: Split up long line

app/models/record.ts
```diff
@@ -89,4 +105,41 @@ export default abstract class Record {
         .update(fields);
     }
   }
-}
+
+  /**
+   * Validates and saves each record (using a single statement).  Throws an error if any
+   * record is not valid.  Records will be inserted and have their id, createdAt, and
+   * updatedAt fields set. If running SQLite, the id field will not be set as it only returns
```
chris-durbin, line 112: If it doesn't set the id field, that doesn't cause any problems for sqlite?

app/models/record.ts
```diff
@@ -89,4 +105,41 @@ export default abstract class Record {
...
+        throw new TypeError(`${recordConstructor.name} is invalid: ${JSON.stringify(errors)}`);
+      }
+      setDateFields(record, fields);
+    }
+    let stmt = transaction(table).insert(fieldsList);
+    if (db.client.config.client === 'pg') {
+      stmt = stmt.returning('id'); // Postgres requires this to return the id of the inserted record
+    }
+    try {
+      const recordIds = await stmt;
```
chris-durbin, line 137: Based on your comment about sqlite it seems like this code shouldn't be called for sqlite otherwise we're setting an id for the first record, but that id should be for the last record.

app/routers/service-response-router.ts
```diff
@@ -14,6 +14,7 @@ export default function router(): Router {
   const result = Router();
   result.use(json({
     type: 'application/json',
+    limit: '500kb',
```
chris-durbin, line 17: I'm wondering what error you ran into that caused this. What happens if a callback has a body larger than 500kb - I assume the request will fail?

Since the results are mostly stored in the STAC catalog file and not sent as part of the callback I guess we shouldn't get anywhere near this limit though.

app/workers/work-reaper.ts
```diff
@@ -30,19 +31,25 @@ export default class WorkReaper implements Worker {
         db, notUpdatedForMinutes, jobStatus,
       );
       if (workItemIds.length) {
-        const numItemsDeleted = await deleteWorkItemsById(db, workItemIds);
-        this.logger.info(`Work reaper removed ${numItemsDeleted} work items`);
+        const chunkedWorkItemIds = _.chunk(workItemIds, batchSize);
+        for (const workItemIdsChunk of chunkedWorkItemIds) {
+          const numItemsDeleted = await deleteWorkItemsById(db, workItemIdsChunk);
+          this.logger.debug(`Work reaper removed ${numItemsDeleted} work items`);
```
chris-durbin, line 37: I think logging at info level was appropriate so we could see what was being cleaned up. Anything at debug level will not be logged in a deployed environment normally.

test/workflow-orchestration.ts
```diff
@@ -435,4 +439,65 @@ describe('Workflow chaining for a collection configured for swot reprojection an
       expect(workflowSteps[1].serviceID).to.equal('sds/swot-reproject:latest');
     });
   });
+});
+
+describe('Request that spans multiple CMR pages', function () {
```
chris-durbin, line 444: It looks like this test must be somehow limiting the default query cmr page size to 3. Where is that happening?

app/util/db.ts
vinnyinverso: Yup, tested with both.

app/backends/workflow-orchestration.ts
vinnyinverso: OK -- will double check.

app/models/record.ts
vinnyinverso: It just doesn't set the id in memory, so it's only a problem if you need the id's. Thankfully in our case we're just inserting the objects and not using them in the code that follows.

app/models/record.ts
vinnyinverso: Yea good point.

app/routers/service-response-router.ts
vinnyinverso: The call to the backend was too large. I forget the name of the error but the default limit on the JSON parser body size is pretty small.

vinnyinverso: I logged the actual size of the failed request body to get an idea of a reasonable setting.

chris-durbin:
I'm realizing there are a lot of other tickets in the no granule limit epic that we're going to want to have in place before we remove the system limit altogether. I think for this PR we should put the limit back in (I realize there's a lot of deleted tests that will need to be put back in) - maybe raise it up to something larger than 2000 - say 3000? That way we can see multiple scrolls for CMR, but are still protecting harmony before we add things in like the preview mode and supporting paging of multiple STAC catalogs.

chris-durbin:
> I'm realizing there are a lot of other tickets in the no granule limit epic that we're going to want to have in place before we remove the system limit altogether. I think for this PR we should put the limit back in (I realize there's a lot of deleted tests that will need to be put back in) - maybe raise it up to something larger than 2000 - say 3000? That way we can see multiple scrolls for CMR, but are still protecting harmony before we add things in like the preview mode and supporting paging of multiple STAC catalogs.

I think https://bugs.earthdata.nasa.gov/browse/HARMONY-276 is the ticket where we actually remove the system limit altogether (or potentially raise the value to something really large depending on what we decide).

chris-durbin (approved):
Tested out locally successfully. I set CMR_MAX_PAGE_SIZE=3 and verified that it correctly scrolled through the CMR results and for non-aggregated services immediately added work items to be processed while still adding more query-cmr tasks to scroll through the results.

I verified that all input granules I requested were processed for harmony-service-example, netcdf-to-zarr(aggregated) and podaac-concise(aggregated).

I also submitted a large request for 5000 granules and saw that the system limit was in place to only process up to 2100 granules.

indiejames (approved):
Tested locally

PR #191, opened 2022-04-26
HARMONY-1164: Fix inconsistent job status

chris-durbin:
## Jira Issue ID
HARMONY-1164

## Description
We've been seeing some inconsistent job status when several work items complete work for a job around the same time. Sometimes the progress percent would go down and in some instances the status would go back to running when the request should have been successful. When performing a workload run with just a single request going at a time with a random number of granules between 1 and 97 for a harmony-service-example request I would consistently see this error condition where all work items were successful, but the job was still in the running state. I'd usually see it within just a couple of requests from the start of running and never made it past 10 requests.

After this change to make sure the job is locked when updating the status after a work item completes I've been able to run a workload run in my own sandbox for hours without issue (more than 250 requests).

Note that on a couple of occasions I've seen a case where a work item stays in the running state and the job stays running. While that needs to be figured out and fixed there is a separate ticket for handling it. This ticket is just to make sure the job status is consistent with the work items.

## Local Test Steps
Testing locally is difficult because you'd need to have many granules processed in parallel. I would recommend deploying to a sandbox environment where we run 50 in parallel.

Run a workload run with just the harmony-service-example with a random number of granules:
`HTTP_PROXY=socks5h://localhost:8080 locust --tags random`

In the locust UI choose to run just 1 worker at a time (for maximum contention on work items updating for the same job and also so the run will stop as soon as one job is hung in the running state). Make sure that the requests continue to go through successfully, or if things do get stuck make sure that not all of the work items are in the successful state for that job.

## PR Acceptance Checklist
* [X] Acceptance criteria met
* ~[ ] Tests added/updated (if needed) and passing~
* ~[ ] Documentation updated (if needed) ~

### DIFF
--- a/app/models/job.ts
+++ b/app/models/job.ts
@@ -237,14 +237,27 @@ export class Job extends Record implements JobRecord {
 
   /**
   * Returns a Job with the given jobID using the given transaction
+  * Optionally locks the row.
   *
   * @param transaction - the transaction to use for querying
   * @param jobID - the jobID for the job that should be retrieved
   * @returns the Job with the given JobID or null if not found
   */
-  static async byJobID(transaction: Transaction, jobID: string): Promise<Job | null> {
-    const jobList = await this.queryAll(transaction, { where: { jobID } }, true, 0, 1);
-    return jobList.data.shift();
+  static async byJobID(
+    transaction: Transaction, jobID: string, getLinks = true, lock = false,
+  ): Promise<Job | null> {
+
+    let query = transaction('jobs').select().where({ jobID });
+    if (lock) {
+      query = query.forUpdate();
+    }
+
+    const result = await query;
+    const job = new Job(result[0]);
+    if (getLinks) {
+      job.links = (await getLinksForJob(transaction, job.jobID)).data;
+    }
+    return job;
   }
 
   /**
@@ -387,7 +400,7 @@ export class Job extends Record implements JobRecord {
 
   /**
    *  Checks the status of the job to see if the job is paused.
-   * 
+   *
    * @returns true if the `Job` is paused
    */
   isPaused(): boolean {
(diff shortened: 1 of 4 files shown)


app/models/job.ts
```diff
@@ -237,14 +237,29 @@ export class Job extends Record implements JobRecord {
...
   */
-  static async byJobID(transaction: Transaction, jobID: string): Promise<Job | null> {
-    const jobList = await this.queryAll(transaction, { where: { jobID } }, true, 0, 1);
-    return jobList.data.shift();
+  static async byJobID(
+    transaction: Transaction, jobID: string, getLinks = true, lock = false,
+  ): Promise<Job | null> {
+
+    let query = transaction('jobs').select().where({ jobID });
+    if (lock) {
```
chris-durbin, line 253: This is really the key change for the PR - Lock the row in the jobs table when necessary. We will set lock to true as part of updating the job status in a work item update.

app/models/job.ts
```diff
@@ -237,14 +237,29 @@ export class Job extends Record implements JobRecord {
...
+      query = query.forUpdate();
+    }
+
+    const result = await query;
+    const job = new Job(result[0]);
+    if (getLinks) {
```
chris-durbin, line 259: Optimization - only load the job links if they are needed from the context of where this function is called.

fixtures/cmr.uat.earthdata.nasa.gov-443/165099102895369026
```diff
@@ -0,0 +1,25 @@
+POST /search/clear-scroll
```
chris-durbin, line 1: I'm not sure where this fixture came from, but it kept getting generated for me. Running mocha now I don't see any new fixtures generated.

vinnyinverso (approved):
I like the bonus optimization, nice job.

indiejames (approved):

PR #192, opened 2022-04-26
Harmony 1139 token refresh upon resume

indiejames:
## Jira Issue ID
HARMONY-1139

## Description
Updates access token when non-admin user resumes a job.

## Local Test Steps
* Start a long running job
* Pause the job
* Check the workflow-steps table for the rows for the job to get the `accessToken` from the `operation` column.
* Log out of EDL
* Log back into EDL (not sure if this will force a new token - if not update the dB directly as described below)
* Resume the job
* Check the workflow-steps table for the rows for the job to get the `accessToken` from the `operation` column.
* See that the new accessToken is different from the previous one

Alternatively it may be easier to edit the `accessToken` value in the `operation` string directly and update the database with that, then compare that value to the value retrieved after resuming.

## PR Acceptance Checklist
* [x] Acceptance criteria met

### DIFF
--- a/app/util/job.ts
+++ b/app/util/job.ts
@@ -8,6 +8,9 @@ import { ConflictError, NotFoundError, RequestValidationError } from './errors';
 import isUUID from './uuid';
 import { clearScrollSession } from './cmr';
 import { WorkItemStatus } from '../models/work-item-interface';
+import { getWorkflowStepsByJobId } from '../models/workflow-steps';
+import DataOperation, { CURRENT_SCHEMA_VERSION } from '../models/data-operation';
+import { createDecrypter, createEncrypter } from './crypto';
 
 /**
  * Cleans up the temporary work items for the provided jobID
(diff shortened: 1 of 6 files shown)


app/util/job.ts
```diff
@@ -169,3 +122,67 @@ export function validateJobId(jobID: string): void {
...
+  const decrypter = createDecrypter(env.sharedSecretKey);
+  await db.transaction(async (tx) => {
+    const job = await lookupJob(tx, jobID, username);
+    if (username && token) {
+      // update access token
+      const workflowSteps = await getWorkflowStepsByJobId(tx, jobID);
+      for (const workflowStep of workflowSteps) {
+        const { operation } = workflowStep;
+        const op = new DataOperation(JSON.parse(operation), encrypter, decrypter);
+        op.model.accessToken = token;
```
chris-durbin, line 178: After changing this to use the setter directly I was able to successfully test locally:
```
op.accessToken = token
```

chris-durbin (approved):
Approved assuming the one code change to use the setter is made and the tests all pass.

vinnyinverso (approved):

PR #197, opened 2022-05-04
Use correct projection string in l2 example

flamingbear:
- Replaces `proj_string` with `l2proj_string` to correctly encode the query string in the l2w *Access Level 2 swath regridded data* section of the notebook. 
- fixes typo `repsonse` with `response`
- Does not reproduce the `example-notebook.html` file since the notebook can't be run to completion currently.

### DIFF
--- a/docs/Harmony API introduction.ipynb
+++ b/docs/Harmony API introduction.ipynb
@@ -78,8 +78,7 @@
    "outputs": [],
    "source": [
     "# ---- Change this to save to a directory where you have write permissions\n",
-    "local_dir = 'tmp/test_output/'\n",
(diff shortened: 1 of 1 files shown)


owenlittlejohns (approved):
Thanks for making these updates. Glad to know that the issue we found wasn't due to a bug in the Swath Projector.

General comment (mainly to the Harmony team) - I wonder if this documentation notebook should be updated to use `harmony-py` when making requests. I'm a bit biased, but I think it makes the process to make a request and download the results quite a lot simpler for the end-user.

chris-durbin (approved):

PR #201, opened 2022-05-10
HARMONY-1177: Default to binding only localhost network interface.

chris-durbin:
## Jira Issue ID
HARMONY-1177

## Description
Only bind harmony to listen on the localhost network interface.

## Local Test Steps
Start harmony and verify it logs that it is listening on ports 3000 and 3001 on 127.0.0.1.
Verify harmony works.

Note for deployed environments we will use 0.0.0.0. I tested that in sandbox and submitted a harmony-ci-cd PR. Also if for any reason the user's O/S does not support 127.0.0.1 they can override it with the HOST_BINDING environment variable.

## PR Acceptance Checklist
* [X] Acceptance criteria met

### DIFF
--- a/app/server.ts
+++ b/app/server.ts
@@ -63,9 +63,10 @@ function addRequestId(appLogger: Logger): RequestHandler {
  * listening on the provided port.
  *
  * @param port - The port the server should listen on
+ * @param hostBinding - The host network interface to bind against
  * @returns The running express application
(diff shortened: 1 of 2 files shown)


indiejames (approved):
Tested locally and was able to hit frontend/backend on localhost, but not on private IP

PR #202, opened 2022-05-12
Harmony 790 expiration dates

indiejames:
## Jira Issue ID
HARMONY-790

## Description
Adds expiration date to job status page and STAC items

## Local Test Steps
* Issue an async Harmony request
* Go to the job status page and verify that the status includes a `dataExpiration` field that is 30 days in the future
* Append a '/0' to the end of the STAC catalog link to get the first STAC item
* Verify that the STAC item contains a a `stac_extensions` field with the url of the 'timestamps' extension and that it contains a `exprires` field in the `properties` section.

## PR Acceptance Checklist
* [x] Acceptance criteria met


chris-durbin:
There are a bunch of files in `example/workflow-stac-examples` that we should update to show the new stac_extensions and expires field.

app/models/job.ts
```diff
@@ -11,6 +11,9 @@ import Record from './record';
 import { Transaction } from '../util/db';
 import JobLink, { getLinksForJob, JobLinkOrRecord } from './job-link';
 
+// how long data generated by this job will be available
+export const EXPIRATION_DAYS = 30;
```
chris-durbin, line 15: Any reason to not make this an environment variable? Are you thinking there might be an issue if it's set differently from when the job was first run?

app/models/job.ts
```diff
@@ -806,4 +812,17 @@ export class Job extends Record implements JobRecord {
     const links = this.links.filter((link) => link.rel === rel);
     return links.map(removeEmptyProperties) as JobLink[];
   }
+
+  /**
+   *  Computes and returns the date the data produced by the job will expire based on `createdAt`
```
chris-durbin, line 817: When I was testing this I immediately started thinking - should this instead be based on when the last file output file completed? I'm just thinking about future jobs that could take days or weeks to run.

I'm not sure what we would need to do to make sure the data doesn't get cleaned up before then - but I'm guessing we'll have to figure that out at some point to be able to extend expirations. Maybe revisit then?

test/helpers/jobs.ts
```diff
@@ -15,7 +15,7 @@ import { RecordConstructor } from '../../app/models/record';
 export const adminUsername = 'adam';
 
 export const expectedJobKeys = [
-  'username', 'status', 'message', 'progress', 'createdAt', 'updatedAt', 'links', 'request', 'numInputGranules', 'jobID',
+  'username', 'status', 'message', 'progress', 'createdAt', 'updatedAt', 'links', 'request', 'numInputGranules', 'jobID', 'dataExpiration',
```
chris-durbin, line 18: I know this is the test, but for the source code as well I think we should change the order of the fields so we have all the times together `createdAt`, `updatedAt`, and then `dataExpiration`.

chris-durbin (approved):
Tested out successfully locally. Both the job status and the STAC catalog items have the appropriate expiration fields.

app/models/job.ts
indiejames: I considered making it an environment variable, but rejected it for these reasons
1. The AC says "hard coded to 30 days"
2. The 30 days is being enforced by a lifecycle policy no the bucket, not by code that would read an environment variable
3. We already have too many environment variables so I didn't want to add another without good reason

app/models/job.ts
indiejames: Yeah - I think it's important to have a single "guaranteed to last this long" date that should be good for the whole job, not just parts. Any data produced would last "at least" as long as this.

test/helpers/jobs.ts
indiejames: Done

vinnyinverso (approved):

vinnyinverso:
Just double checking -- is there no documentation that shows example Harmony responses that needs to be updated?

indiejames:
> 

Done

PR #204, opened 2022-05-16
Harmony 1138 - Add state change buttons (pause, cancel, etc.) to the UI

vinnyinverso:
## Jira Issue ID
HARMONY-1138, HARMONY-1137

## Description
1137 adds relevant job state change buttons to /workflow-ui/:jobID and /admin/workflow-ui/:jobID. There is also now an alert that shows the job message/status as the state of the job is being changed (see screenshot).

1138 adds the cancel link to the job status JSON:
```
{
  "username": "vinverso",
  "status": "paused",
  "message": "The job is paused. The job may be resumed using the provided link.",
  "links": [
    {
      "title": "Cancels the job.",
      "href": "http://localhost:3000/jobs/18ef672c-9e61-45a8-862b-926c3884c341/cancel",
      "type": "application/json",
      "rel": "canceler"
    },
    {
      "title": "Resumes the job.",
      "href": "http://localhost:3000/jobs/18ef672c-9e61-45a8-862b-926c3884c341/resume",
      "type": "application/json",
      "rel": "resumer"
    },
    {
      "href": "http://localhost:3000/jobs/18ef672c-9e61-45a8-862b-926c3884c341?page=1&limit=2000",
      "title": "The current page",
      "type": "application/json",
      "rel": "self"
    }
  ], ...
}
```

## Local Test Steps
Submit a large request for a job so that the preview mode will be activated. You should see relevant buttons in the top right corner of the screen that allow you to change job state.

Screen Shot 2022-05-16 at 9 48 40 AM

Also check that the cancel link is present at http://localhost:3000/jobs/:jobID

## PR Acceptance Checklist
* [x] Acceptance criteria met


app/models/job.ts
```diff
@@ -193,6 +193,26 @@ export const statesToDefaultMessages: any = Object.values(stateMachine.states).r
...
 
+/**
+ * Validate that a desired transition (for job status) is acceptable according to the state machine.
+ * @param currentStatus - the current job status
+ * @param desiredStatus - the desired job status
+ * @param event - the event that would precipitate the transition
+ * @param errorMessage - the error message to throw if the transition is invalid
+ * @throws ConflictError if the transition is invalid
+ */
+export function validateTransition(
```
chris-durbin, line 204: Nice I was hoping we'd refactor this out.

public/js/workflow-ui/nav-links.js
```diff
@@ -0,0 +1,119 @@
...
+
+/**
+ * Responds to a nav link click event
+ * (hits relevant Harmony url, shows user the response).
+ * @param {Event} event - the click event
+ * @param {object} broker - pubsub broker
+ */
+async function handleClick(event, broker) {
+  event.preventDefault();
+  toasts.showUpper('Changing job state..');
```
chris-durbin, line 30: Either one too many or one too few dots!

chris-durbin (approved):
Tested everything successfully for the user and admin workflow-ui routes. The changes are great. I also love the colors and message showing up at the top of the job details page. Fantastic job and thanks for adding those tests for the links that are returned and testing out everything for job sharing.

app/views/workflow-ui/jobs/index.mustache.html
```diff
@@ -103,7 +100,6 @@
             </div>
         </div>
     </footer>
-    <script type="text/javascript" src="https://status.earthdata.nasa.gov/assets/banner_widget.js"></script>
```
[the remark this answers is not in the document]
vinnyinverso: I removed it but it's still on the main page. Got annoyed that it constantly takes up so much real estate. I can add it back if you want.

app/util/links.ts
```diff
@@ -71,3 +73,117 @@ export function getStatusLink(urlRoot: string, jobID: string, rel: string): Link
...
+ * to a user for a particular job.
+ * Note that this only returns actions that users can precipitate via state change links
+ * (e.g. JobEvent.FAIL will not be returned).
+ * @param job - the job to return valid actions (JobEvents) for
+ * @returns a set of JobEvent
+ */
+export function getLinkRelevantJobEvents(job: Job): Set<JobEvent> {
+  const transitions: [JobEvent, JobStatus][] = [
+    // [event, resultant status]
+    [JobEvent.CANCEL, JobStatus.CANCELED],
```
indiejames, line 135: I don't like how we are somewhat leaking logic out of the state machine and into the code here. I would like it better if we could leverage the state machine definition to simply check all state/event/state transitions here to determine what is viable. I realize that there are some events that don't make sense for a user and aren't captured by the state machine (yet), so you don't need to change this, but we should give this some thought down the line.

app/util/links.ts
```diff
@@ -71,3 +73,117 @@ export function getStatusLink(urlRoot: string, jobID: string, rel: string): Link
...
+    // are only interested in it when the job is currently previewing.
+    // e.g. skipping preview is valid for a paused job but doesn't make sense to a user
+    transitions.push([ JobEvent.SKIP_PREVIEW, JobStatus.RUNNING ]);
+  }
+  const validEvents = new Set<JobEvent>();
+  for (const [event, newStatus] of transitions) {
+    try {
+      validateTransition(job.status, newStatus, event);
+      validEvents.add(event);
+    } catch {}
```
indiejames, line 150: Empty catch blocks are usually a bad sign (I know, I'm guilty too) as they indicate that you are using try/catch for normal control flow not just exceptional things. One good solution might be to refactor the `vaidateTransition` function into two functions, one that returns an error message if there is an error (or null if not) and one that wraps the other functions and throws an exception if the error message is not null. Then you could just call the first function here and check it's return value.

indiejames (approved):
Some minor comments

app/util/links.ts
vinnyinverso: I think I get what you're saying. The job.ts class (pause method for example) `validateTransition(this.status, JobStatus.PAUSED, JobEvent.PAUSE)` seems to be doing the same type of thing that we're doing here, so is that leaking as well? Or is the fact that this logic is not staying in the job.ts class causing the "leak"? I could easily move this list of transitions and this method for that matter back into that job class if that helps. Also, not sure if it's relevant but maybe this could be useful for encapsulating more logic in the state machine https://xstate.js.org/docs/guides/guards.html#guards-condition-functions

Even better, would the goal be to ideally move _all_ of this to the state machine (rather than simply the job class)? Just trying to make sure I understand.

app/util/links.ts
vinnyinverso: Thanks for the explanation, fixed!

app/util/links.ts
vinnyinverso: OK, sounds good.

PR #212, opened 2022-06-03
HARMONY-1187: Update retry settings for work item polling and updates

vinnyinverso:
## Jira Issue ID
HARMONY-1187

## Description
Tweaks backoff settings for work item polling and updates per our tech meeting conversation:
- adds an env var for max retries
- sets a max of 1 min delay between retries
- ensures PUT /work can succeed even if a deployment is blocking

## Local Test Steps
Set MAX_GET_WORK_RETRIES, and MAX_PUT_WORK_RETRIES to something reasonable (e.g. 5).

GET /work backoff:
With services running, stop the Harmony server. This should trigger back off behavior (for polling) which should be logged in the manager container of all services.

PUT /work backoff:
In order to test the back off behavior for work item updates add some code like this at the top of the updateWorkItem function in workflow-orchestration.ts:
```
res.status(500).send();
return;
```
Start up the harmony server, make a request, and check the query-cmr manager logs.

## PR Acceptance Checklist
* [x] Acceptance criteria met


indiejames (approved):

chris-durbin (commented):
Tested out locally - we'll get rid of `MAX_GET_WORK_RETRIES` since that just resets the retry count and causes more frequent retries (it does not kill the pod).

We need to add:
```
  MAX_PUT_WORK_RETRIES: "${MAX_PUT_WORK_RETRIES}"
```
to `bin/create-k8s-config-maps-and-secrets` so that we can override that configuration parameter.

I verified that `MAX_PUT_WORK_RETRIES` works as expected and gives up on updating the status for the work item after the retries are exhausted and polls for the next item to work on.

vinnyinverso:
I ended up putting the environment variable in the main harmony environment defaults file. This is because when we are creating the config map secrets (`bin/create-k8s-config-maps-and-secrets`) it needs to read `MAX_PUT_WORK_RETRIES` 
 from either the main env-defaults file or the main .env file (as these take precedence over the defaults file for the service runner). 

If the value is only specified in `tasks/service-runner/env-defaults` you end up with `MAX_PUT_WORK_RETRIES=''`.

I also retested everything with these changes.

chris-durbin (approved):

PR #214, opened 2022-06-13
HARMONY-1192: Change regexes to support non-alpha characters in dims

indiejames:
## Jira Issue ID
HARMONY-1192

## Description
Adds support for non-alphanumeric characters in dimension names, e.g., '/lev' or 'my lev'.

## Local Test Steps
Issue the following query and verify the result. (Not sure why this query is treated as async when it only has one granule):
```
http://localhost:3000/C1215726323-GES_DISC/ogc-api-coverages/1.0.0/collections/all/coverage/rangeset?forceAsync=true&subset=lat%28-10%3A10%29&subset=lon%28-160%3A-75%29&subset=%2Flev%28800%3A900%29&maxResults=1
```
The job should run and the workflow step in the database should have an operation that includes

```json
"subset":{"dimensions":[{"name":"/lev","min":800,"max":900}
```

## PR Acceptance Checklist
* [x] Acceptance criteria met

### DIFF
--- a/app/frontends/ogc-coverages/util/subset-parameter-parsing.ts
+++ b/app/frontends/ogc-coverages/util/subset-parameter-parsing.ts
@@ -4,20 +4,20 @@ import { ParameterParseError } from '../../../util/parameter-parsing';
 const rangeSeparator = ':';
 const unbounded = '*';
 // Regex to match lat(-10:10) or lon(*:20)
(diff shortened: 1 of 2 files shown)


vinnyinverso (approved):
The query is async because forceAsync=true.

indiejames:
> The query is async because forceAsync=true.

Good catch - I missed that when I copied the query over from the Jira ticket.

chris-durbin (approved):

PR #216, opened 2022-06-17
HARMONY-1200: Remove jobs table locking when getting work and updating work items

indiejames:
## Jira Issue ID
HARMONY-1200

## Description
Changes to fix work requests from returning no data even when work items are available. Should still preserve fair queuing for the most part (best effort, not guaranteed).

## Local Test Steps
* Deploy to sandbox and issue a long-running request
* Verify that the workers are getting work items (no long periods of 'polling for work' in the logs)
* Issue on or more additional short requests
* Verify that the new requests are worked in parallel with the long-running one 

## PR Acceptance Checklist
* [x] Acceptance criteria met

### DIFF
--- a/app/backends/workflow-orchestration.ts
+++ b/app/backends/workflow-orchestration.ts
@@ -349,7 +349,7 @@ export async function updateWorkItem(req: HarmonyRequest, res: Response): Promis
   logger.info(`Updating work item for ${id} to ${status}`);
   await db.transaction(async (tx) => {
     const workItem = await getWorkItemById(tx, parseInt(id, 10));
-    const job: Job = await Job.byJobID(tx, workItem.jobID, false, true);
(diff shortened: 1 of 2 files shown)


chris-durbin (approved):
Tested successfully in sandbox. Fair queueing still worked well - had 2 requests for cdurbin, and 1 request for cdurbin_test and the work items were worked roughly:

25% cdurbin job1
25% cdurbin job2
50% cdurbin_test only job

I also checked the database metrics and saw 46% CPU usage with a capacity size of 4 units. Prior to this change running the same test would result in 80-90% CPU usage and a capacity size of 16, so this is making a dramatic improvement in our database utilization (and will reduce costs).

I also verified in all cases while I was tailing the logs "Polling for work" was only logged once immediately followed by executing the work (verifying the ticket itself).

vinnyinverso (approved):

PR #225, opened 2022-07-08
Reposition GDAL service to bottom of list

frankinspace:
## Jira Issue ID

N/A

## Description

Moved GDAL service definition to the bottom of the services list. This was done primarily so that HOSS would be prioritized over GDAL when subsetting. Further discussion can be found in this slack thread https://eosdis.slack.com/archives/C01DC2R2TRR/p1657303101081809

## Local Test Steps

Replaced local copy of services.yml with this modified version. Ran a subset query using UAT MUR collection C1238621141-POCLOUD which is associated to both GDAL and HOSS. Observed that HOSS was used for subsetting instead of GDAL

## PR Acceptance Checklist
* [ ] Acceptance criteria met


config/services.yml
```diff
@@ -22,40 +22,7 @@ x-turbo-config: &default-turbo-config
       FALLBACK_AUTHN_ENABLED: !Env ${FALLBACK_AUTHN_ENABLED}
 
 https://cmr.earthdata.nasa.gov:
-  - name: nasa/harmony-gdal-adapter
```
owenlittlejohns, line 25: It looks like this PR removes HGA from production. I was just expecting this entry to shift down to the bottom of the production entries, in the same way the UAT HGA service entry has.

frankinspace:
Is there any documentation update needed? I don't think so

config/services.yml
frankinspace: The diff came out very confusing for some reason. If you look at the file itself you can see it is still in both places.
OPS: https://github.com/nasa/harmony/blob/e826ee27688fc658d487fc350033e9010d01cf0b/config/services.yml#L412
UAT: https://github.com/nasa/harmony/blob/e826ee27688fc658d487fc350033e9010d01cf0b/config/services.yml#L870

frankinspace: I added comments above the entries to make the diff a little bit more clear

owenlittlejohns: Ah, yeah. That diff was misleading. Looks good to me, then!

owenlittlejohns (approved):

PR #233, opened 2022-07-18
HARMONY-701: Adds an ignoreErrors=true parameter to continue processing a job even if some work items fail

chris-durbin:
## Jira Issue ID
HARMONY-701

## Description
Adds an `ignoreErrors=true` parameter to continue processing a job even if some work items fail. There are a couple of caveats:
1.) If a queryCmr request fails the job will still fail
2.) If more than a configurable number of items fail the job is marked as failed - currently set to 2000. This was needed to avoid having to worry about paging through errors in the job status response (note it would need a second set of paging parameters for the job status endpoint since we already page the job links).

## Local Test Steps
Test out the database migrations with Postgres with:
NODE_ENV=production DATABASE_URL=<postgres URL> knex --cwd db migrate:latest

To simulate some random failures:
1.) Replace line 529 of `workflow-orchestration.ts`:
```
const continueProcessing = await handleFailedWorkItems(tx, job, workItem, thisStep, status, logger, errorMessage);
```
with 
```
const failHalfTheTime =  Math.floor(Math.random() * 2);
let theStatus = status;
if (failHalfTheTime == 1) {
  theStatus = WorkItemStatus.FAILED;
  workItem.errorMessage = 'You failed me');
}

const continueProcessing = await handleFailedWorkItems(tx, job, workItem, thisStep, theStatus, logger, errorMessage);
```
2.) Submit a request with 10 items or so with ignoreErrors = true
http://localhost:3000/C1233800302-EEDTEST/ogc-api-coverages/1.0.0/collections/all/coverage/rangeset?outputCrs=EPSG%3A4326&format=image%2Fpng&ignoreErrors=true&maxResults=10

If the request fails on query-cmr verify that the job is marked as failed.
If the request fails on some items after query-cmr finishes successfully make sure the job is marked as complete_with_errors and has an errors field in the job status with details on the errors.

You can also change the code to fail all non query-cmr items to fail and verify that the job is marked as failed.

## PR Acceptance Checklist
* [X] Acceptance criteria met

### DIFF
--- a/app/models/job.ts
+++ b/app/models/job.ts
@@ -19,12 +19,13 @@ import env = require('../util/env');
 const { awsDefaultRegion } = env;
 
 const serializedJobFields = [
-  'username', 'status', 'message', 'progress', 'createdAt', 'updatedAt', 'dataExpiration', 'links', 'request', 'numInputGranules', 'jobID',
+  'username', 'status', 'message', 'progress', 'createdAt', 'updatedAt', 'dataExpiration',
+  'links', 'request', 'numInputGranules', 'jobID',
 ];
 
 export const jobRecordFields = [
-  'username', 'status', 'message', 'progress', 'createdAt', 'updatedAt', 'request', 'numInputGranules',
-  'jobID', 'requestId', 'batchesCompleted', 'isAsync',
+  'username', 'status', 'message', 'progress', 'createdAt', 'updatedAt', 'request',
+  'numInputGranules', 'jobID', 'requestId', 'batchesCompleted', 'isAsync', 'ignoreErrors',
 ];
 
 const stagingBucketTitle = `Results in AWS S3. Access from AWS ${awsDefaultRegion} with keys from /cloud-access.sh`;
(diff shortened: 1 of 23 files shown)


app/models/job.ts
```diff
@@ -633,6 +655,19 @@ export class Job extends Record implements JobRecord {
     this.updateStatus(JobStatus.SUCCESSFUL, message);
   }
 
+  /**
+   * Updates the status to complete_with_errors, providing the optional message. Generally you
+   * should only set a message if there is information to provide to users about the result, as
+   * providing a message will override any prior message, including warnings.
+   * You must call `#save` to persist the change
+   *
+   * @param message - (optional) a human-readable success message.  See method description.
```
indiejames, line 664: s/success/status

chris-durbin: Done

test/ignore-errors.ts
```diff
@@ -0,0 +1,461 @@
...
+
+  describe('when making a request for a single granule and it completes successfully', function () {
+    hookRangesetRequest('1.0.0', collection, 'all', { query: { ...reprojectAndZarrQuery, ...{ maxResults: 1 } } });
+    hookRedirect('joe');
+    hookClearScrollSessionExpect();
+
+    before(async function () {
+      const res = await getWorkForService(this.backend, 'harmonyservices/query-cmr:latest');
+      const { workItem, maxCmrGranules } = JSON.parse(res.text);
+      expect(maxCmrGranules).to.equal(1);
```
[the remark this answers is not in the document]
chris-durbin: This is similar setup as another test. It's just doing a sanity check which I guess makes sense.

PR #234, opened 2022-07-26
Harmony 1240 search after

indiejames:
## Jira Issue ID
HARMONY-1240

## Description
Replace use of the CMR scroll API with the search after API

## Local Test Steps
* Check out this branch and start Harmony
* Execute a query
* Verify the results are what you expect
* Look in your local staging bucket under the `SearchParams` folder and verify that you see a folder with a UUID for a name. Under that folder you should see a file named `serializedQuery` - this should have the JSON for your search request. This should work with shapefiles as well. The shapefile is already uploaded to S3, so you should just see a `geojson` field in the search JSON pointing to the url of the shapefile in S3.

## PR Acceptance Checklist
* [x] Acceptance criteria met


app/util/object-store.ts
```diff
@@ -212,6 +228,20 @@ export class S3ObjectStore {
...
+   * object
+   *
+   * @param paramsOrUrl - a map of parameters (Bucket, Key) indicating the object to
+   *   be retrieved or the object URL
+   * @returns a string containing the contents of the object
+   * @throws TypeError - if an invalid URL is supplied
+   */
+  async download(paramsOrUrl: string | BucketParams): Promise<string> {
+    const getObjectResponse = this.getObject(paramsOrUrl);
+    return streamToString(getObjectResponse.createReadStream());
```
chris-durbin, line 242: Don't you need to `await` on this call?

db/migrations/20220721152441_increase_size_of_scroll_id_column.js
```diff
@@ -0,0 +1,15 @@
...
+  const result = knex.schema
+    .alterTable('work_items', (t) => {
+      t.string('scrollID', 4096).alter();
+    });
+
+  return result;
+};
+
+exports.down = function down(knex) {
+  return knex.schema
```
chris-durbin, line 11: I think in this case I wouldn't allow a rollback to reduce the column size since it will likely break if there are any IDs created greater than 32 characters. I would just put a comment explaining why the rollback does nothing.

fixtures/cmr.uat.earthdata.nasa.gov-443/165876324053364067
```diff
@@ -0,0 +1,30 @@
+POST /search/granules.json
```
chris-durbin, line 1: I'm assuming there are a lot of fixtures for CMR calls that should be deleted as a result of this change.

package.json
```diff
@@ -9,7 +9,7 @@
     "coverage": "nyc mocha",
     "start": "ts-node -r tsconfig-paths/register app/server.ts",
     "start-dev": "strict-npm-engines && ts-node-dev --no-notify -r tsconfig-paths/register --watch app/views --respawn app/server",
-    "start-dev-fast": "TS_NODE_TRANSPILE_ONLY=true ts-node-dev --no-notify -r tsconfig-paths/register --respawn app/server",
+    "start-dev-fast": "TS_NODE_TRANSPILE_ONLY=true ts-node-dev --no-notify -r tsconfig-paths/register --respawn  --respawn --inspect=127.0.0.1:9229 app/server",
```
chris-durbin, line 12: What's the difference between this and the `watch-debug` command? Can we get rid of the `nodemon` dev dependency we use for watch-debug?

tasks/query-cmr/test/query-granules.ts
```diff
@@ -167,36 +148,27 @@ describe('query#queryGranulesScrolling', function () {
...
-    hookQueryGranulesScrolling(3000);
+    hookQueryGranules(3000);
 
-    it('the STAC output is not limitted', function () {
+    it('the STAC output is not limited', function () {
       expect(this.result[1].length).to.equal(3);
     });
   });
-
-  describe('when the CMR returns an error', async function () {
```
chris-durbin, line 194: I think we should still have a test where the CMR returns an error that the error message is returned.

test/workflow-orchestration.ts
```diff
@@ -660,4 +658,103 @@ describe('When a request spans multiple CMR pages', function () {
...
+      });
+
+      describe('when the CMR hits changes', async function () {
+
+        before(async function () {
+          const res = await getWorkForService(this.backend, 'harmonyservices/query-cmr:latest');
+          const { workItem } = JSON.parse(res.text);
+          workItem.status = WorkItemStatus.SUCCESSFUL;
+          workItem.hits = finalNumInputGranules;
+          workItem.results = [
```
chris-durbin, line 731: These files don't exist anymore - there's a new way to pass back the catalogs - take a look at the updates to the other tests in this file.

test/workflow-orchestration.ts
```diff
@@ -660,4 +658,103 @@ describe('When a request spans multiple CMR pages', function () {
       });
     });
   });
+
+  describe('and the number of granules returned by the CMR changes while paging', function () {
```
chris-durbin, line 662: I'm more interested in seeing that the request queues the right things and eventually completes the request.

I'd change to a small number of granules like initial 5 and then updated to 2 and change the page size to something small and then simulate through work items completing.

Also add a test going in the other direction starting with 2 and then finishing with 3.

I'm thinking the case of a synchronous request with one granule getting another granule is unlikely so if we blow up when that happens it's ok.

bin/create-k8s-config-maps-and-secrets
```diff
@@ -38,6 +38,7 @@ data:
   OAUTH_REDIRECT_URI: "${OAUTH_REDIRECT_URI}"
   FALLBACK_AUTHN_ENABLED: "${FALLBACK_AUTHN_ENABLED}"
   CMR_ENDPOINT: "${CMR_ENDPOINT}"
+  CMR_MAX_PAGE_SIZE: "${CMR_MAX_PAGE_SIZE}"
```
chris-durbin, line 41: I'm a little worried about this being set as an environment variable in the query-cmr service. I thought each query-cmr work item is supposed to say how many granules to pull back in order to handle maxResults correctly and not pull back too many items.

app/backends/workflow-orchestration.ts
```diff
@@ -517,7 +522,11 @@ export async function updateWorkItem(req: HarmonyRequest, res: Response): Promis
     );
     const allWorkItemsForStepComplete = (completedWorkItemCount == thisStep.workItemCount);
 
-    await maybeClearScrollSession(workItem.scrollID, allWorkItemsForStepComplete, status);
+    if (hits && job.numInputGranules != hits) {
```
chris-durbin, line 525: There's a complication with the `maxResults` parameter since the hits count will always end up being higher than numInputGranules if we've restricted it with the parameter. I think we can solve it by only checking if the number of hits is less than `job.numInputGranules`. So we'll never allow processing more granules than the initial numInputGranules (no different than today with scrolling), but we will successfully complete a request if there end up being fewer granules because some were deleted from CMR while the request was processing.

chris-durbin (changes requested):

test/workflow-orchestration.ts
chris-durbin: Note from my other comment that I think we still need the test of 2 going to 3, but the request should only process the first 2 granules and complete rather than trying to process the 3rd granule.

bin/create-k8s-config-maps-and-secrets
chris-durbin: @vinnyinverso - can you review to figure out if the paging through items and limiting to maxResults still works the same?

app/backends/workflow-orchestration.ts
indiejames: Done.

app/util/cmr.ts
```diff
@@ -540,27 +539,71 @@ export async function getVariablesForCollection(
 }
 
 /**
- * Queries and returns the CMR JSON granules for the given collection ID with the given query
- * params.  Uses multipart/form-data POST to accommodate large queries and shapefiles.
+ * Generate an s3 url to use to store/lookup stored query parameters
+ * 
+ * @param sessionKey - The session key
+ * @returns
```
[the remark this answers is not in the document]
indiejames: Fixed

app/util/object-store.ts
indiejames: No, since `streamToString` also returns a `Promis<string>`.

db/migrations/20220721152441_increase_size_of_scroll_id_column.js
indiejames: Done.

tasks/query-cmr/test/query-granules.ts
indiejames: Done

test/workflow-orchestration.ts
indiejames: Fixed

bin/create-k8s-config-maps-and-secrets
indiejames: this is just used as the upper limit - the actual number used is set on the request
```
export async function queryGranulesWithSearchAfter(
  token: string,
  limit = cmrMaxPageSize,
  query?: CmrQuery,
  sessionKey?: string,
  searchAfterHeader?: string,
): Promise<CmrGranuleHits> {
  const baseQuery = {
    page_size: Math.min(limit, cmrMaxPageSize),
  };
  ```

indiejames: I can probably remove this if you like as I don't think we are ever don't pass in the limit.

indiejames: Done

vinnyinverso: > @vinnyinverso - can you review to figure out if the paging through items and limiting to maxResults still works the same?

It looks like it'll be fine (also talked to James about it). Going to test it out locally now too.

test/workflow-orchestration.ts
indiejames: Done

fixtures/cmr.uat.earthdata.nasa.gov-443/165876324053364067
indiejames: Cleaned up

package.json
indiejames: This lets me start the debugger on a specific port (don't know if that is possible with watch-debug) so I can debug harmony, service runner, and query-cmr all from one vscode session (each gets its own debug port).

chris-durbin (approved):

vinnyinverso (approved):

PR #251, opened 2022-08-22
Harmony 1238 stuck jobs bug

indiejames:
## Jira Issue ID
HARMONY-1238

## Description
Makes updates to work-items asynchronous to prevent services having to wait. This allows locking the jobs table to avoid race conditions on updates since the work-item update can take longer without bogging down services.

## Local Test Steps
Can't force the race condition this is supposed to eliminate - best attempt is to run a few really long queries and make sure they don't get in a state where all the work items are done, but the job is not.

## PR Acceptance Checklist
* [x] Acceptance criteria met

### DIFF
--- a/app/backends/workflow-orchestration.ts
+++ b/app/backends/workflow-orchestration.ts
@@ -604,7 +604,12 @@ export async function updateWorkItem(req: HarmonyRequest, res: Response): Promis
     { workItemID: id, status, hits, serializedResults: JSON.stringify(results), scrollID, errorMessage, totalGranulesSize },
   );
 
-  // asynchronously handle the update
+  // asynchronously handle the update so that the service is not waiting for a response
+  // during a potentially long update. If the asynchronous update fails the work-item will
+  // eventually be retried by the timeout handler. In any case there is not much the service
+  // can do if the update fails, so it is OK for us to ignore the promise here. The service
+  // can still retry for network or similar failures, but we don't want it to retry for things
+  // like 409 errors.
   // eslint-disable-next-line @typescript-eslint/no-floating-promises
   handleWorkItemUpdate(update, req.context.logger);
 
--- a/app/backends/workflow-orchestration.ts
+++ b/app/backends/workflow-orchestration.ts
@@ -19,7 +19,7 @@ import JobError, { getErrorCountForJob } from '../models/job-error';
 import WorkItemUpdate from '../models/work-item-update';
 
 const MAX_TRY_COUNT = 1;
-const RETRY_DELAY = 1000;
+const RETRY_DELAY = 1000 * 120;
 const QUERY_CMR_SERVICE_REGEX = /harmonyservices\/query-cmr:.*/;
 
 /**
@@ -424,7 +424,7 @@ async function handleFailedWorkItems(
     continueProcessing = job.ignoreErrors;
     if (!job.isComplete()) {
       let jobMessage;
-      
+
       if (errorMessage) {
         jobMessage = `WorkItem [${workItem.id}] failed with error: ${errorMessage}`;
       }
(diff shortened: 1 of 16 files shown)


app/backends/workflow-orchestration.ts
```diff
@@ -590,8 +584,30 @@ export async function updateWorkItem(req: HarmonyRequest, res: Response): Promis
...
+export async function updateWorkItem(req: HarmonyRequest, res: Response): Promise<void> {
+  const { id } = req.params;
+  const { status, hits, results, scrollID, errorMessage } = req.body;
+  const totalGranulesSize = req.body.totalGranulesSize ? parseFloat(req.body.totalGranulesSize) : 0;
+
+  const update = new WorkItemUpdate(
+    { workItemID: id, status, hits, serializedResults: JSON.stringify(results), scrollID, errorMessage, totalGranulesSize },
+  );
+
+  // asynchronously handle the update
```
chris-durbin, line 607: Give us more info on the comment on why we are doing this - what problem are we looking to address? What happens if the asynchronous update fails, and why is it ok that we ignore it?

app/models/work-item-interface.ts
```diff
@@ -63,6 +63,35 @@ export interface WorkItemRecord {
   createdAt: Date;
 }
 
+/**
+ * Interface the defines the fields in a work-item update from a service
```
chris-durbin, line 67: Interface *that* defines

app/models/work-item-interface.ts
```diff
@@ -63,6 +63,35 @@ export interface WorkItemRecord {
   createdAt: Date;
 }
 
+/**
```
chris-durbin, line 66: I just realized - I don't think this interface or the database record are actually used (no SQL updates or migrations in here). I would just leave those files in a branch for you somewhere if you think we might want to track the updates in a database table at some point. Pushing them to main will make people think this code is being used.

chris-durbin (commented):
I haven't tested yet.

app/models/work-item-interface.ts
indiejames: I thought you said to leave them in when we talked about it the other day. I'll pull them out.

app/backends/workflow-orchestration.ts
indiejames: Done.

test/work-items/work-backends.ts
```diff
@@ -281,13 +281,12 @@ describe('Work Backends', function () {
...
-              this.job.status =
-                this.workItem.status = updateState;
-              this.response = await updateWorkItem(this.backend, this.workItem);
+              this.workItem.status = updateState;
+              await updateWorkItem(this.backend, this.workItem);
             });
-            it('fails the update', function () {
-              expect(this.response.status).to.equal(409);
-              expect(this.response.text).to.equal(`Job was already ${terminalState}.`);
+            it('fails the update', async function () {
```
vinnyinverso, line 287: Not sure if "fails the update" accurately describes what's being tested here anymore?

vinnyinverso: Never mind, I think I get the gist!

indiejames: I just re-worded it to "fails to update the work item"

app/models/work-item-interface.ts
indiejames: Done.

app/models/work-item-interface.ts
indiejames: Removed.

vinnyinverso (approved):

chris-durbin (approved):
Tested in a sandbox environment. Database performance looked good - there were a minimal number of slow queries logged and all under two seconds. The CPU load on the database was extremely low and only required a database capacity of 2 units.

PR #252, opened 2022-08-22
Harmony 918 - Retry long-running work items

vinnyinverso:
## Jira Issue ID
HARMONY-918

## Description
This pull request modifies the work failer to ensure that old work items are retried or failed (if retries are exhausted) when they have not been updated for some time limit. It does this by first retrieving the old work items and then calling the same function that backend service pods use to update a work item's status (which handles retries, etc.).

Question -- should we modify `FAILABLE_WORK_AGE_MINUTES` in bamboo now that we're not immediately failing the items?

## Local Test Steps
I don't know if there is a better way to do this but I tested this by setting `WORK_FAILER_PERIOD_SEC` to 120 and `FAILABLE_WORK_AGE_MINUTES` to 1, and then added `return;` at the top of `updateWorkItem` in workflow-orchestration so that I could keep work items in the `RUNNING` state and prevent them from being updated via the service pods calling back to Harmony.

I then issued a small request. Each time the work failer ran and requeued the long-running work item, I checked the workflow ui to see that the item was properly set to `READY`, and then picked up again and set to `RUNNING`. 

When the item was ready again I killed the pod that was already processing that item (from the previous try) so that it would restart a fresh/free pod to pick up the re-queued item.

I repeated this until the item exhausted all retries and failed along with the job.

I also let the work failer run when an item was in the `READY` state, and made sure it didn't try to process that.

```
2022-08-22T15:29:44.645Z [info] [work-failer]: Starting work failer
2022-08-22T15:29:44.647Z [warn] [work-failer]: Retry limit of 3 exceeded
2022-08-22T15:29:44.647Z [warn] [work-failer]: Updating work item for 649 to failed with message Work item has not been updated for over 1 minutes.
2022-08-22T15:29:44.647Z [info] [work-failer]: Work failer processed work item updates for 1 jobs and 1 work items.
```

## PR Acceptance Checklist
* [x] Acceptance criteria met


indiejames (approved):

app/workers/work-failer.ts
```diff
@@ -25,42 +27,50 @@ export default class WorkFailer implements Worker {
...
+  async processWorkItemUpdates(lastUpdateOlderThanMinutes: number): Promise<{ workItemIds: number[], jobIds: string[] }> {
+    let response: {
+      workItemIds: number[],
+      jobIds: string[]
+    } = { workItemIds: [], jobIds: [] };
+    const workItems = await getWorkItemsByUpdateAgeAndStatus(
+      db, lastUpdateOlderThanMinutes, [WorkItemStatus.RUNNING],
+      [JobStatus.RUNNING, JobStatus.RUNNING_WITH_ERRORS],
+    );
+    if (workItems.length) {
```
chris-durbin, line 43: I think this is taking advantage of 0 being falsy? I'd rewrite it as `workItems.length > 0` or `workItems?.length > 0` if it's possible `getWorkItemsByUpdateAgeAndStatus` doesn't return an empty list when there are no items though it would be best if it returned an empty list.

test/workers/work-failer.ts
```diff
@@ -2,116 +2,171 @@ import { describe } from 'mocha';
 import MockDate from 'mockdate';
```
chris-durbin, line 2: It would be nice to have an end-to-end test with the `ignore-errors` flag as well to make sure that the job status is set correctly after all retries from the work failer have failed. Right now I think we have all the functionality tested since you are reusing the same code path `processWorkItemUpdate` for these updates so it's not a must to add this test, only if there's a straightforward way to add one end to end test.

app/workers/work-failer.ts
vinnyinverso: ☑️

test/workers/work-failer.ts
vinnyinverso: ☑️

chris-durbin (approved):
We discussed and decided leaving FAILABLE_WORK_AGE_MINUTES set to the same default value of 4 hours was good and that we'd have a new ticket to add a retry button on the workflow-ui to force a work item to be picked up again so a user or admin can force an item to be picked up again if it seems to be stuck in the running state.

PR #253, opened 2022-08-22
HARMONY-1264: Fix performance bottlenecks with polling for work.

chris-durbin:
## Jira Issue ID
HARMONY-1264

## Description
While testing in my sandbox environment I found several database performance issues when issuing a single request for 230K granules.

1.) We were timing out polling for work from the database at 1 second, but the query kept running in the background.
2.) We only allowed 7 concurrent connections causing blocking for requests trying to get info from the database (polling, workflow-ui, job status, etc)
3.) The fair queueing policy queries needed to be tweaked to improve performance (I was seeing them average over 70 seconds to complete).
4.) I added an index for a query on service ID and work item status because I saw the query plans doing table scans when querying on those two fields. I do not actually have evidence that this index improved the performance, but now I'm rarely seeing queries over two seconds. The change for 3 seems to have fixed the main performance issue, but it seems like we should keep this index based on the query plan. 

## Local Test Steps
Locally you can regression test to make sure that nothing is broken. To really test you need to deploy to a sandbox environment and increase the granule limit greater than 230K and then issue a large request:
{root}/C1234724470-POCLOUD/ogc-api-coverages/1.0.0/collections/all/coverage/rangeset?subset=lat(-45%3A45)&subset=lon(0%3A180)&skipPreview=true&ignoreErrors=true

## PR Acceptance Checklist
* [X] Acceptance criteria met

### DIFF
--- a/app/models/work-item.ts
+++ b/app/models/work-item.ts
@@ -112,7 +112,7 @@ export async function getNextWorkItem(
 ): Promise<WorkItem> {
   let workItemData;
   const acceptableJobStatuses = _.cloneDeep(activeJobStatuses);
-  // Now that we use search-after instead of scrolling we could allow pausing of query-cmr
+  // TODO: Now that we use search-after instead of scrolling we could allow pausing of query-cmr
   // work items and start them back up when the job is resumed because there is no session
   if (serviceID.includes('query-cmr')) {
     acceptableJobStatuses.push(JobStatus.PAUSED);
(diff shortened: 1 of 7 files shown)


vinnyinverso (approved):

vinnyinverso:
That's a lot of fixes, great job.

app/models/work-item.ts
```diff
@@ -112,14 +112,14 @@ export async function getNextWorkItem(
 ): Promise<WorkItem> {
   let workItemData;
   const acceptableJobStatuses = _.cloneDeep(activeJobStatuses);
-  // The query-cmr service should keep going for paused jobs to avoid the ten minute CMR
-  // scroll session timeout
+  // Now that we use search-after instead of scrolling we could allow pausing of query-cmr
```
indiejames, line 115: This comment is slightly confusing since you didn't make that change. Maybe add a TODO?

indiejames (approved):

PR #255, opened 2022-09-01
Harmony 1172 - Retain original results limited message after resuming job

vinnyinverso:
## Jira Issue ID
HARMONY-1172

## Description
When a job is paused and then resumed it should now always retain the original results limited message. 

~I am not totally in love with the ultimate solution but I wanted to avoid making a major change given that our job messages and associated logic seems like it is reaching a pretty stable point.~

Refactored the job message logic to pull from a map of messages. Made some changes to job serialization as well that hopefully simplify things.

## Local Test Steps
http://localhost:3000/C1233800302-EEDTEST/ogc-api-coverages/1.0.0/collections/all/coverage/rangeset?maxResults=2

Pause and resume the job. Does the message look OK (are relevant parts, e.g. results limited messaging, retained after resumption)?

## PR Acceptance Checklist
* [x] Acceptance criteria met

### DIFF
--- a/app/middleware/cmr-granule-locator.ts
+++ b/app/middleware/cmr-granule-locator.ts
@@ -89,8 +89,16 @@ function getMaxGranules(req: HarmonyRequest, collection: string):
   return { maxGranules: maxResults, reason };
 }
 
-export const baseResultsLimitedMessage = (hits: number, maxGranules: number): string => `CMR query identified ${hits} granules, but the request has been limited `
-+ `to process only the first ${maxGranules} granules`;
+/**
+ * Constructs the base of the results limited message.
+ * @param hits - number of CMR hits
+ * @param maxGranules - limit for granule processing
+ * @returns the base of the results limited message
+ */
+export function baseResultsLimitedMessage(hits: number, maxGranules: number): string {
+  return `CMR query identified ${hits} granules, but the request has been limited `
+    + `to process only the first ${maxGranules} granules`;
+}
 
 /**
  * Create a message indicating that the results have been limited and why - if necessary
--- a/app/middleware/cmr-granule-locator.ts
+++ b/app/middleware/cmr-granule-locator.ts
@@ -82,6 +82,9 @@ function getMaxGranules(req: HarmonyRequest, collection: string): { maxGranules:
   return { maxGranules: maxResults, reason };
 }
 
+export const baseResultsLimitedMessage = (hits: number, maxGranules: number): string => `CMR query identified ${hits} granules, but the request has been limited `
++ `to process only the first ${maxGranules} granules`;
+
 /**
  * Create a message indicating that the results have been limited and why - if necessary
  * 
@@ -99,8 +102,7 @@ function getResultsLimitedMessage(req: HarmonyRequest, collection: string): stri
   const { maxGranules, reason } = getMaxGranules(req, collection);
 
   if (operation.cmrHits > maxGranules) {
-    message = `CMR query identified ${operation.cmrHits} granules, but the request has been limited `
-      + `to process only the first ${maxGranules} granules`;
+    message = baseResultsLimitedMessage(operation.cmrHits, maxGranules);
 
     switch (reason) {
       case GranuleLimitReason.MaxResults:
--- a/app/models/job.ts
+++ b/app/models/job.ts
@@ -99,26 +99,6 @@ export class SerializedJob {
 
   errors?: JobError[];
 
-  /**
-   * Returns only the links with a rel that matches the passed in value.
-   *
-   * @param rel - the relation to return links for
-   * @returns the job output links with the given rel
-   */
-  getRelatedLinks(rel: string): JobLink[] {
-    const links = this.links.filter((link) => link.rel === rel);
-    return links.map(removeEmptyProperties) as JobLink[];
-  }
-
-  /**
-   * Constructs a serialized job from an object.
-   * @param fields - the fields to initialize the SerializedJob with.
-   */
-  constructor(fields: SerializedJob) {
-    if (fields) {
-      Object.assign(this, fields);
-    }
-  }
 }
 
 export interface JobQuery {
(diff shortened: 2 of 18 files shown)


app/models/job.ts
```diff
@@ -226,6 +226,40 @@ export const statesToDefaultMessages: any = Object.values(stateMachine.states).r
...
+function removeMessageParts(
+  message: string,
+  partsToRemove,
+  partsFilter: (p: string) => boolean = undefined,
+): string {
+  if (!message) {
+    return;
+  }
+  let acceptableParts = message
+    .split('.')
```
chris-durbin, line 250: Deciding that a '.' means the end of a sentence seems problematic, but if you're only removing a portion if it's an exact match to some message it should be fine.

app/models/job.ts
```diff
@@ -600,17 +634,24 @@ export class Job extends Record implements JobRecord {
...
-  pause(message = statesToDefaultMessages.paused): void {
+  pause(): void {
     validateTransition(this.status, JobStatus.PAUSED, JobEvent.PAUSE);
-    this.updateStatus(JobStatus.PAUSED, message);
+    let newMessage = `${statesToDefaultMessages[JobStatus.PAUSED]}.`;
+    const messagePartsToRemove = activeJobStatuses.map((status) => statesToDefaultMessages[status]);
+    const retainedMessage = removeMessageParts(
+      this.message,
+      messagePartsToRemove,
+      (part) => part.includes('CMR query identified'));
```
chris-durbin, line 650: Oh we're using includes... so it is pattern matching and then the splitting on "." may be problematic if we change messages in the future.

chris-durbin (approved):
Tested it out successfully - I'm a little wary that we could change some of our messages or add new ones in the future that make the way we're parsing them break this functions, but I think the tests we have should help us catch breakages so we can worry about it later.

app/models/job.ts
```diff
@@ -226,6 +226,40 @@ export const statesToDefaultMessages: any = Object.values(stateMachine.states).r
...
+ * "The CMR identified 10 granules." when a job goes from previewing to running.
+ * @param message - the job message that may need changing
+ * @param partsToRemove - an array of strings to remove (sentences without periods--e.g. ["the job is paused"])
+ * @param partsFilter - an optional filter for removal of message parts that do not match this filter
+ * @returns the new message, with partsToRemove removed and sentence structure maintained,
+ * or an empty string if the new message has no retained parts
+ */
+function removeMessageParts(
+  message: string,
+  partsToRemove,
```
indiejames, line 243: add the type `string[]` here

app/models/job.ts
indiejames: I don't want to make this more complicated than it has to be, but does it make more sense to store the message as an array (or even a map) of messages that we can more easily pick out the parts we want to filter out (or keep)? Then we just have to concatenate the message parts whenever we want to send or log the message.

app/models/job.ts
```diff
@@ -226,6 +226,40 @@ export const statesToDefaultMessages: any = Object.values(stateMachine.states).r
...
+  partsFilter: (p: string) => boolean = undefined,
+): string {
+  if (!message) {
+    return;
```
indiejames, line 247: I think this `return` might break your contract of returning an empty string if there are no retained parts (I think it returns null or undefined).

indiejames (changes requested):
Just a couple comments regarding an empty return from a function and a missing parameter type.

vinnyinverso:
Refactored the job message logic to pull from a map of messages.

vinnyinverso:
Made some changes to job serialization as well that hopefully simplify things.

app/models/job.ts
```diff
@@ -514,11 +587,22 @@ export class Job extends Record implements JobRecord {
    */
   constructor(fields: JobRecord) {
     super(fields);
-    this.updateStatus(fields.status || JobStatus.ACCEPTED, fields.message);
+    let initialMessage: string;
+    try {
+      // newer jobs will have stringified JSON stored in the database
+      this.statesToMessages = JSON.parse(fields.message);
+      initialMessage = this.message;
+    } catch (e) {
```
chris-durbin, line 595: When using this pattern a lot of times we like to look for a specific exception type rather than catching any exception. I'm not sure if there's a specific exception to expect in this case.

chris-durbin (approved):
Verified in a sandbox environment that all of my existing jobs had the same job messages. Submitted additional jobs and verified requests with limited max results had the correct message and jobs with default messages worked exactly the same as before. Also verified the ticket is fixed where the message is retained after resuming. I noticed that pausing and canceling leave the message as the default message (removing the warning message about the number of input granules), but I think that makes sense.

app/models/services/turbo-service.ts
```diff
@@ -22,4 +23,20 @@ export default class TurboService extends BaseService<TurboServiceParams> {
   async _run(_logger: Logger): Promise<InvocationResult> {
     return null;
   }
+}
+
+/**
+ * Extends TurboService for testing purposes.
+ */
+export class TestTurboService extends TurboService {
```
indiejames, line 31: This should probably be in a file under the 'test' directory.

test/jobs/message.ts
```diff
@@ -0,0 +1,230 @@
...
+    subsetting: {
+    },
+  },
+};
+
+/**
+ * Build an operation for the tests.
+ * @returns DataOperation
+ */
+function buildOperation(message: string): DataOperation {
```
indiejames, line 30: Is this a function that might be of use in other tests? Should it be moved to a util file for that reason?

indiejames (approved):
Tested locally.

test/jobs/message.ts
vinnyinverso: ✅

app/models/services/turbo-service.ts
vinnyinverso: ✅

app/middleware/cmr-granule-locator.ts
```diff
@@ -89,6 +89,9 @@ function getMaxGranules(req: HarmonyRequest, collection: string):
   return { maxGranules: maxResults, reason };
 }
 
+export const baseResultsLimitedMessage = (hits: number, maxGranules: number): string => `CMR query identified ${hits} granules, but the request has been limited `
```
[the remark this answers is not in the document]
vinnyinverso: ✅

app/models/job.ts
```diff
@@ -74,6 +69,38 @@ export interface JobRecord {
   collectionIds: string[];
 }
 
+/**
+ * The format of a Job when returned to an end user. Serialized in
+ * the sense of displaying to an end user, not in the sense of serializing
+ * to the database.
+ */
+export class SerializedJob {
```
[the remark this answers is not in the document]
vinnyinverso: ✅

app/models/job.ts
vinnyinverso: ✅

PR #263, opened 2022-09-19
Harmony 1265 - Add status filters, retry button, retry count to work items table in workflow UI

vinnyinverso:
## Jira Issue ID
HARMONY-1265

## Description
Adds status filters, retry buttons, and a retry count column to the work items table in the workflow UI. Apologies for the largeish PR but it was much easier to develop these 3 things in tandem.

## Local Test Steps
It's probably easiest to test this with a larger request http://localhost:3000/C1233800302-EEDTEST/ogc-api-coverages/1.0.0/collections/all/coverage/rangeset. You should see a retry button you can use to trigger retries for RUNNING items. You can also filter by status and view the retry counts. Note that since processing may happen fairly quickly, you might see (in the UI) items transition directly from READY to SUCCESSFUL when triggering retries.

## PR Acceptance Checklist
* [x] Acceptance criteria met


app/views/workflow-ui/job/index.mustache.html
```diff
@@ -6,6 +6,11 @@
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
     <link rel="stylesheet" href="../../../css/eui.min.css">
     <link rel="stylesheet" href="../../../css/default.css">
+    <link rel="stylesheet" href="https://cdn.jsdelivr.net/npm/bootstrap-icons@1.9.1/font/bootstrap-icons.css">
+    <link href="https://unpkg.com/@yaireo/tagify/dist/tagify.css" rel="stylesheet" type="text/css" />
```
chris-durbin, line 10: James and I were noticing that sometimes we wait some time for the UI to load and the bottom left of the browser says 'waiting for unpkg.com' - just want to sanity check this is the best way to load the CSS.

chris-durbin (approved):
Tested out everything successfully locally. All the retry behavior worked as expected and the filters worked. The only thing I noticed was that if I paged through items with one filter set and then changed to a different filter it was not resetting the page number. For my case I paged through the ready items and then filtered for successful, and it did not show the successful items. I needed to change the query parameter in the URL back to 1 since there were only a few successful items at the time.

I think a future feature to add a 'first' and 'last' button in addition to 'previous' and 'next' would help with this case if we don't want to change the filter behavior.

app/views/workflow-ui/job/index.mustache.html
indiejames: Can we serve the CSS up via CloudFront?

indiejames (approved):

app/views/workflow-ui/job/index.mustache.html
vinnyinverso: I'm not sure about CloudFront but am open to exploring that if necessary.

PR #266, opened 2022-09-26
Harmony 1272 smart work item timeouts

indiejames:
## Jira Issue ID
HARMONY-1272

## Description
Uses a simple heuristic to retry stalled running work items that are relatively old based on previously successful work items for the same job/service.

## Local Test Steps
1. Add the following code to tasks/service-runner/app/workers/pull-worker.ts after line 170 to simulate stuck workers
```
// TODO remove this when done testing
if (Math.random() < 0.333) {
  return;
}
```
2. Build the service runner image using `npm run build` in the tasks/service-runner directory.
3. Run `bin/restart-services`
4. Run a query with multiple work items and watch the work items in the workflow-ui. You should see them fail about 1/3 of the time and get retried rather quickly (way less than two hours). The retry interval should increase as longer successful work items come in, and it should be about 1.5 times the duration of the slowest successful work item.

## PR Acceptance Checklist
* [x] Acceptance criteria met

### DIFF
--- a/app/models/work-item.ts
+++ b/app/models/work-item.ts
@@ -1,13 +1,12 @@
 import { subMinutes } from 'date-fns';
 import { ILengthAwarePagination } from 'knex-paginate';
 import _ from 'lodash';
-import { interquartileRange, max, quantile } from 'simple-statistics';
 import logger from '../util/log';
 import db, { Transaction } from '../util/db';
 import DataOperation from './data-operation';
 import { activeJobStatuses, Job, JobStatus } from './job';
 import Record from './record';
-import WorkflowStep, { getWorkflowStepByJobIdServiceId } from './workflow-steps';
+import WorkflowStep from './workflow-steps';
 import { WorkItemRecord, WorkItemStatus, getStacLocation, WorkItemQuery } from './work-item-interface';
 
 // The step index for the query-cmr task. Right now query-cmr only runs as the first step -
@@ -620,11 +619,6 @@ export async function getTotalWorkItemSizeForJobID(
   return totalSize;
 }
 
-const MIN_WORK_ITEMS_FOR_DURATION = 3;
-const MAX_WORK_ITEMS_FOR_DURATION = 10;
-const PERCENT_WORK_ITEMS_FOR_DURATION = 0.01;
-const MAX_WORK_ITEMS_WINDOW_FOR_DURATION = 100;
-
 /**
  * Compute the threshold (in milliseconds) to be used to expire work items for a given job/service
  * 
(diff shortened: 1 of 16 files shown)


tasks/service-runner/app/workers/pull-worker.ts
```diff
@@ -142,29 +142,32 @@ async function _pullAndDoWork(repeat = true): Promise<void> {
...
   try {
-    await fs.access(terminationFilePath);
-    // TERMINATING file exists so PreStop handler is requesting termination
-    logger.debug('RECEIVED TERMINATION REQUEST');
-    // removing the WORKING file is done in the `finally` block at the end of this function
-    return;
-  } catch {
-    // expected if file does not exist
-  }
+    // check to see if we are terminating
```
indiejames, line 146: Had to move this inside this try block to make sure the WORKING file gets cleaned up.

app/models/work-item.ts
```diff
@@ -602,4 +617,44 @@ export async function getTotalWorkItemSizeForJobID(
...
+  let threshold = 7200000;
+
+  try {
+    // use a simple heuristic of 2 times the longest duration of all the successful work items
+    // for this job/service
+    const result = await db(WorkItem.table)
+      .where({
+        jobID,
+        serviceID,
+        'status': 'successful',
```
vinnyinverso, line 642: Might want to use the enum here.

app/models/work-item.ts
```diff
@@ -602,4 +617,44 @@ export async function getTotalWorkItemSizeForJobID(
...
+    } else {
+      logger.debug('Using default threshold');
+    }
+    logger.debug(`Threshold is ${threshold}`);
+
+  } catch (e) {
+    logger.error(`Failed to get MAX duration for service ${serviceID} of job ${jobID}`);
+  }
+
+  return threshold;
```
vinnyinverso, line 658: Do we need any sort of sanity check that this is a positive number since the default value for the duration column is negative?

vinnyinverso:
This is really great overall. I have two things that I'm curious about.

Does the `threshold = 2.0 * result.max` work well for items that complete with smaller durations?

- Suppose most items early on in the step are taking ~1s, but later items in the dataset are taking ~2s (e.g. because their files are larger or some other service-specific detail causes them to process for a bit longer, or Harmony is not processing their updates as fast due to heavy system load). Would there be enough buffer to allow those later items to complete without continually retrying (before eventually failing)?

- What would happen during a deployment? It seems like a large number of items could quickly be failed if a few items succeeded fairly quickly prior to the deployment.

app/models/work-item.ts
indiejames: good catch. done.

app/models/work-item.ts
indiejames: Done.

indiejames:
> This is really great overall. I have two things that I'm curious about.
> 
> Does the `threshold = 2.0 * result.max` work well for items that complete with smaller durations?
> 
> * Suppose the first item  in the step takes ~1s, but later items are taking ~3s (e.g. because their files are larger or some other service-specific detail causes them to process for a bit longer, or Harmony is not processing their updates as fast due to heavy system load). Would there be enough buffer to allow those later items to complete without continually retrying (before eventually failing)?

Yes, there is at least a one minute buffer between retries and also if a granule gets retried then the original worker responds the new threshold will be increased according to the duration the original worker reports.

> * What would happen during a deployment? It seems like a large number of items could quickly be failed if a few items succeeded fairly quickly prior to the deployment.
Items will only fail in that case if the pods are getting evicted, in which case the items _should_ be retried.

chris-durbin (approved):
Deployed this code to sandbox and tested. I ran a 110K granule request and continually re-ran deployments while the request was going - I deployed at least 10 times throughout the run. In the end everything succeeded except the 4 S3 granules that always fail. I watched using the workflow-ui and the retries worked as expected with no work item left in a running state for more than 10 minutes throughout the run.

vinnyinverso (approved):

PR #277, opened 2022-10-17
Harmony 1304 - Hide retry and logs columns when functionality is inaccessible due to permissions

vinnyinverso:
## Jira Issue ID
HARMONY-1304

## Description
This ensures that when a user has inadequate permissions to view the logs or retry button(s) on the workflow UI, the corresponding columns will be hidden. Previously we were just hiding the buttons, but not also the columns. This was confusing, e.g. for a user who will never see the logs button but does see a column header named "logs".

## Local Test Steps

Submit a job (as an admin or non-admin, http://localhost:3000/C1233800302-EEDTEST/ogc-api-coverages/1.0.0/collections/all/coverage/rangeset?maxResults=2), then login to harmony with a different non-admin user and take a look at the other user's job. Make sure that there is no "logs" and no "retry" column.

## PR Acceptance Checklist
* [x] Acceptance criteria met

### DIFF
--- a/app/frontends/workflow-ui.ts
+++ b/app/frontends/workflow-ui.ts
@@ -275,10 +275,9 @@ function workItemRenderingFunctions(job: Job, isAdmin: boolean, requestUser: str
         ' title="view logs"><i class="bi bi-body-text"></i></a>';
     },
     workflowItemRetryButton(): string {
-      const sharedWithNonAdmin = (!isAdmin && (job.username != requestUser));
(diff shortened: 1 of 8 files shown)


indiejames (approved):

vinnyinverso:
> at getJobIfAllowed (/Users/cddurbin/workspace/harmony/harmony/app/util/job.ts:265:11)

Fixed -- thanks for catching that.

PR #278, opened 2022-10-18
Harmony 1257 - Append work item logs to logs.json when items are retried

vinnyinverso:
## Jira Issue ID
Harmony 1257

## Description
When we first implemented the ability to see logs from the workflow ui there was no concept of re-tries, so there was an implicit assumption that a log file would only be written once. Now that that's no longer the case, I have updated the code to append new logs if there is already an existing log file.

## Local Test Steps
In service-runner.ts change `if (status.status === 'Success') {` to `if (status.status === '') {`. Rebuild service runner and restart your services, then submit a request http://localhost:3000/C1233800302-EEDTEST/ogc-api-coverages/1.0.0/collections/all/coverage/rangeset?maxResults=2. When the work item fails, check the logs file, which should include logs for each invocation of the service (with the same work item).

## PR Acceptance Checklist
* [x] Acceptance criteria met

### DIFF
--- a/app/frontends/workflow-ui.ts
+++ b/app/frontends/workflow-ui.ts
@@ -10,7 +10,7 @@ import db from '../util/db';
 import version from '../util/version';
 import env = require('../util/env');
 import { keysToLowerCase } from '../util/object';
-import { COMPLETED_WORK_ITEM_STATUSES, getItemLogsLocation, WorkItemQuery, WorkItemStatus } from '../models/work-item-interface';
+import { getItemLogsLocation, WorkItemQuery, WorkItemStatus } from '../models/work-item-interface';
 import { getRequestRoot } from '../util/url';
 import { belongsToGroup } from '../util/cmr';
 import { getAllStateChangeLinks, getJobStateChangeLinks } from '../util/links';
@@ -269,7 +269,7 @@ function workItemRenderingFunctions(job: Job, isAdmin: boolean, requestUser: str
     workflowItemCreatedAt(): string { return this.createdAt.getTime(); },
     workflowItemUpdatedAt(): string { return this.updatedAt.getTime(); },
     workflowItemLogsButton(): string {
-      const isComplete = COMPLETED_WORK_ITEM_STATUSES.indexOf(this.status) > -1;
+      const isComplete = [WorkItemStatus.FAILED, WorkItemStatus.SUCCESSFUL].indexOf(this.status) > -1;
       if (!isComplete || !isAdmin || this.serviceID.includes('query-cmr')) return '';
       const logsUrl = `/admin/workflow-ui/${job.jobID}/${this.id}/logs`;
       return `<a type="button" target="__blank" class="btn btn-light btn-sm logs-button" href="${logsUrl}"` +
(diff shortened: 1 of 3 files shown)


app/frontends/workflow-ui.ts
```diff
@@ -269,7 +269,7 @@ function workItemRenderingFunctions(job: Job, isAdmin: boolean, requestUser: str
     workflowItemCreatedAt(): string { return this.createdAt.getTime(); },
     workflowItemUpdatedAt(): string { return this.updatedAt.getTime(); },
     workflowItemLogsButton(): string {
-      const isComplete = COMPLETED_WORK_ITEM_STATUSES.indexOf(this.status) > -1;
+      const isComplete = [WorkItemStatus.FAILED, WorkItemStatus.SUCCESSFUL].indexOf(this.status) > -1;
```
chris-durbin, line 272: Why did you need to make this change? I assume the other status is 'canceled'?

tasks/service-runner/app/service/service-runner.ts
```diff
@@ -174,6 +175,22 @@ export async function runQueryCmrFromPull(workItem: WorkItemRecord, maxCmrGranul
...
+ * @param workItem - the work item that the logs are for
+ * @param logs - logs array from the k8s exec call
+ */
+export async function uploadLogs(workItem: WorkItemRecord, logs: (string | object)[]): Promise<ManagedUpload.SendData> {
+  let fileContent: (string | object)[] = logs;
+  const s3 = objectStoreForProtocol('s3');
+  const logsLocation = getItemLogsLocation(workItem);
+  if (await s3.objectExists(logsLocation)) { // append to existing logs
+    const logsArray = await s3.getObjectJson(logsLocation);
+    fileContent = logsArray.concat(fileContent);
```
chris-durbin, line 189: What do you think about adding a blank line and then 'Start of retry' or something along those lines to separate the messages between retries?

chris-durbin (approved):

app/frontends/workflow-ui.ts
vinnyinverso: Yes I'm assuming if it's canceled, the majority of the time we won't have logs to look at, nor will we care to look at them.

vinnyinverso: The tricky thing is that there is no efficient way to check whether the logs button should be shown other than by looking at the status of the work item. The alternative would be to show the logs button for all of the canceled work items, and just return an error response for those that are missing.

tasks/service-runner/app/service/service-runner.ts
vinnyinverso: Yeah, that sounds like a good idea. I will add that.

indiejames (approved):

tasks/service-runner/app/service/service-runner.ts
chris-durbin: I just realized with my suggestion you might want to add the line as JSON if the logs are formatted as JSON so that the formatting in a web browser still looks right.

vinnyinverso: No luck with the newline, but I can definitely ensure the format matches whatever is being logged. The message I have so far is:
> 'Start of service execution (retryCount=0, id=0)'

PR #287, opened 2022-11-09
HARMONY-1317: Ensure 2 items complete before calculating smart timeout threshold

vinnyinverso:
## Jira Issue ID
HARMONY-1317

## Description
For work item timeouts, this change ensures that at least 2 items complete before calculating the smart timeout threshold, otherwise it will return the default timeout threshold.

## Local Test Steps
Follow steps from https://github.com/nasa/harmony/pull/266 and also log the `computeWorkItemDurationOutlierThresholdForJobService` logic to see if the computation is working as expected.

## PR Acceptance Checklist
* [x] Acceptance criteria met

### DIFF
--- a/app/workers/work-failer.ts
+++ b/app/workers/work-failer.ts
@@ -51,15 +51,15 @@ export default class WorkFailer implements Worker {
     const workItems = await getWorkItemsByUpdateAgeAndStatus(
       db, lastUpdateOlderThanMinutes, [WorkItemStatus.RUNNING],
       [JobStatus.RUNNING, JobStatus.RUNNING_WITH_ERRORS],
-      ['w.id', 'w.jobID', 'serviceID', 'startedAt'],
+      ['w.id', 'w.jobID', 'serviceID', 'startedAt', 'workflowStepIndex'],
     );
     if (workItems?.length > 0) {
       // compute duration thresholds for each job/service
       for (const workItem of workItems) {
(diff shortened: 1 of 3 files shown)


indiejames (approved):

app/workers/work-failer.ts
```diff
@@ -51,15 +51,15 @@ export default class WorkFailer implements Worker {
     const workItems = await getWorkItemsByUpdateAgeAndStatus(
       db, lastUpdateOlderThanMinutes, [WorkItemStatus.RUNNING],
       [JobStatus.RUNNING, JobStatus.RUNNING_WITH_ERRORS],
-      ['w.id', 'w.jobID', 'serviceID', 'startedAt'],
+      ['w.id', 'w.jobID', 'serviceID', 'startedAt', 'workflowStepIndex'],
```
chris-durbin, line 54: I'm curious why you needed the workflow step index - in case the same service is used multiple times in the chain? I think the change makes sense for that case, but wondering if there was another reason.

chris-durbin (approved):
Nit: update the JIRA ticket number to HARMONY-1317 in the PR overview.

app/workers/work-failer.ts
vinnyinverso: There's a function within `getWorkItemsByUpdateAgeAndStatus` that I'm calling in order to get work item counts that requires it.

PR #288, opened 2022-11-09
Harmony 1279 predictable order

indiejames:
## Jira Issue ID
HARMONY-1279

## Description
Fixes a bug in preserving order in which granules are batched independent of when work items finish.

## Local Test Steps
Set MAX_BATCH_INPUTS=3 in .env and execute a `concise` query several times. Verify that the generated outputs are consistent.

## PR Acceptance Checklist
* [x] Acceptance criteria met


chris-durbin (approved):
Sanity checked aggregation requests still worked for me. I think most of those fixtures are duplicates (happens to me when I run with `mocha -w` and the fixtures do not exist when I  first start mocha). You should probably delete all the fixtures added with this PR and then run the tests again to just generate the ones needed.

chris-durbin:
I also tested in sandbox with 10 items per batch and watched as a few of the early items in the request stayed running while the later ones completed. No new batches were created while the early items were still running. As the early ones started getting retried and finishing, batches would get created at the expected time and my 407 granule request ended up completing successfully.

indiejames:
> 

I delete them, but they reappeared when I ran `npm run test`.

PR #290, opened 2022-11-10
HARMONY-1274: Add tests that batching obeys services.yml configuration

indiejames:
## Jira Issue ID
HARMONY-1274

## Description
Adds tests to verify that batching for aggregation services obeys configuration set in services.yml

## Local Test Steps
* add `max_batch_inputs: 3` after line 613 in services.yml
* restart harmony
* Issue the following request
```
http://localhost:3000/C1234208438-POCLOUD/ogc-api-coverages/1.0.0/collections/all/coverage/rangeset?skipPreview=true&maxResults=7&concatenate=true
```
* Verify using the workflow-ui that the job has three concise work items
* Use the following SQL query to verify that there were seven batch_items created in three batches
```sql
select * from batch_items;
```
## PR Acceptance Checklist
* [x] Acceptance criteria met

### DIFF
--- a/test/aggregation-batching.ts
+++ b/test/aggregation-batching.ts
@@ -15,6 +15,7 @@ import { Job } from '../app/models/job';
 import WorkItem from '../app/models/work-item';
 import { objectStoreForProtocol } from '../app/util/object-store';
 import { truncateAll } from './helpers/db';
+import { hookServices } from './helpers/stub-service';
 
 /**
  * Create a work item update for a query-cmr get work response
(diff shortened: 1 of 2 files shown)


test/aggregation-batching.ts
```diff
@@ -54,6 +55,15 @@ describe('when testing a batched aggregation service', function () {
   hookServersStartStop({ skipEarthdataLogin: false });
   const collection = 'C1243729749-EEDTEST';
   describe('with only one batch that should be created', function () {
+    let batchInputsStub;
```
chris-durbin, line 58: It doesn't look like this part is needed (the test was passing before).

chris-durbin (approved):

test/aggregation-batching.ts
indiejames: I added this just to be explicit about what values were getting set.

vinnyinverso (approved):

PR #293, opened 2022-11-15
HARMONY-1280: As a harmony user, I want my aggregation request to be able to complete even if some items fail

chris-durbin:
## Jira Issue ID
HARMONY-1280

## Description
Updates batched aggregrations to be able to run successfully even if some items in prior steps fail or if some of the batched aggregations themselves fail when the user has requested `ignoreErrors=true`.

I've tested in the sandbox quite a bit with large requests with several failing items and verifying in the end the batched aggregations are created correctly and complete successfully.

## Local Test Steps
Submit the following request which has one successful and two failing granules:
http://localhost:3000/C1234208438-POCLOUD/ogc-api-coverages/1.0.0/collections/bathymetry/coverage/rangeset?concatenate=true&subset=lon(-160%3A160)&subset=lat(-80%3A80)&maxResults=3&skipPreview=true&ignoreErrors=true&granuleId=G1234495188-POCLOUD,G1234515613-POCLOUD,G1234515574-POCLOUD,G1234516982-POCLOUD,G1234518216-POCLOUD,G1234518228-POCLOUD,G1236679413-POCLOUD

Verify that the request eventually completes with errors and concise runs to produce the final output with just one granule.

## PR Acceptance Checklist
* [X] Acceptance criteria met

### DIFF
--- a/app/backends/workflow-orchestration.ts
+++ b/app/backends/workflow-orchestration.ts
@@ -269,6 +269,8 @@ async function createAggregatingWorkItem(
  * @param allWorkItemsForStepComplete - true if all the work items for the current step are complete
  * @param results - an array of paths to STAC catalogs
  * @param outputItemSizes - an array of sizes (in bytes) of the output items for the current step
+ *
+ * @returns true if it created a work item
  */
 async function createNextWorkItems(
   tx: Transaction,
@@ -279,8 +281,10 @@ async function createNextWorkItems(
   allWorkItemsForStepComplete: boolean,
   results: string[],
   outputItemSizes: number[],
-): Promise<void> {
-  if (results && results.length > 0) {
+): Promise<boolean> {
+  let createdWorkItem = false;
+  if (results && results.length > 0 || nextWorkflowStep.isBatched) {
+    createdWorkItem = true;
     // if we have completed all the work items for this step or if the next step does not
     // aggregate then create a work item for the next step
     if (nextWorkflowStep.hasAggregatedOutput) {
@@ -296,7 +300,7 @@ async function createNextWorkItems(
         }
         // TODO add other services that can produce more than one output and so should have their
         // batching sortIndex propagated to child work items to provide consistent batching
-        await handleBatching(
+        createdWorkItem = await handleBatching(
           tx,
           logger,
           nextWorkflowStep,
@@ -336,6 +340,7 @@ async function createNextWorkItems(
       }
     }
   }
+  return createdWorkItem;
 }
 
 /**
@@ -475,6 +480,7 @@ async function handleFailedWorkItems(
         const errorCount =  await getErrorCountForJob(tx, job.jobID);
         if (errorCount > env.maxErrorsForJob) {
           jobMessage = `Maximum allowed errors ${env.maxErrorsForJob} exceeded`;
+          logger.warn(jobMessage);
           continueProcessing = false;
         }
       }
@@ -533,9 +539,14 @@ export async function handleWorkItemUpdate(
   logger: Logger): Promise<void> {
   const { workItemID, hits, results, scrollID } = update;
   let { errorMessage, status } = update;
+  let createdWorkItem = false;
   if (status === WorkItemStatus.SUCCESSFUL) {
     logger.info(`Updating work item ${workItemID} to ${status}`);
   }
+
+  // if (workItemID == 5) {
+  //   console.log('CDD - this is it');
+  // }
   // get the jobID for the work item
   const jobID = await getJobIdForWorkItem(workItemID);
 
@@ -554,6 +565,9 @@ export async function handleWorkItemUpdate(
 
   try {
     await db.transaction(async (tx) => {
+      if (workItemID == 5) {
+        console.log('CDD - this is it');
+      }
       const job = await Job.byJobID(tx, jobID, false, true);
       // lock the work item to we can update it - need to do this after locking jobs table above
       // to avoid deadlocks
@@ -634,7 +648,7 @@ export async function handleWorkItemUpdate(
           tx, workItem.jobID, workItem.workflowStepIndex + 1,
         );
         if (nextWorkflowStep && (status !== WorkItemStatus.FAILED || nextWorkflowStep?.isBatched)) {
-          await createNextWorkItems(
+          createdWorkItem = await createNextWorkItems(
             tx,
             nextWorkflowStep,
             logger,
@@ -665,10 +679,15 @@ export async function handleWorkItemUpdate(
           }
           // If all granules are finished mark the job as finished
           job.completeBatch(thisStep.workItemCount);
-          if (allWorkItemsForStepComplete) {
+          // if (allWorkItemsForStepComplete) {
+          console.log(`CDD: Created work item is ${createdWorkItem}`);
+          if (allWorkItemsForStepComplete && !createdWorkItem && (!nextWorkflowStep || nextWorkflowStep.workItemCount === 0)) {
+          // if (allWorkItemsForStepComplete && !createdWorkItem && !nextWorkflowStep) {
             const finalStatus = await getFinalStatusForJob(tx, job);
             await completeJob(tx, job, finalStatus, logger);
           } else {
+            console.log('CDD unexpected 1');
+            // Either previewing or next step is a batched step and this item failed
             // Special case to pause the job as soon as any single granule completes when in the previewing state
             if (job.status === JobStatus.PREVIEWING) {
               job.pause();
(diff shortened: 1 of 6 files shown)


app/backends/workflow-orchestration.ts
```diff
@@ -526,138 +537,162 @@ export async function handleWorkItemUpdate(
   update: WorkItemUpdate,
   operation: object,
   logger: Logger): Promise<void> {
-  const { workItemID, status, hits, results, scrollID, errorMessage } = update;
+  const { workItemID, hits, results, scrollID } = update;
+  let { errorMessage, status } = update;
+  let createdWorkItem = false;
```
indiejames, line 542: Can you change this do `didCreateWorkItem` or something similar that indicates it is a boolean? I was confused where it was assigned and was thinking it was a `WorkItem` that got created.

app/backends/workflow-orchestration.ts
```diff
@@ -526,138 +537,162 @@ export async function handleWorkItemUpdate(
...
+            await maybeQueueQueryCmrWorkItem(tx, workItem, logger);
+          } else {
+            // Failed to create the next work items when there should be work items.
+            // Fail the job rather than leaving it orphaned in the running state
+            logger.error('The work item update should have contained results to queue a next work item, but it did not.');
+            const message = 'Harmony internal failure: could not create the next work items for the request.';
+            await completeJob(tx, job, JobStatus.FAILED, logger, message);
+          }
+        } else if (!nextWorkflowStep || allWorkItemsForStepComplete) {
+          // Finished with the chain for this granule
```
indiejames, line 670: I don't think this comment is true if `nextWorkflowStep` is `true`, which it might be.

app/backends/workflow-orchestration.ts
```diff
@@ -526,138 +537,162 @@ export async function handleWorkItemUpdate(
...
+          } else {
+            // Either previewing or next step is a batched step and this item failed
+            if (job.status === JobStatus.PREVIEWING) {
+              // Special case to pause the job as soon as any single granule completes when in the previewing state
+              job.pause();
+            }
+            await job.save(tx);
           }
+        } else { // Currently only reach this condition for batched aggregation requests
           await job.save(tx);
```
indiejames, line 688: We are calling `job.save(tx)` from several places. I wonder if we should pull it of of the `if/else` blocks and do it afterwards. We could rewrite `completeJob` to not `save` the job to avoid the redundant save.

app/backends/workflow-orchestration.ts
```diff
@@ -526,138 +537,162 @@ export async function handleWorkItemUpdate(
...
-        job.completeBatch(thisStep.workItemCount);
-        if (allWorkItemsForStepComplete) {
-          const finalStatus = await getFinalStatusForJob(tx, job);
-          await completeJob(tx, job, finalStatus, logger);
-        } else {
-          // Special case to pause the job as soon as any single granule completes when in the previewing state
-          if (job.status === JobStatus.PREVIEWING) {
-            job.pause();
+
+        if (nextWorkflowStep && status === WorkItemStatus.SUCCESSFUL) {
```
indiejames, line 657: the logic in this section is getting kind of confusing for me

indiejames (changes requested):
A few initial comments. I'll continue with this in a bit.

app/backends/workflow-orchestration.ts
```diff
@@ -526,138 +537,162 @@ export async function handleWorkItemUpdate(
...
-    }
+  try {
+    await db.transaction(async (tx) => {
+      const job = await Job.byJobID(tx, jobID, false, true);
+      // lock the work item to we can update it - need to do this after locking jobs table above
+      // to avoid deadlocks
+      const workItem = await getWorkItemById(tx, workItemID, true);
+      const thisStep = await getWorkflowStepByJobIdStepIndex(tx, workItem.jobID, workItem.workflowStepIndex);
+
+      // If the job was already in a terminal state then send 409 response
```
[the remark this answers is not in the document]
chris-durbin: Done.

app/backends/workflow-orchestration.ts
```diff
@@ -526,138 +537,162 @@ export async function handleWorkItemUpdate(
...
+        duration,
+        totalItemsSize,
+        outputItemSizes);
+
+      const completedWorkItemCount = await workItemCountForStep(
+        tx, workItem.jobID, workItem.workflowStepIndex, COMPLETED_WORK_ITEM_STATUSES,
+      );
+      const allWorkItemsForStepComplete = (completedWorkItemCount == thisStep.workItemCount);
+
+      if (hits && job.numInputGranules > hits) {
```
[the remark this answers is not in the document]
chris-durbin: Done.

app/backends/workflow-orchestration.ts
chris-durbin: Done.

app/backends/workflow-orchestration.ts
```diff
@@ -526,138 +537,162 @@ export async function handleWorkItemUpdate(
...
-      } else {
-        // Finished with the chain for this granule
-        if (status != WorkItemStatus.FAILED) {
-          await addJobLinksForFinishedWorkItem(tx, job, results, logger);
+      const continueProcessing = await handleFailedWorkItems(tx, job, workItem, thisStep, status, logger, errorMessage);
+      if (continueProcessing) {
+        const nextWorkflowStep = await getWorkflowStepByJobIdStepIndex(
+          tx, workItem.jobID, workItem.workflowStepIndex + 1,
+        );
+        if (nextWorkflowStep && (status !== WorkItemStatus.FAILED || nextWorkflowStep?.isBatched)) {
```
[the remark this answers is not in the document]
chris-durbin: The ignoreErrors logic is handled entirely within `handleFailedWorkItems` (one of the few places we've managed to do that in workflow orchestration). It will not enter this block if `ignoreErrors` is false and the status is failed because `continueProcessing` will be false if there's an error.

app/backends/workflow-orchestration.ts
```diff
@@ -526,138 +537,162 @@ export async function handleWorkItemUpdate(
...
+          createdWorkItem = await createNextWorkItems(
+            tx,
+            nextWorkflowStep,
+            logger,
+            workItem,
+            status,
```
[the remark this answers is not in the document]
chris-durbin: Done

app/backends/workflow-orchestration.ts
chris-durbin: A successful work item would have entered the prior if block when `nextWorkflowStep` is `true`, so the only way to be in this block having  `nextWorkflowStep=true` would be if the item failed, in which case we are done processing that granule, so the comment is correct.

app/backends/workflow-orchestration.ts
chris-durbin: Yeah, this whole file and `aggregation-batching`. At least we have significant coverage in the tests to capture many of the possible permutations. It should also allow us to refactor things in the future. I considered trying to rewrite to capture all the possible permutations upfront rather than having deeply nested checks, but that would have had a lot of duplicated code. I think the requirements and logic are really complex at this point so it's going to be hard to not have complex branching code.

app/util/aggregation-batch.ts
```diff
@@ -182,28 +181,40 @@ async function createStacCatalogForBatch(
...
   const { jobID, serviceID, stepIndex } = workflowStep;
   const batchItemUrls = await getItemUrlsForJobServiceBatch(tx, jobID, serviceID, batch.batchID);
-  // create STAC catalog for the batch
-  const catalogUrl = await createStacCatalogForBatch(
-    jobID,
-    stepIndex,
-    batch.batchID,
-    workflowStep.collectionsForOperation()[0],
-    batchItemUrls);
+  if (batchItemUrls && batchItemUrls.length > 0) {
```
indiejames, line 193: Not sure if it is actually better, but I think you can use
```
if (batchItemUrls?.length > 0) {
```
here as well

chris-durbin: I started with that that but didn't like it because I'm doing the comparison with 0, but yes the code did work.

indiejames (changes requested):
Just a few small comments now.

test/ignore-errors.ts
```diff
@@ -135,6 +135,7 @@ describe('when setting ignoreErrors=true', function () {
           firstSwotItem = JSON.parse(res.text).workItem;
           firstSwotItem.status = WorkItemStatus.FAILED;
           firstSwotItem.results = [];
+          firstSwotItem.results = [];
```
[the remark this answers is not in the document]
chris-durbin: Done.

app/backends/workflow-orchestration.ts
indiejames: OK, 'finished with the chain' implies success to me, but maybe not

indiejames (approved):

vinnyinverso (approved):

PR #303, opened 2022-11-25
Harmony 1323, 1324 - Add kibana pod logs links to workflow ui work items page

vinnyinverso:
## Jira Issue ID
HARMONY-1323, HARMONY-1324

## Description
Adds a column to the work items table of the workflow UI for accessing pod logs in Kibana. It will show an individual link corresponding to each run/try/retry (0, 1, 2, etc).
Screen Shot 2022-11-25 at 4 17 50 PM

## Local Test Steps
Rebuild service-runner.

Paste the below variable in your .env file: 

`METRICS_ENDPOINT='<METRICS BASE>?_g=(filters:!(),refreshInterval:(pause:!t,value:0),time:(from:'{{from}}',to:'{{to}}'))&_a=(columns:!(),filters:!(),index:<INDEX>,interval:auto,query:(language:kuery,query:'{{query}}'),sort:!(!('@timestamp',desc)))'`

We won't actually be able to successfully see any logs since our local logs won't go to the metrics logs dashboards, but it will be helpful to see how `{{to}}`, `{{from}}`, and `{{query}}` are replaced in this string and, when the link is clicked via the UI, to see how Kibana uses these values in the dashboard.

Browse to the actual SIT metrics logs dashboard first to see what the actual base URL and index values look like.

Replace `<METRICS BASE>` with the actual base metrics URL (e.g. `https://base-url/s/space-identifier/app/discover#/`) for SIT. Replace `<INDEX>` with the actual index ID. 

Run a request and periodically kill the worker container to simulate retries so that you can see multiple pod logs links. Follow the links and see if the filters are applied in Kibana as you would expect. Note that we use GMT timezone in the URL, but you may see some times in EST (subtracts 5 hours from GMT) on your dashboard.

## PR Acceptance Checklist
* [x] Acceptance criteria met

### DIFF
--- a/app/frontends/workflow-ui.ts
+++ b/app/frontends/workflow-ui.ts
@@ -264,7 +264,11 @@ function workItemRenderingFunctions(job: Job, isAdmin: boolean, requestUser: str
   badgeClasses[WorkItemStatus.SUCCESSFUL] = 'success';
   badgeClasses[WorkItemStatus.RUNNING] = 'info';
   return {
-    workflowItemRunnerIds(): string { return (this.runnerIds.join(', ')); },
+    workflowItemRunnerIds(): string { 
+      return this.runnerIds.map((runnerId: string) => {
+        return `<li><a class="dropdown-item" href="#">${runnerId}</a></li>`;
+      }).join(''); 
+    },
     workflowItemBadge(): string { return badgeClasses[this.status]; },
     workflowItemStep(): string { return sanitizeImage(this.serviceID); },
     workflowItemCreatedAt(): string { return this.createdAt.getTime(); },
--- a/app/frontends/workflow-ui.ts
+++ b/app/frontends/workflow-ui.ts
@@ -266,7 +266,7 @@ function workItemRenderingFunctions(job: Job, isAdmin: boolean, requestUser: str
   return {
     workflowItemRunners(): string { 
       return this.runners.map((runner: { id: string, startedAt: number }) => {
-        return `<li><a class="dropdown-item" href="#">${runner.id} ${runner.startedAt}</a></li>`;
+        return `<li><a class="dropdown-item" href="#">${runner.id} ${(new Date(runner.startedAt)).toISOString()}</a></li>`;
       }).join(''); 
     },
     workflowItemBadge(): string { return badgeClasses[this.status]; },
--- a/app/frontends/workflow-ui.ts
+++ b/app/frontends/workflow-ui.ts
@@ -271,8 +271,8 @@ function workItemRenderingFunctions(job: Job, isAdmin: boolean, requestUser: str
     workflowItemPodLogs(): string {
       if (!this.runners || this.runners.length < 1) return '';
       return this.runners.map((runner: { id: string, startedAt: number }, i: number) => {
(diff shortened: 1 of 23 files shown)


env-defaults
```diff
@@ -60,6 +60,10 @@ CALLBACK_URL_ROOT=http://harmony:3001
 # The CMR Endpoint to use (e.g. URL for local, SIT, UAT, or production)
 CMR_ENDPOINT=https://cmr.uat.earthdata.nasa.gov
 
+# The Metrics Endpoint to use (e.g. URL for SIT, UAT, or production)
+# Should contain placeholders like {{query}}, {{from}}, and {{to}}, which are replaced at runtime.
+METRICS_ENDPOINT=http://localhost:3000?query={{query}}&from={{from}}&to={{to}}
```
chris-durbin, line 65: I think the endpoint should be just a URL - any type of construction should ideally happen in the code. It looks like you might need one more configuration item which is the name of the index to query?

tasks/service-runner/config/query-cmr-sidecar.yaml
```diff
@@ -76,6 +76,10 @@ spec:
             valueFrom:
               fieldRef:
                 fieldPath: metadata.namespace
+          - name: MY_POD_UID
```
chris-durbin, line 79: It looks like the sidecar and service-template files are only used locally. We'll need a change to the ci-cd repo as well to add `MY_POD_UID`.

app/frontends/workflow-ui.ts
```diff
@@ -268,6 +268,17 @@ function workItemRenderingFunctions(job: Job, isAdmin: boolean, requestUser: str
   badgeClasses[WorkItemStatus.SUCCESSFUL] = 'success';
   badgeClasses[WorkItemStatus.RUNNING] = 'info';
   return {
+    workflowItemPodLogs(): string {
+      if (!this.runners || this.runners.length < 1) return '';
+      return this.runners.map((runner: { id: string, startedAt: number }, i: number) => {
+        const to = this.status === WorkItemStatus.RUNNING ? 'now' : this.updatedAt.toISOString();
+        const url = env.metricsEndpoint
```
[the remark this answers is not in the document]
chris-durbin: Note that using the default METRICS_ENDPOINT I am seeing:
`query=kubernetes.pod_id%3A "2e6641e3-9e27-4be3-9eb3-a9b9d8034a6a"`

http://localhost:3000/?query=kubernetes.pod_id%3A%20%222e6641e3-9e27-4be3-9eb3-a9b9d8034a6a%22&from=2022-11-28T11:04:44.850Z&to=2022-11-28T11:05:01.225Z

chris-durbin (changes requested):
I tested out locally and in sandbox after rebuilding and pushing images. I think we'll need a ci-cd PR to set MY_POD_UID because I am seeing the pod set to `undefined` in sandbox. I noticed we have multiple links - one for each retry. I was thinking maybe we could just have a single link which added each of the pods as it's own filter to show all the logs at once. We could also set the text for the link to 'Metrics' or something instead of 0.

vinnyinverso:
I addressed the comments and tested in sandbox.

chris-durbin (approved):

indiejames (approved):

PR #305, opened 2022-11-29
HARMONY-1309: Fix issue with skip-preview and resuming job requests

chris-durbin:
## Jira Issue ID
HARMONY-1309

## Description
JSON schema validation errors were occurring for request to skip-preview or resume requests in certain cases. The underlying cause was the subset field was being removed from the operation when constructing workflow specific operations, but the schema requires the subset field. So this makes sure `subset: {}` will always be present.

## Local Test Steps
Test with this request from the ticket:

http://localhost:3000/C1245618475-EEDTEST/ogc-api-coverages/1.0.0/collections/all/coverage/rangeset?forceAsync=true&maxResults=500&concatenate=true&format=application%2Fx-zarr

Click on the skip-preview link from the browser on the job status page and verify the state is changed to running. Then click the pause link and verify the request is paused. Then click the resume link and verify the request is running.

Note that the reason skip-preview and resume worked ok from the workflow-ui for some of us was because we are admins and the code that causes the operation to be updated with the token only happens when a username is included and the workflow-ui path removes the username if you are an admin.

## PR Acceptance Checklist
* [X] Acceptance criteria met


test/jobs/skip-job-preview.ts
```diff
@@ -216,8 +220,32 @@ describe('Skipping job preview', function () {
...
+          hookRangesetRequest('1.0.0', 'C1245618475-EEDTEST', 'all', { query: { maxResults: 500, format: 'application/x-zarr' }, username: 'joe' });
+          hookRedirect('joe');
 
-        describe('When a job is previewing', function () {
+          it('puts the job in the previewing state', function () {
+            const job = JSON.parse(this.res.text);
+            expect(job.status).to.eql('previewing');
+          });
+
+          describe('when skipping the preview it sets the job to the running status', async function () {
```
chris-durbin, line 232: I tried many different things, but the only way I could get access to the job ID to use to call skipPreview was from within a before block, so that's why the test setup describe/before/it looks so clunky.

app/util/job.ts
```diff
@@ -194,11 +191,36 @@ export async function resumeAndSaveJob(
         await workflowStep.save(tx);
       }
     }
-    job.resume();
+    if (jobFnName === 'resume') {
+      job.resume();
+    } else if (jobFnName === 'skipPreview') {
+      job.skipPreview();
+    }
```
indiejames, line 198: Throw an error here if `jobFnName` is not 'resume' or 'skipPreview'. Probably should do that first thing in this function. Unless there is a valid use case to call this function without either of those.

app/util/job.ts
```diff
@@ -161,23 +161,20 @@ export async function pauseAndSaveJob(
 }
 
 /**
- * Resume a paused job then save it.
+ * Updates the user access token in the database and applies the appropriate
+ * job function - either resume or skipPreview.
```
indiejames, line 165: Is there a valid use case for calling this function with neither 'resume' nor 'skipPreview'? If not, I think doing dispatch using a string argument is not great. Better to have two separate functions for each functionality that call a common 'updateAccessToken' function to do the common work.

app/util/job.ts
```diff
@@ -194,11 +191,36 @@ export async function resumeAndSaveJob(
...
+ * belong to the user.
+ */
+export async function resumeAndSaveJob(
+  jobID: string,
+  _logger: Logger,
+  username?: string,
+  token?: string,
+
+): Promise<void> {
+  await updateTokenAndChangeState(jobID, username, token, 'resume');
```
indiejames, line 221: Like I said earlier, I think it would be better if this function just called `updatedToken` without the change state part using a shared transaction and then did the 'resume' part and job saving here.

app/util/job.ts
```diff
@@ -215,25 +237,7 @@ export async function skipPreviewAndSaveJob(
...
-        op.accessToken = token;
-        const serialOp = op.serialize(CURRENT_SCHEMA_VERSION);
-        workflowStep.operation = serialOp;
-        await workflowStep.save(tx);
-      }
-    }
-    job.skipPreview();
-    await job.save(tx);
-  });
+  await updateTokenAndChangeState(jobID, username, token, 'skipPreview');
```
indiejames, line 240: See previous comments

app/util/job.ts
indiejames: This comment does not apply if you refactor as described in my other comments.

indiejames (changes requested):

vinnyinverso (approved):

app/util/job.ts
chris-durbin: That's where I started, but the transaction does not start at the beginning of the code so it makes sense that it does the entire thing.

I really want to just pass in the function to call but from a bunch of searching I cannot find a way you can do that with a method on an object (thought that maybe Symbol construction could work). Anyway what I ended up with is the best I could come up with. I'll give it more thought today.

chris-durbin: Oh - if the common function returns a database transaction I could use the transaction to make the status change in the calling function... Feels a little clunky to return a database connection but it seems like passing in a string to tell it what to do is also clunky. Not sure which would be better.

chris-durbin: Figured out how to do my original intent by passing in a function that takes a job as an argument. It's no longer clunky.

app/util/job.ts
chris-durbin: Refactored.

app/util/job.ts
chris-durbin: Refactored.

app/util/job.ts
chris-durbin: Refactored.

app/util/job.ts
```diff
@@ -194,11 +192,32 @@ export async function resumeAndSaveJob(
...
+ * belong to the user.
+ */
+export async function resumeAndSaveJob(
+  jobID: string,
+  _logger: Logger,
+  username?: string,
+  token?: string,
+
+): Promise<void> {
+  await updateTokenAndChangeState(jobID, username, token, ((job) => job.resume()));
```
indiejames, line 218: Nice use of anonymous functions.

indiejames (approved):

PR #320, opened 2023-01-06
Harmony 1327 - Add date filters to workflow ui

vinnyinverso:
## Jira Issue ID
harmony-1327

## Description
This adds date filters to the jobs and work items page of the workflow ui.
If you have a larger list of items or jobs, and scroll down the page, the filter inputs should also stick to the top of the page so that you can still see them as you scroll.

## Local Test Steps
Submit some requests, navigate to the workflow ui, and try out the filters. Everything should be in your browser timezone.

## PR Acceptance Checklist
* [x] Acceptance criteria met


chris-durbin (commented):
A couple of small things I noticed:

1. I think when setting a date and time filter on the main page we should carry over the same time range to the work items page.
2. Neither of the radio buttons is selected at first, it seems like when you apply a search when neither is selected it uses createdAt - we should probably default to that.
3. There's no text on the time picker to indicate you are finished setting your time. Maybe add a link in the bottom right corner with 'set' as the text to close the modal?

indiejames (approved):
Tested locally

chris-durbin (approved):

PR #325, opened 2023-01-24
HARMONY-1214: Add destinationUrl parameter to harmony api.

ygliuvt:
## Jira Issue ID
HARMONY-1214

## Description
Add destinationUrl parameter to harmony api. This is the first step of adding support of user specified s3 destination url. It only saves the provided destinationUrl in the jobs table. Followup tickets will add support for validations and staging using the destination url.

## Local Test Steps
curl -Ln -bj "http://localhost:3000/C1233800302-EEDTEST/ogc-api-coverages/1.0.0/collections/all/coverage/rangeset?granuleId=G1233800343-EEDTEST&destinationUrl=s3://dummy" -o ~/Downloads/output.tif

and verify the request completes successfully and the `destination_url` field in jobs table is populated with `s3://dummy` as indicated in the request parameter.

## PR Acceptance Checklist
* [x] Acceptance criteria met

### DIFF
--- a/README.md
+++ b/README.md
@@ -138,6 +138,7 @@ define the environment variables needed to run the service and execute the local
 You can do this with the following steps:
 
 1. Build the image for your service
+
 2. Add entries into the `env-defaults` file for your service. See the `HARMONY_SERVICE_EXAMPLE`
    entries for examples. Be sure to prefix the entries with the name of your service.
    Set the value for the `INVOCATION_ARGS` environment variable. This should be how you would run
@@ -150,12 +151,13 @@ You can do this with the following steps:
   ```shell
   MY_SERVICE_INVOCATION_ARGS='python -m my-service'
   ```
+
 3. Add an entry for your service (lowercase) to the `.env` file:
 ```shell
 LOCALLY_DEPLOYED_SERVICES=my-service
 ```
-Note that the name used must be the kebab case version of the environment variable prefix used
-in `env-defaults`.
+Note that the name used must be the kebab case version of the environment variable prefix used in `env-defaults`.
+
 4. Run
 ```bash
 ./bin/deploy-services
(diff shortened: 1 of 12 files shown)


app/models/job.ts
```diff
@@ -355,6 +356,8 @@ export class Job extends DBRecord implements JobRecord {
   collectionIds: string[];
 
   ignoreErrors: boolean;
+  
+  destination_url?: string;
```
vinnyinverso, line 360: Is it possible to use camel case like the other properties, or was the use of underscore intentional?

vinnyinverso (commented):
Great job -- only had one question.

app/models/job.ts
ygliuvt: The underscore here is intentional as I was told that we have trouble with camel case column names in Postgres and want to move to field names with underscores. Chris has started to use snake case for column names in the new table `user_work`. The name of the variable actually maps to the real column name in the database table, so I used snake case here even though it is an eye sore.

README.md
```diff
@@ -173,7 +175,7 @@ To use Earthdata Login with a locally running Harmony, you must first set up a n
...
+You must select "401" as the application type for Harmony to work correctly with command line clients and clients like QGIS. Set the redirect URL to http://localhost:3000/oauth2/redirect for local Harmony. Leave Required User information empty. Redirect Time for Earthdata Login Splash page (in seconds): 3 seconds. Check the checkbox for By checking this box, I confirm that my application is compatible with EDL policy. Leave the other checkbox unchecked. Then create the new application. After the application is created, you can use the "manage" -> "App Groups" tab to add the "EOSDIS Enterprise" group to the application. This "EOSDIS Enterprise" group will allow CMR searches issued by Harmony to be able to use your Earthdata Login tokens.  Update `OAUTH_CLIENT_ID`, `OAUTH_UID` and `OAUTH_PASSWORD` in .env with the information from your Earthdata Login application.
```
chris-durbin, line 178: I despise the Earthdata Login Splash and would never recommend setting it to anything above 0. There's no reason for artificial sleeps. I would at least remove the sentence from our documentation "Redirect Time for Earthdata Login Splash page (in seconds): 3 seconds."

README.md
```diff
@@ -202,7 +204,7 @@ Optional:
 * [awscli-local](https://github.com/localstack/awscli-local) - CLI helpers for interacting with localstack
 * [Python](https://www.python.org) version 3.7 - Useful for locally running and testing harmony-docker and other backend services
 
-## Running Harmony (Not From Quick Start)
+## Running Harmony For Local Dev (Not From Quick Start)
```
chris-durbin, line 207: spell out development

app/models/data-operation.ts
```diff
@@ -289,6 +289,8 @@ export default class DataOperation {
 
   ignoreErrors?: boolean;
 
+  destinationUrl: string;
```
chris-durbin, line 292: We should remove this line. There is already a stagingLocation in the data operation model. We should set the stagingLocation as appropriate if there is a destinationUrl for the job.

app/models/services/base-service.ts
```diff
@@ -321,6 +321,7 @@ export default abstract class BaseService<ServiceParamType> {
       message: message,
       collectionIds: this.operation.collectionIds,
       ignoreErrors: this.operation.ignoreErrors,
+      destination_url: this.operation.destinationUrl,
```
chris-durbin, line 324: I'm trying to figure out the right logic for this, but essentially if the `operation.stagingLocation` is set to something other than the expected default I think we would want to set the destination_url on the request. I think we need to know if the destination_url is external for later tickets, so I think we do want to differentiate between external or our own default location.

app/models/services/no-op-service.ts
```diff
@@ -63,6 +63,7 @@ export default class NoOpService extends BaseService<void> {
       links,
       request: getRequestUrl(req),
       numInputGranules: this.operation.cmrHits,
+      destination_url: this.operation.destinationUrl,
```
chris-durbin, line 66: This will return a destination url for the no op service. I think we should eventually get rid of the no op service, but in any case since we aren't doing anything and won't be staging anything to a bucket we should remove this line.

chris-durbin (commented):
We probably want to talk through how we pass along the destinationUrl provided by the end user request and the already existing stagingLocation in the data operation.

app/models/services/base-service.ts
ygliuvt: Right. We can talk about this in the technical meeting today.

vinnyinverso (approved):

chris-durbin (approved):
Tested locally and verified the destinationUrl parameter is now accepted and saved to the database. Regression tested everything works when setting and not setting the parameter. We discussed and agreed no changes need to be made to the way we are associating the destinationUrl with the data operation right now, we are not adding it as part of the model. We will have a separate ticket for using the request context instead of the data operation in the future.

PR #327, opened 2023-01-25
DAS-1766 - Do not batch NetCDF-to-Zarr outputs

owenlittlejohns:
## Jira Issue ID

DAS-1766

## Description

This PR updates all NetCDF-to-Zarr services chains to prevent the output from being batched.

I also had to update some of the unit tests, as they were using one of the NetCDF-to-Zarr service chains to verify ignore-errors functionality with a batched service. I swapped those tests over to use a service chain of the PO.DAAC L2 Subsetter and Concise services.

## Local Test Steps

* Pull this branch.
* Restart your Harmony to use the updated configuration file.
* Perform the following request:
```
localhost:3000/C1245618475-EEDTEST/ogc-api-coverages/1.0.0/collections/all/coverage/rangeset?forceAsync=true&format=application%2Fx-zarr&concatenate=true&maxResults=450
```

You should see that there is only one output Zarr store.

## PR Acceptance Checklist
* [x] Acceptance criteria met


chris-durbin:
We still have netcdf-to-zarr configured to not concatenate by default. Are we ready to switch it back to concatenate by default (meaning netcdf-to-zarr is ready for unlimited granules in a request)?

test/ignore-errors.ts
```diff
@@ -24,8 +24,16 @@ const reprojectAndZarrQuery = {
...
+const l2ssAndConciseQuery = {
+  subset: 'lat(0:90)',
+  concatenate: false,
+  maxResults: 1,
+  ignoreErrors: true,
+};
+
 describe('when setting ignoreErrors=true', function () {
   const collection = 'C1233800302-EEDTEST';
+  const l2ssCollection = 'C1243729749-EEDTEST';
```
chris-durbin, line 36: Thanks for finding another EEDTEST collection for us to use and sorry it required so many updates... there's a lot of branching in the code but clearly a little too much repetition in the tests!

owenlittlejohns:
The concatenation default is a good question. I think we're probably best to keep things as they are for now (not doing concatenation by default) until the completion of TRT-316/DAS-1752, which should really make things unlimited. Once that's working we should switch over to concatenate by default. Removing the batching now will enable that work to be tested at a large scale as it is done this PI.

I'll leave a comment on DAS-1752 to ensure that update to `service.yml` at the end doesn't get forgotten.

test/ignore-errors.ts
```diff
@@ -700,89 +708,89 @@ describe('when setting ignoreErrors=true', function () {
...
 
         before(async function () {
           let shouldLoop = true;
           // retrieve and fail work items until one exceeds the retry limit and actually gets marked as failed
           while (shouldLoop) {
-            const res = await getWorkForService(this.backend, 'sds/swot-reproject:latest');
-            firstSwotItem = JSON.parse(res.text).workItem;
-            firstSwotItem.status = WorkItemStatus.FAILED;
-            firstSwotItem.results = [];
+            const res = await getWorkForService(this.backend, 'ghcr.io/podaac/l2ss-py:sit');
```
flamingbear, line 721: Should the harmony tests be using the SIT version of this? Is that because it's not in PROD yet?

owenlittlejohns: I mimicked this from `test/concatenation.ts`, where the PO.DAAC SIT images are used. I've not plumbed the depths to find exactly where the available services are defined, so I'm not sure if a production-tagged image would also work.

owenlittlejohns:
@chris-durbin - I just wanted to check if you were good with this being merged? If so I'll update the branch with the latest from `main` and merge this PR. Thanks!

chris-durbin (approved):

owenlittlejohns:
I ran `npm audit` and `npm audit fix` to resolve the vulnerability reported in the failed builds. (Looks like it was a nested dependency, so only ended up changing `package-lock.json`, not `package.json`)

PR #329, opened 2023-01-31
HARMONY-1216: Added destinationUrl s3 bucket same AWS region validation

ygliuvt:
## Jira Issue ID
HARMONY-1216

## Description
As ESDIS, I want Harmony to produce a clear, actionable error if a user specifies an S3 location outside of Harmony's region, so I can avoid egressing all results

## Local Test Steps
This has to be tested outside of localstack. e.g. 
http://internal-harmony-yliu10-frontend-xxxx.us-west-2.elb.amazonaws.com/C1233800302-EEDTEST/ogc-api-coverages/1.0.0/collections/all/coverage/rangeset?granuleId=G1233800343-EEDTEST&format=image%2Fpng&destinationUrl=s3%3A%2F%2Fyliu10-east-1-bucket%2Fabcd

## PR Acceptance Checklist
* [x] Acceptance criteria met


app/middleware/parameter-validation.ts
```diff
@@ -30,6 +34,46 @@ function validateLinkTypeParameter(req: HarmonyRequest): void {
...
+ * Validate that the bucket name provided is in the same AWS region as the given region
+ *
+ * @param bucketName - The name of the s3 bucket
+ * @param region - The name of the aws region
+ */
+async function validateBucketIsInRegion(bucketName: string, region: string): Promise<void> {
+  try {
+    const bucketRegion = await defaultObjectStore().getBucketRegion(bucketName);
+    if (bucketRegion != region) {
+      throw new RequestValidationError(`Destination bucket '${bucketName}' must be in the '${region}' region, but was in '${bucketRegion}'.`);
```
ygliuvt, line 47: This message is not exactly actionable. I think we want to write some kind of document first which is out of scope for this ticket, then reference it here.

chris-durbin: Agreed - the scope of this ticket is for the validation of the region. As part of implementing the ticket that provides full documentation on the destinationUrl issue we should update this documentation - you can throw a TODO comment with the ticket number in here to make tracking it easier.

app/middleware/parameter-validation.ts
```diff
@@ -30,6 +34,46 @@ function validateLinkTypeParameter(req: HarmonyRequest): void {
...
+ * @param req - The client request
+ */
+async function validateDestinationUrlParameter(req: HarmonyRequest): Promise<void> {
+  const keys = keysToLowerCase(req.query);
+  const destUrl = keys.destinationurl?.toLowerCase();
+  if (destUrl) {
+    if (!destUrl.startsWith('s3://')) {
+      throw new RequestValidationError(`Invalid destinationUrl '${destUrl}' must start with s3://`);
+    }
+    const bucketName = destUrl.substring(5).split('/')[0];
```
chris-durbin, line 69: I'm assuming it is not possible for this to throw an exception? Otherwise I think an internal error would go back to the user.

chris-durbin (approved):
Paired while testing this out in Sandbox. Validations look good and a good set of tests.

app/middleware/parameter-validation.ts
ygliuvt: The ticket for writing the cross account s3 delivery setup document is HARMONY-1218.

app/middleware/parameter-validation.ts
ygliuvt: Yes, the check above should prevent any exception from being thrown in this call.

indiejames (approved):

PR #330, opened 2023-02-01
Harmony 1342 granule name

indiejames:
## Jira Issue ID
HARMONY-1342

## Description
Adds a `granuleName` parameter to Harmony. This corresponds to the `readable_granule_name` in the CMR. Wildcards `*` and `?` are supported.

NOTE: I also refactored some of the other parameter handling to avoid so many `if` statements in `get-coverage-rangesets.ts` - hopefully more readable.

## Local Test Steps
* checkout this branch
* rebuild the query-cmr image
```
cd  tasks/queyr-cmr && npm run build
```
* run Harmony
* issue a query using `granuleName`
```
http://localhost:3000/C1234088182-EEDTEST/ogc-api-coverages/1.0.0/collections/blue_var/coverage/rangeset?format=image%2Fpng&maxResults=1&granuleName=2020_01_01_7f00ff_antimeridia*
```

## PR Acceptance Checklist
* [x] Acceptance criteria met


app/schemas/ogc-api-coverages/1.0.0/ogc-api-coverages-v1.0.0.yml
```diff
@@ -911,3 +913,14 @@ components:
...
+      name: granuleName
+      in: query
+      description: |
+        passed to the CMR search as the readable_granule_name parameter. Supports * and ? wildcards
+        for multiple and single character matches. Wildcards can be used any place in the name,
+        but leading wildcards are discouraged as they require a lot of resources for the underlying
+        search.
+      required: false
+      schema:
+        type: string
```
ygliuvt, line 926: Do we only want to allow a single `granuleName` value for this parameter? CMR `readable_granule_name` supports multiple values in an array.

indiejames: It supports a comma separated list

ygliuvt: I think the code actually support multiple values in an array. So we need to update here to reflect that.

test/parameters/granule_name.ts
```diff
@@ -0,0 +1,50 @@
...
+const regularQuery = {
+  format: 'application/x-zarr',
+  granuleName: '001_08_7f00ff_oceania_east',
+};
+
+const wildCardQuery = {
+  format: 'application/x-zarr',
+  granuleName: '001_*',
+};
+
```
ygliuvt, line 16: Maybe add a multiple granuleName values test for completeness.

app/schemas/ogc-api-coverages/1.0.0/ogc-api-coverages-v1.0.0.yml
indiejames: as well as multiple entries

ygliuvt: I see. This is not obvious without reading the actual parsing function in the code. As an API doc, we want to make this clear here. It is also not following the pattern set by the `granuleId` parameter. If there are people actually using the comma separated string for `granuleId`, we probably should update the documentation for `granuleId` as well.

test/parameters/granule_name.ts
indiejames: Done.

ygliuvt (approved):

app/schemas/ogc-api-coverages/1.0.0/ogc-api-coverages-v1.0.0.yml
indiejames: This uses exactly the pattern that granuleId uses.

ygliuvt: `granuleId` is of type array:
```
granuleId:
      name: granuleId
      in: query
      description: The CMR Granule ID for the granule which should be retrieved
      required: false
      style: form
      explode: true
      schema:
        type: array
        items:
          type: string
          minLength: 1
```

chris-durbin: Agree with Leo that the schema should be the same as granuleId - specifically:
```
      style: form
      explode: true
      schema:
        type: array
        items:
          type: string
          minLength: 1
```
instead of
```
     schema:
        type: string
```

indiejames: I see - my mistake. The code that handles the multi-value parameters is the same, but the schema was not. Fixed.

chris-durbin (approved):
Tested out locally with wildcards as well as multiple granuleName parameters and the results found were correct.

PR #336, opened 2023-02-08
Harmony 1369 - Handle cases where work item failure messages exceed job message column limit

vinnyinverso:
## Jira Issue ID
HARMONY-1369

## Description
This fixes a bug that we saw in production which was causing jobs to not get marked as failed even though a work item had failed. The work item failure message, which gets used in the ultimate job failure message was far too large to fit within the postgres column limits.
We are now truncating those failure message to avoid this case. The other status messages mostly have a fixed length, so do not need to be handled in any special way.

## Local Test Steps
In the workflow orchestration file in `updateWorkItem` you can hard code `status` to `WorkItemStatus.FAILED` and set `failureMessage` to something really big (greater than 4096 chars). After submitting a request, the job should fail and the associated job message should be truncated and end with `...`.

## PR Acceptance Checklist
* [x] Acceptance criteria met


app/models/job.ts
```diff
@@ -999,10 +999,14 @@ export class Job extends DBRecord implements JobRecord {
    * @throws {@link Error} if the job is invalid
    */
   async save(transaction: Transaction): Promise<void> {
+    const textLimit = 4096; // this.request and this.message need to be under the 4,096 char limit
+    const reservedMessageChars = 1000; // reserve 1k chars for non-failure messages (which tend to be smaller)
     // Need to validate the original status before removing it as part of saving to the database
     // May want to change in the future to have a way to have non-database fields on a record.
     this.validateStatus();
-    this.request = truncateString(this.request, 4096);
+    const tuncatedFailureMessage = truncateString(this.getMessage(JobStatus.FAILED), textLimit - reservedMessageChars);
```
chris-durbin, line 1007: truncated typo

chris-durbin (approved):
Reproduced the problem on the main branch and verified this change fixed it. Saw a small issue with the way the error message is displayed on the workflow-ui scrolling way off to the right of the error box rather than filling the red box by word wrapping.

PR #340, opened 2023-02-09
HARMONY-1221: Remove dataExpiration from job status and expire from STAC item when destinationUrl is set

indiejames:
## Jira Issue ID
HARMONY-1221

## Description
Removes the data expirations fields from the status page and the STAC items if the destinationUrl parameter is set

## Local Test Steps
1. Create a custom staging bucket in localstack
aws --endpoint-url=http://localhost:4566 s3 mb s3://my-staging-bucket
2. Check out this branch
3. Run harmony
4. Issue the following query
http://localhost:3000/C1233800302-EEDTEST/ogc-api-coverages/1.0.0/collections/blue_var/coverage/rangeset?format=image%2Fpng&granuleName=001_*&destinationUrl=s3%3A%2F%2Fmy-staging-bucket
5. Verify that the status page does not have the `dataExpiration` field
6. When the job is finished, refresh the status page then follow the link to the STAC catalog. Append '/0' to the url of the STAC catalog to get the first STAC item back. Verify that the timestamps do not include an 'expires' field.
7. Repeat the above, but remove the destinationUrl parameter and verify that the `dataExperiation` field and `expires` field are back

## PR Acceptance Checklist
* [x] Acceptance criteria met

### DIFF
--- a/app/models/job.ts
+++ b/app/models/job.ts
@@ -356,7 +356,7 @@ export class Job extends DBRecord implements JobRecord {
   collectionIds: string[];
 
   ignoreErrors: boolean;
-  
+
   destination_url?: string;
 
   /**
@@ -1041,6 +1041,9 @@ export class Job extends DBRecord implements JobRecord {
       numInputGranules: this.numInputGranules,
       jobID: this.jobID,
     };
+    // need this line to prevent null values fro showing up in data expiration field
+    Object.keys(serializedJob).forEach((k) => serializedJob[k] == null && delete serializedJob[k]);
+
     if (urlRoot && linkType !== 'none') {
       serializedJob.links = serializedJob.links.map((link) => {
         const serializedLink = link.serialize();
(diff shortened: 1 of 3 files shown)


app/models/job.ts
```diff
@@ -1041,6 +1041,9 @@ export class Job extends DBRecord implements JobRecord {
       numInputGranules: this.numInputGranules,
       jobID: this.jobID,
     };
+    // need this line to prevent null values fro showing up in data expiration field
```
chris-durbin, line 1044: from

chris-durbin (approved):

app/models/job.ts
indiejames: Fixed

app/models/job.ts
```diff
@@ -1041,6 +1041,9 @@ export class Job extends DBRecord implements JobRecord {
...
+    Object.keys(serializedJob).forEach((k) => serializedJob[k] == null && delete serializedJob[k]);
```
vinnyinverso, line 1045: There is a helper function `removeEmptyProperties` that you could use.

vinnyinverso (approved):

ygliuvt (approved):

app/models/job.ts
indiejames: Done.

PR #344, opened 2023-02-15
HARMONY-1357: As a Harmony user I want to store my results in my own S3 bucket.

ygliuvt:
## Jira Issue ID
HARMONY-1357

## Description
This PR completes the normal workflow of putting OGC request results in the user specified s3 bucket.
Changes made:

1. Changed normal workflow orchestration to put intermediary results in the artifacts bucket rather than the staging bucket so that the intermediary results are not leaked to the client user. The existing naming convention of intermediary results remains the same. e.g. `s3://local-artifact-bucket/public/podaac/l2-subsetter-concise/0b6a7a69-2084-4265-bfd7-02f92f3f5cb6/315/JA1_GPS_2PeP220_111_20071231_005214_20071231_014826_bathymetry_subsetted.nc4`.

2. Changed normal workflow orchestration to put the final result in the staging bucket using a new naming convention: `<stagingLocation>/public/<jobId>/<stepId>/...`.

3. When destinationUrl is present, put the final result in the specified destinationUrl location: `<destinationUrl>/<jobId>/<stepId>/...`.

4. Changed STAC result links to not include cloud access JSON and sh links when destinationUrl is present.

## Local Test Steps
1. Set up the necessary services locally to run a multi-step OGC request. e.g. 
```
curl -Ln -bj "http://localhost:3000/C1234208438-POCLOUD/ogc-api-coverages/1.0.0/collections/bathymetry/coverage/rangeset?concatenate=true&subset=lon(-160%3A160)&subset=lat(-80%3A80)&skipPreview=true&maxResults=3"
```
verify that the request is successful and the intermediary results are placed into the artifacts bucket (see Description above for detail) and only the final result is placed in the staging bucket. 
verify the result STAC has the correct links to the final staging location.

2. Run the same multi-step OGC request with a valid destinationUrl that has the correct permissions set up. e.g. 
```
curl -Ln -bj "http://localhost:3000/C1234208438-POCLOUD/ogc-api-coverages/1.0.0/collections/bathymetry/coverage/rangeset?concatenate=true&subset=lon(-160%3A160)&subset=lat(-80%3A80)&skipPreview=true&maxResults=3&destinationUrl=s3://test-bucket/www"
```
verify that the request is successful and the intermediary results are placed into the artifacts bucket (see Description above for detail) and only the final result is placed in the destinationUrl (see Description above for detail). 

verify under destinationUrl, there is a path created with the requestId of the above request and a file named `harmony-job-status-link` is placed there. Verify the content of the file `harmony-job-status-link` is the job status link of the above request. e.g. `https://harmony.sit.earthdata.nasa.gov/jobs/1f215d92-0bd8-4255-9530-516deb147690`.

verify the job status link shows the result STAC with the correct link to the final result in the destinationUrl location and there is no cloud access JSON and sh links.

## PR Acceptance Checklist
* [x] Acceptance criteria met

### DIFF
--- a/app/models/services/base-service.ts
+++ b/app/models/services/base-service.ts
@@ -216,7 +216,7 @@ export default abstract class BaseService<ServiceParamType> {
       destPath = destPath.endsWith('/') ? destPath.slice(0, -1) : destPath;
       return defaultObjectStore().getUrlString(destPath, requestId + '/');
     } 
-    return defaultObjectStore().getUrlString(env.stagingBucket, 'public/' + requestId + '/');
+    return defaultObjectStore().getUrlString(env.stagingBucket, `public/${requestId}/`);
   }
 
   /**
--- a/app/models/services/base-service.ts
+++ b/app/models/services/base-service.ts
@@ -190,7 +190,7 @@ export default abstract class BaseService<ServiceParamType> {
 
     if (!this.operation.stagingLocation) {
       const prefix = `public/${config.name || this.constructor.name}/${uuid()}/`;
-      this.operation.stagingLocation = defaultObjectStore().getUrlString(env.stagingBucket, prefix);
+      this.operation.stagingLocation = defaultObjectStore().getUrlString(env.artifactBucket, prefix);
     }
   }
 
@@ -202,6 +202,21 @@ export default abstract class BaseService<ServiceParamType> {
    */
   get capabilities(): ServiceCapabilities {
     return this.config.capabilities;
+  }
+
+  /**
+   * Returns the final staging location of the service.
+   *
+   * @returns the final staging location of the service
+   */
+  finalStagingLocation() : string {
+    const { requestId, destinationUrl } = this.operation;
+    if (destinationUrl) {
+      let destPath = destinationUrl.substring(5);
+      destPath = destPath.endsWith('/') ? destPath.slice(0, -1) : destPath;
+      return defaultObjectStore().getUrlString(destPath, requestId + '/');
+    } 
+    return defaultObjectStore().getUrlString(env.stagingBucket, 'public/' + requestId + '/');
   }
 
   /**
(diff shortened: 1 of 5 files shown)


test/models/services.ts
```diff
@@ -831,5 +831,49 @@ describe('createWorkflowSteps', function () {
     it('creates a third and final workflow step for the shapefile subsetter', function () {
       expect(steps[2].serviceID).to.equal('shapefile subsetter');
     });
+
+    it('first step uses artifact bucket as the staging location', function () {
```
chris-durbin, line 835: Nit: the `it` is supposed to be read as if part of a sentence - so something like
```
it('uses the artifact bucket as the staging location for the first step',
```

app/models/services/base-service.ts
```diff
@@ -204,6 +204,21 @@ export default abstract class BaseService<ServiceParamType> {
...
+   * @returns the final staging location of the service
+   */
+  finalStagingLocation() : string {
+    const { requestId, destinationUrl } = this.operation;
+    if (destinationUrl) {
+      let destPath = destinationUrl.substring(5);
+      destPath = destPath.endsWith('/') ? destPath.slice(0, -1) : destPath;
+      return defaultObjectStore().getUrlString(destPath, requestId + '/');
+    } 
+    return defaultObjectStore().getUrlString(env.stagingBucket, 'public/' + requestId + '/');
```
chris-durbin, line 219: Nit: style thing - if you find you are adding more than one + to concatenate a string it might be more readable as `public/${requestId}/`. I'm not really sure which is more readable, but figured you'd want to know there's another way too.

chris-durbin (approved):
Looks good. Note that I haven't tested it out, make sure to get feedback from James' testing before merging in.

app/models/services/base-service.ts
ygliuvt: I like your suggestion better.

indiejames (approved):
Tested in sandbox

PR #346, opened 2023-02-16
Harmony 1368 - Add service filter to the workflow ui

vinnyinverso:
## Jira Issue ID
HARMONY-1368

## Description
Users can now filter their jobs by service.
Screen Shot 2023-02-16 at 1 50 51 PM

## Local Test Steps
1 - Run the database migrations
2 - Submit a couple of requests for different services. Utilize the filters on the workflow ui (e.g. `service: harmony/service-example`) to filter jobs by service. Utilize the "negate services" check box to find jobs that do _not_ match the specified services.

Note that the filter will only work for jobs that were run after this change (i.e. the filter will not work for jobs where service_name=undefined).

## PR Acceptance Checklist
* [x] Acceptance criteria met


app/frontends/workflow-ui.ts
```diff
@@ -128,13 +139,18 @@ function jobRenderingFunctions(logger: Logger, requestQuery: Record<string, any>
       try {
         const url = new URL(this.request);
         const path = url.pathname + url.search;
-        return path;
+        return truncateString(path, 100);
```
indiejames, line 142: Maybe move these hardcode lengths into a constant or env var to make it easy to change them all at once if needed.

app/frontends/workflow-ui.ts
```diff
@@ -128,13 +139,18 @@ function jobRenderingFunctions(logger: Logger, requestQuery: Record<string, any>
...
       } catch (e) {
         logger.error(`Could not form a valid URL from job.request: ${this.request}`);
         logger.error(e);
         return this.request;
       }
     },
+    jobMessage(): string {
+      if (this.message) {
+        return truncateString(this.message, 100);
+      }
```
indiejames, line 152: Same comment about hard-coded length

indiejames (approved):
Tested locally. Other than a small comment about hard-coded values, the only thing I noticed was that when I started typing a service filter it would show ghosted completion suggestions. When I see that my expectation is that I will be able to use tab completion, but it doesn't do anything here. Not necessary for this PR, but might be a nice-to-have.

chris-durbin (approved):
Tested this out successfully and it's really useful. One thing I noticed is that we are calling each chain a service in this context. So if you wanted to for example get a list of any jobs that used netcdf-to-zarr (the underlying service image) you would want to add filters for all of the chains that include netcdf-to-zarr (like harmony/swot-repr-netcdf-to-zarr and harmony/podaac-l2-subsetter-netcdf-to-zarr and the vanilla harmonyservices/netcdf-to-zarr). Talking it through with Vinny that does seem like the best approach.

owenlittlejohns (approved):
Thanks for this update! I tried it out locally and it makes filtering my queries really slick. I also like that the filters can be combined (e.g., status and service, or a couple of services).

PR #350, opened 2023-02-22
Harmony 1377 - Make any request that uses a destination_url asynchronous

vinnyinverso:
## Jira Issue ID
HARMONY-1377

## Description
Makes any request that uses a destination_url asynchronous.

Sync requests are incompatible with the destination_url parameter. To paraphrase what Chris said -- if they are asking us to put data in their bucket we don't want to stream the response back.

## Local Test Steps
1- create a local bucket, e.g. `awslocal s3 mb s3://vin2 --region us-west-2`
2- submit a request that would otherwise be synchronous if the destinationUrl were not set: http://localhost:3000/C1233800302-EEDTEST/ogc-api-coverages/1.0.0/collections/all/coverage/rangeset?maxResults=1&destinationUrl=s3://vin2
3- make sure you're redirected to the job status page (the job was made async)

## PR Acceptance Checklist
* [x] Acceptance criteria met


app/models/services/base-service.ts
```diff
@@ -544,6 +544,9 @@ export default abstract class BaseService<ServiceParamType> {
     if (operation.requireSynchronous) {
```
ygliuvt, line 544: This is interesting -- I just found out that WMS requests must be synchronous. I failed to find a description of WMS request in Harmony documentation. We should consider put this in the future Harmony landing documentation page.

ygliuvt:
Not certain about this: Should we put the this asynchronous nature of destinationUrl request in the OGC api doc? Feels like we are really light on the swagger api document and I don't know if we want to put this kind of info in there. I am ok with not addressing it in this PR.

ygliuvt (approved):

vinnyinverso:
> Should we put the this asynchronous nature of destinationUrl request in the OGC api doc?

Makes sense to me.

chris-durbin (approved):

PR #353, opened 2023-02-24
HARMONY-1379: Fixed confusing error message when harmony rangeset request URL is invalid.

ygliuvt:
## Jira Issue ID
HARMONY-1379

## Description
Fixed confusing error message when harmony rangeset request URL is invalid.

## Local Test Steps
http://localhost:3000/C1233800302-EEDTEST/ogc-api-coverages/1.0.0/collections/all/coverage/rangeset123?granuleId=G1233800343-EEDTEST

verify the returned status code is 404 with error message: 
`{"code":"harmony.NotFoundError","description":"Error: The requested page was not found."}`

## PR Acceptance Checklist
* [x] Acceptance criteria met

### DIFF
--- a/app/routers/router.ts
+++ b/app/routers/router.ts
@@ -200,8 +200,8 @@ export default function router({ skipEarthdataLogin = 'false' }: RouterConfig): 
   result.get('/versions', asyncHandler(getVersions));
   result.use('/docs/api', swaggerUi.serve, swaggerUi.setup(yaml.load(ogcCoverageApi.openApiContent), { customJs: '/js/docs/analytics-tag.js' }));
   result.get(collectionPrefix('wms'), asyncHandler(service(serviceInvoker)));
-  result.get(/^.*?\/ogc-api-coverages\/.*?\/collections\/.*?\/coverage\/rangeset/, asyncHandler(service(serviceInvoker)));
-  result.post(/^.*?\/ogc-api-coverages\/.*?\/collections\/.*?\/coverage\/rangeset/, asyncHandler(service(serviceInvoker)));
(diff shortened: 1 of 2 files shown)


app/routers/router.ts
```diff
@@ -200,8 +200,8 @@ export default function router({ skipEarthdataLogin = 'false' }: RouterConfig):
   result.get('/versions', asyncHandler(getVersions));
   result.use('/docs/api', swaggerUi.serve, swaggerUi.setup(yaml.load(ogcCoverageApi.openApiContent), { customJs: '/js/docs/analytics-tag.js' }));
   result.get(collectionPrefix('wms'), asyncHandler(service(serviceInvoker)));
-  result.get(/^.*?\/ogc-api-coverages\/.*?\/collections\/.*?\/coverage\/rangeset/, asyncHandler(service(serviceInvoker)));
-  result.post(/^.*?\/ogc-api-coverages\/.*?\/collections\/.*?\/coverage\/rangeset/, asyncHandler(service(serviceInvoker)));
+  result.get(/^.*?\/ogc-api-coverages\/.*?\/collections\/.*?\/coverage\/rangeset$/, asyncHandler(service(serviceInvoker)));
```
chris-durbin, line 203: Note that there is one case that this returns a 404 now, but used to succeed before (when the url included a trailing slash on the endpoint).

We could keep the same behavior as before by allowing `/rangeset\/?$/`, but it probably should have always failed. I think the change in the behavior probably makes sense to not allow a trailing slash, but wanted to point it out in case there is a reason we should allow it.

chris-durbin (approved):

indiejames (approved):

app/routers/router.ts
ygliuvt: It makes sense to support the / at the end. I will update the regex.

PR #367, opened 2023-03-17
HARMONY-1386: Add collection capabilities endpoint

chris-durbin:
## Jira Issue ID
HARMONY-1386

## Description
Added an endpoint to get back a list of capabilities for a collection in JSON format. HARMONY-1393 was written to handle the HTML view. You can ignore all of the HTML in this PR, it is not used.

## Local Test Steps

Test pulling back collection capabilities for various collections and make sure that the returned values make sense for that collection.

Examples:
http://localhost:3000/capabilities?collectionId=C1225776654-ASF
http://localhost:3000/capabilities?shortName=harmony_example
http://localhost:3000/capabilities?collectionId=C1233800302-EEDTEST

## PR Acceptance Checklist
* [X] Acceptance criteria met


app/frontends/capabilities.ts
```diff
@@ -0,0 +1,141 @@
...
+ * @throws RequestValidationError if no collection is found or parameters are invalid
+ */
+async function loadCollectionInfo(req: HarmonyRequest): Promise<CmrCollection> {
+  const query = keysToLowerCase(req.query);
+  const { collectionid, shortname } = query;
+  let collections;
+  let pickedCollection;
+  if (!collectionid && !shortname) {
+    throw new RequestValidationError('Must specify either collectionId or shortName');
+  } else if (collectionid && shortname) {
```
indiejames, line 39: nit: these don't really need to be `else'd` since you are throwing in each case.

app/routers/router.ts
```diff
@@ -251,6 +252,9 @@ export default function router({ skipEarthdataLogin = 'false' }: RouterConfig):
 
   result.get('/admin/configuration/log-level', asyncHandler(setLogLevel));
 
+  result.get('/capabilities', asyncHandler(getCollectionCapabilitiesJson));
+  // Enable HTML view with HARMONY-1393
+  // result.get('/capabilities.html', asyncHandler(getCollectionCapabilitiesHtml));
```
indiejames, line 257: I thought you were planning to use the Accept header to determine what type to use for the response?

app/frontends/capabilities.ts
```diff
@@ -0,0 +1,141 @@
...
+      + 'is correct and that you have access to it.';
+      throw new RequestValidationError(message);
+    }
+    pickedCollection = collections[0];
+  } else {
+    collections = await getCollectionsByShortName(shortname, req.accessToken);
+    if (collections.length === 0) {
+      const message = `Unable to find collection short name ${shortname} in the CMR. Please `
+      + ' make sure the short name is correct and that you have access to the collection.';
+      throw new RequestValidationError(message);
```
indiejames, line 55: Again I think a 404 makes more sense here than a 400.

indiejames (commented):
I verified that the output looks by running locally. Most of my comments are nits, except for the one about 404 vs. 400. I'm willing to go with the crowd on that one, though.

app/frontends/capabilities.ts
chris-durbin: While redundant I prefer this style - IMO makes it more obvious when branches are taken vs not.

app/frontends/capabilities.ts
```diff
@@ -0,0 +1,141 @@
...
+    throw new RequestValidationError('Must specify either collectionId or shortName');
+  } else if (collectionid && shortname) {
+    throw new RequestValidationError('Must specify only one of collectionId or shortName, not both');
+  } else if (collectionid) {
+    collections = await getCollectionsByIds([collectionid], req.accessToken);
+    if (collections.length === 0) {
+      const message = `${collectionid} must be a CMR collection identifier, but `
+      + 'we could not find a matching collection. Please make sure the collection ID'
+      + 'is correct and that you have access to it.';
+      throw new RequestValidationError(message);
```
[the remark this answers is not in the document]
chris-durbin: I made it consistent with the way we handle it everywhere else in the code when we do not find matching collections/variables/granules. Right now those are 400s instead of 404s. I can see arguments for both.

app/frontends/capabilities.ts
chris-durbin: Same comment

app/frontends/capabilities.ts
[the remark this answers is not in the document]
chris-durbin: I like 404 better too - went ahead and changed.

app/frontends/capabilities.ts
chris-durbin: I like 404 better too - went ahead and changed.

indiejames (approved):

app/frontends/capabilities.ts
ygliuvt: nit: I would prefer refactor the validation code into its only function to make this large function more readable.

ygliuvt (approved):

PR #381, opened 2023-04-19
Harmony 1410 work item update queue

indiejames:
## Jira Issue ID
HARMONY-1410

## Description
Uses a queue for processing work-item updates.

## Local Test Steps
1. Pull this branch
2. Run the bin/start-postgres-localstack script (may need to delete the service/pod for localstack first)
3. List the queues in localstack - you should see two
```
> awslocal sqs list-queues
{
    "QueueUrls": [
        "http://localhost:4566/queue/large-work-item-update-queue",
        "http://localhost:4566/queue/work-item-update-queue"
    ]
}
```
4. Run a few queries against your local instance and verify that they work

## PR Acceptance Checklist
* [x] Acceptance criteria met


app/backends/workflow-orchestration.ts
```diff
@@ -755,10 +760,118 @@ export async function handleWorkItemUpdate(
...
+      await queue.deleteMessages(messages.map((msg) => msg.receipt));
+    } catch (e) {
+      defaultLogger.error(`Error deleting work item updates from queue: ${e}`);
+    }
+  }
+}
+
 /**
  * Update a work item from a service response. This function stores the update without further
  * processing and then responds quickly. Processing the update is handled asynchronously
```
ygliuvt, line 873: Update the comment to say the work item update is put on a queue.

app/backends/workflow-orchestration.ts
```diff
@@ -781,21 +894,20 @@ export async function updateWorkItem(req: HarmonyRequest, res: Response): Promis
     duration,
   };
   const workItemLogger = req.context.logger.child({ workItemId: update.workItemID });
-  if (typeof global.it === 'function') {
-    // tests break if we don't await this
-    await handleWorkItemUpdate(update, operation, workItemLogger);
+
+  // we use separate queues for small and large work item updates
+  let queueType = WorkItemUpdateQueueType.SMALL_ITEM_UPDATE;
+  if (results?.length > 1) {
```
ygliuvt, line 900: Not sure why when more than one results, we set the queue type to LARGE_ITEM_UPDATE.

ygliuvt (approved):

app/workers/work-item-update-queue-processor.ts
```diff
@@ -0,0 +1,47 @@
...
+    this.logger = config.logger;
+    this.queueType = config.queueType;
+  }
+
+  async start(): Promise<void> {
+    this.isRunning = true;
+    let firstRun = true;
+    while (this.isRunning) {
+      if (!firstRun) {
+        await sleep(env.workItemUpdateQueueProcessorPeriodSec * 1000);
```
chris-durbin, line 30: Why do we have this sleep here? Shouldn't it start immediately because of the
```
WaitTimeSeconds: 20
```
in the queue configuration? Does this parameter need to be in sync with that one?

env-defaults
```diff
@@ -158,6 +168,9 @@ REAPABLE_WORK_AGE_MINUTES=1440
 # The time (in seconds) between invocations of the work failer service
 WORK_FAILER_PERIOD_SEC=120
 
+# The time (in seconds) between invocations of the work-item update queue processor
+WORK_ITEM_UPDATE_QUEUE_PROCESSOR_PERIOD_SEC=1
```
chris-durbin, line 172: Ok I was looking at the logs it looks like things take 21 seconds when there is nothing to pull. So we have long polling of 20 seconds and then this 1 second delay where we won't pick up any messages. I don't really see any reason for ever delaying things so I think I'd like to just see this config removed and not have the threads processing the queues ever call sleep since we're using long polling.

chris-durbin (approved):
Really clean implementation despite having a file called queue-factory. Tested successfully locally. I only have one real comment which is to remove the sleep in between calls to get items from the queue.

app/workers/work-item-update-queue-processor.ts
indiejames: Moved this sleep into the `catch` block to not sleep normally, but will sleep if there is an error to avoid a tight loop.

env-defaults
indiejames: Fixed

app/backends/workflow-orchestration.ts
indiejames: Done.

PR #384, opened 2023-04-25
HARMONY-1423: As a harmony user, I want to be able to retrieve harmony STAC catalogs for requests without requiring EDL authentication.

ygliuvt:
## Jira Issue ID
HARMONY-1423

## Description
As a harmony user, I want to be able to retrieve harmony STAC catalogs for requests without requiring EDL authentication.
There has to be some kind of documentation that needs to be updated, but I can't find any.

## Local Test Steps
Submit a harmony request and see it run to completion successfully.
`curl "http://localhost:3000/stac/<job-id>"` and verify you can retrieve the stac catalog of the job.
`curl "http://localhost:3000/stac/<job-id>/<item-index>"` and verify you can retrieve the stac item of the job.

## PR Acceptance Checklist
* [x] Acceptance criteria met

### DIFF
--- a/test/stac/stac-item.ts
+++ b/test/stac/stac-item.ts
@@ -5,7 +5,7 @@ import itReturnsTheExpectedStacResponse from '../helpers/stac-item';
 import { buildJob } from '../helpers/jobs';
 import hookServersStartStop from '../helpers/servers';
 import { hookTransaction } from '../helpers/db';
-import { stacItem, hookStacItem } from '../helpers/stac';
+import { hookStacItem } from '../helpers/stac';
 import { JobStatus } from '../../app/models/job';
 
 const runningJob = buildJob({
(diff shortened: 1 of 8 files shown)


test/stac/stac-item.ts
```diff
@@ -328,4 +315,178 @@ describe('STAC item route', function () {
       });
     });
   });
+
+  describe('For a guest user who does not own the job', function () {
```
chris-durbin, line 319: It looks like most of the steps here are copied from another test. I'd just keep the one for a 409 when the job isn't complete and then a single 200 with JSON and get rid of all the other parts that are already tested elsewhere. Or if you want to just add a loop around the original test to test with different users.

test/stac/stac.ts
```diff
@@ -160,4 +162,63 @@ describe('STAC catalog route', function () {
...
+              title: 'The current page',
+              type: 'application/json',
+            },
+          ]);
+          expect(catalog.stac_version).to.equal('1.0.0');
+          expect(catalog.title).to.include('Harmony output for ');
+          expect(catalog.type).to.equal('Catalog');
+        });
+      });
+      describe('when the linkType is invalid', function () {
```
chris-durbin, line 211: Similar comment - this part doesn't seem related to the user type permissions and looks like it is already tested elsewhere.

chris-durbin (approved):

indiejames (approved):

test/stac/stac-item.ts
ygliuvt: I didn't do the refactoring because I thought it was more readable this way. I will refactor the test as requested.

PR #391, opened 2023-05-12
HARMONY-1439: Moved service-example service to the end of the services in services.yml so that it will be picked last.

ygliuvt:
## Jira Issue ID
HARMONY-1439

## Description
Moved service-example service to the end of the services in services.yml so that it will be picked last.

## Local Test Steps
Submit the following request locally and verify the trajectory-subsetter service is picked to run for the request.

http://localhost:3000/C1242267295-EEDTEST/ogc-api-coverages/1.0.0/collections/%2FBEAM0000%2Fagbd/coverage/rangeset?forceAsync=true&granuleId=G1242274836-EEDTEST

## PR Acceptance Checklist
* [x] Acceptance criteria met


config/services.yml
```diff
@@ -1148,3 +1117,34 @@ https://cmr.uat.earthdata.nasa.gov:
...
+    data_operation_version: '0.17.0'
+    type:
+      <<: *default-turbo-config
+      params:
+        <<: *default-turbo-params
+        env:
+          <<: *default-turbo-env
+          STAGING_PATH: public/harmony/service-example
+    umm_s:
+      - S1237980031-EEDTEST
```
chris-durbin, line 1134: Based on Owen's comment that the service record is for netcdf-to-zarr we need to fix this to create a service-example service record and use that instead. We should also sanity check that no UMM-S record shows up in more than one service chain.

ygliuvt: Created new umm-s records for service-example in PROD and UAT. Verified no duplicate umm-s records in different service chains.

owenlittlejohns: I'll rebuild my local Harmony instance and run the affected regression test suites now.

owenlittlejohns: I re-ran the Trajectory Subsetter, Regridder and Swath Projector regression tests and they were sent along to the correct service now. Nice!

owenlittlejohns (approved):

chris-durbin (approved):

PR #392, opened 2023-05-16
HARMONY-1465: Update service runner to allow tests to use a subdirectory of /tmp

chris-durbin:
## Jira Issue ID
HARMONY-1465, HARMONY-1459, and HARMONY-1461

## Description
HARMONY-1465: Updated service runner to allow tests to use a subdirectory of /tmp - /tmp/harmony. I also updated the mocha configuration to allow a timeout of 5 seconds since some of the tests always take more than 3.5 seconds. Previously this 5 second timeout was only applied when using npm run test.

HARMONY-1461: Updated the harmony start task to allow 3GB of heap space for the node process.

HARMONY-1459: Updated the service metrics to always have one more pod than the current number of queued and in progress work items for a service

I built the image and deployed to my own sandbox environment to test.

## Local Test Steps
HARMONY-1465
1. Run `mocha` in the tasks/service-runner directory and verify everything passes.
2. Build the image with `npm run build`
3. Restart all of your services to use the new service-runner image with `bin/restart-services`
4. Start harmony
5. Submit a request and verify it completes successfully and the produced image is correct.

HARMONY-1461
I verified this change worked in production before. You can run `npm run start` and make sure that the process works fine. I'm not sure if there's a way to get the max allowed heap space information back about the process.

HARMONY-1465
You'll either need to deploy HPAs locally or to your own sandbox.
Submit a request and when the metrics are captured verify that the metrics reflect one more than the currently running pods.

For example with nothing running it used to report 0/1 (avg) for the TARGETS - now it will always report at least 1/1 (avg).

## PR Acceptance Checklist
* [X] Acceptance criteria met

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,7 +7,7 @@
     "test": "strict-npm-engines && eslint --ext .ts --ignore-pattern tasks . && nyc mocha && better-npm-audit audit && lerna run test",
     "test-fast": "TS_NODE_TRANSPILE_ONLY=true mocha",
     "coverage": "nyc mocha",
-    "start": "ts-node -r tsconfig-paths/register --require './app/tracing.ts' app/server.ts",
+    "start": "NODE_OPTIONS=--max-old-space-size=3096 ts-node -r tsconfig-paths/register --require './app/tracing.ts' app/server.ts",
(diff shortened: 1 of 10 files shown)


indiejames (approved):

ygliuvt (approved):
Run harmony locally and verified the new service runner image is used and works with an example request.

PR #395, opened 2023-05-17
HARMONY-1462: Fixed work reaper OOM error in PROD.

ygliuvt:
## Jira Issue ID
HARMONY-1462

## Description
Rewrite work reaper to do batching during record retrieving rather than just deletion in batches. Also changed the retrieval query to only get the field (only id) that is needed by work reaper rather than all fields.

## Local Test Steps
Update the work reaper related configurations (`WORK_REAPER_BATCH_SIZE`, `WORK_REAPER_PERIOD_SEC` and `WORK_REAPER_BATCH_SIZE` to some lower values to make the work reaper job run more easily monitor-able in local dev) and verify it is cleaning up terminal work items and workflow steps as desired.

The OOM error fix verification can be done by reviewing the code change that is outlined in the Description above.

## PR Acceptance Checklist
* [x] Acceptance criteria met

### DIFF
--- a/app/models/work-item.ts
+++ b/app/models/work-item.ts
@@ -399,6 +399,8 @@ export async function getWorkItemsByJobIdAndStepIndex(
  * @param notUpdatedForMinutes - jobs with updateAt older than notUpdatedForMinutes ago
  * will be joined with the returned work items
  * @param jobStatus - only jobs with this status will be joined
+ * @param startingId - the work item id to begin the query with, i.e. query work items with id greater than startingId
+ * @param batchSize - the batch size
  * @returns - all work item ids associated with the jobs that
  * met the updatedAt and status constraints
  */
@@ -406,13 +408,18 @@ export async function getWorkItemIdsByJobUpdateAgeAndStatus(
   tx: Transaction,
   notUpdatedForMinutes: number,
   jobStatus: JobStatus[],
+  startingId = 0,
+  batchSize = 2000,
 ): Promise<number[]> {
   const pastDate = subMinutes(new Date(), notUpdatedForMinutes);
   const workItemIds = (await tx(`${WorkItem.table} as w`)
     .innerJoin(Job.table, 'w.jobID', '=', `${Job.table}.jobID`)
-    .select(...tableFields)
+    .select(['w.id'])
     .where(`${Job.table}.updatedAt`, '<', pastDate)
-    .whereIn(`${Job.table}.status`, jobStatus))
+    .whereIn(`${Job.table}.status`, jobStatus)
+    .where('w.id', '>', startingId)
+    .orderBy('w.id', 'asc')
+    .limit(batchSize))
     .map((item) => item.id);
 
   return workItemIds;
@@ -427,6 +434,8 @@ export async function getWorkItemIdsByJobUpdateAgeAndStatus(
  * @param workItemStatuses - only WorkItems with these statuses will be retrieved
  * @param jobStatuses - only WorkItems associated with jobs with these statuses will be retrieved
  * @param fields - optional parameter to indicate which fields to retrieve - defaults to all
+ * @param startingId - the work item id to begin the query with, i.e. query work items with id greater than startingId
+ * @param batchSize - the batch size
  * @returns all WorkItems that meet the lastUpdateOlderThanMinutes and status constraints
 */
 export async function getWorkItemsByUpdateAgeAndStatus(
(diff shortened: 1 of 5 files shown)


app/models/work-item.ts
```diff
@@ -399,20 +399,27 @@ export async function getWorkItemsByJobIdAndStepIndex(
...
   notUpdatedForMinutes: number,
   jobStatus: JobStatus[],
+  startingId = 0,
+  batchSize = 2000,
 ): Promise<number[]> {
   const pastDate = subMinutes(new Date(), notUpdatedForMinutes);
   const workItemIds = (await tx(`${WorkItem.table} as w`)
     .innerJoin(Job.table, 'w.jobID', '=', `${Job.table}.jobID`)
-    .select(...tableFields)
+    .select(['w.id'])
```
ygliuvt, line 417: This line and the similar one in `workflow-step.ts` are the most important fixes in this PR. With these two lines of code change, we probably don't need the batching change at all.

app/models/work-item.ts
```diff
@@ -399,20 +399,27 @@ export async function getWorkItemsByJobIdAndStepIndex(
...
+ * @param startingId - the work item id to begin the query with, i.e. query work items with id greater than startingId
+ * @param batchSize - the batch size
  * @returns - all work item ids associated with the jobs that
  * met the updatedAt and status constraints
  */
 export async function getWorkItemIdsByJobUpdateAgeAndStatus(
   tx: Transaction,
   notUpdatedForMinutes: number,
   jobStatus: JobStatus[],
+  startingId = 0,
```
indiejames, line 411: It's slightly confusing that you don't actually include `startingId` in the results. I get the logic, just find the naming a little confusing, although I can't think of a better alternative.

app/workers/work-reaper.ts
```diff
@@ -25,48 +25,99 @@ export default class WorkReaper implements Worker {
...
+    while (!done) {
+      try {
+        const workItemIds = await getWorkItemIdsByJobUpdateAgeAndStatus(
+          db, notUpdatedForMinutes, jobStatus, startingId, batchSize,
+        );
+        if (workItemIds.length > 0) {
+          const numItemsDeleted = await deleteWorkItemsById(db, workItemIds);
+          totalDeleted += numItemsDeleted;
+          this.logger.info(`Work reaper removed ${numItemsDeleted} work items, starting id: ${startingId}.`);
+          startingId = Math.max(...workItemIds);
```
indiejames, line 51: not a big deal, but you don't need `Math.max` here since `workItemIds` is already sorted  - the max is always the last one.

app/workers/work-reaper.ts
```diff
@@ -25,48 +25,99 @@ export default class WorkReaper implements Worker {
...
+   * @returns Resolves when the request is complete
+   */
+  async deleteTerminalWorkItems(notUpdatedForMinutes: number, jobStatus: JobStatus[]): Promise<void> {
+    let done = false;
+    let startingId = 0;
+    let totalDeleted = 0;
+    const batchSize = env.workReaperBatchSize;
+    this.logger.info('Work reaper delete terminal work items started.');
+
+    while (!done) {
```
indiejames, line 42: I think it would be a bit cleaner to use something like
```
try {
  let workItemIds;
  while ((workItemIds = await getWorkItemIdsByJobUpdateAgeAndStatus(
          db, notUpdatedForMinutes, jobStatus, startingId, batchSize,
        )).length > 0) {
...
 }
} catch(...)
```

vinnyinverso (approved):

indiejames (approved):

app/workers/work-reaper.ts
```diff
@@ -25,48 +25,99 @@ export default class WorkReaper implements Worker {
...
+
+        if (workflowSteps.length < batchSize) {
+          done = true;
         }
-      } else {
-        this.logger.info('Work reaper did not find any workflow steps to delete');
+      } catch (e) {
+        this.logger.error('Error attempting to delete terminal workflow steps');
+        this.logger.error(e);
+        done = true;
```
chris-durbin, line 101: Can you move the `done = true` up above the logging - just to be sure that an exception with the logging for some reason doesn't cause things to run in a tight infinite loop. Same thing with the exception handling in the other while loop on line 62.

app/workers/work-reaper.ts
ygliuvt: This was my defensive coding kicked in. Agreed.

app/workers/work-reaper.ts
ygliuvt: I am using `done` to not do an extra query on the last batch when the number of retrieved records is less than the batch size. I find putting too much in the `while` condition not as readable as the `done` flag.

app/models/work-item.ts
ygliuvt: yeah, we had similar discussion on this on `work-failer`. Sorry that I have not found a better alternative and I think it is good to use the same naming convention as the existing code in `work-failer`.

PR #397, opened 2023-05-19
DAS-1604 - Add UAT service chain for HyBIG.

owenlittlejohns:
## Jira Issue ID

DAS-1604

## Description

This PR adds a service chain for the Harmony Browse Image Generator (HyBIG).

_Side note:_

I spent a while trying to get the tests to pass, because the new-ish `/docs` endpoint wasn't rendering correctly. It turns out that I now needed to explicitly specify whether the service chain has bounding box subsetting capabilities or not (I had initially just not said anything assuming that was the same as saying `bbox: false`). I'm not sure if that's documented anywhere, but it seems like a subtle change that might be worth calling out.

## Local Test Steps

* Pull this branch.
* Use the `bin/build-image` script in the HyBIG repo to build the service image.
* Update your `.env` file to include `hybig` as one of the `LOCALLY_DEPLOYED_SERVICES`.
* Rebuild your local Harmony instance.
* Run the request below. It should route to `sds/HyBIG` (as seen in the `/workflow-ui`). Note the request output isn't important, as the service itself is not fully implemented.

```
http://localhost:3000/C1256584478-EEDTEST/ogc-api-coverages/1.0.0/collections/all/coverage/rangeset?forceAsync=true&maxResults=1&format=image%2Fpng
```

## PR Acceptance Checklist
* [x] Acceptance criteria met


flamingbear (approved):
This looks good to me.  I was able to build and run the stack here and make a request to the hybig service running my branch and it returned a png, xml and pgw file that all look good.  Nice!

env-defaults
```diff
@@ -269,6 +269,13 @@ HARMONY_GDAL_ADAPTER_LIMITS_CPU=128m
 HARMONY_GDAL_ADAPTER_LIMITS_MEMORY=8Gi
 HARMONY_GDAL_ADAPTER_INVOCATION_ARGS='python -m gdal_subsetter'
 
+HYBIG_IMAGE=sds/harmony-browse-image-generator:latest
+HYBIG_REQUESTS_CPU=128m
+HYBIG_REQUESTS_MEMORY=128Mi
+HYBIG_LIMITS_CPU=128m
```
chris-durbin, line 275: Note to ourselves that I don't think LIMITS_CPU parameters are being used anywhere, but if they were we wouldn't want to limit services to 1/8 a VCPU.

chris-durbin (commented):
We should merge this in along with a change to harmony-ci-cd to deploy the service otherwise requests in UAT that get routed to this service will get stuck in the running state forever.

ygliuvt (approved):

PR #403, opened 2023-05-30
Harmony 1419

indiejames:
## Jira Issue ID
HARMONY-1419

## Description
Wanted to get this PR out before Chris does :)

This adds a k8s service to schedule work-items on the service queues. This is only for local development until harmony-ci-cd is changed as well. Additional changes will be needed to support harmony-in-a-box with queues, but for now queues are disabled by default, so this is not an issue.

## Local Test Steps
1. pull this branch (harmony-1419)
1. set USE_SERVICE_QUEUES=true and NODE_ENV=production in your .env
2. run `bin/restart-services`
3. verify that the scheduler is running using `kubectl get pods -n harmony`
4. run harmony
5. issue queries

## PR Acceptance Checklist
* [x] Acceptance criteria met


vinnyinverso:
Should the whole process fail if a particular queue does not exist for a given scheduler message?

indiejames:
> Should the whole process fail if a particular queue does not exist for a given scheduler message?

Good point. I'll change this to just log an error. Although we probably need some more forceful way to let us know that something is misconfigured.

vinnyinverso:
Tested successfully http://localhost:3000/C1233800302-EEDTEST/ogc-api-coverages/1.0.0/collections/all/coverage/rangeset?granuleId=G1233800343-EEDTEST&format=image%2Ftiff.

test/helpers/queue.ts
```diff
@@ -1,20 +1,74 @@
...
+import { Logger } from 'winston';
+import logger from '../../app/util/log';
+import { getWorkFromDatabase } from '../../app/backends/workflow-orchestration/work-item-polling';
+import { WorkItemQueueType } from '../../app/util/queue/queue';
 
 let serviceQueues;
+let typeQueues;
+
+/**
+ * Process the scheduler queue. This function is only used for tests since they won't have a
```
vinnyinverso, line 14: Maybe add a little more detail on what "Process the scheduler queue" entails?

indiejames: Done.

kubernetes-services/work-scheduler/.eslintrc.yml
```diff
@@ -0,0 +1,9 @@
+rules:
```
chris-durbin, line 1: Can you add a short README at the top level of `kubernetes-services` to explain what goes under the directory and what one needs to do to define a new k8s service?

kubernetes-services/work-scheduler/app/routers/router.ts
```diff
@@ -0,0 +1,19 @@
+import express, { NextFunction } from 'express';
+
+/**
+ * Set up routes for the service. These are used exclusively by Kubernetes.
+ * @returns Router configured with service routes.
+ */
+export default function router(): express.Router {
+  const result = express.Router();
+
+  result.get('/liveness', async (req, res, _next: NextFunction): Promise<void> => {
```
chris-durbin, line 10: nit - only documenting the type of `_next`, but not `req` or `res` - I'm sure this is copied off of something else though.

kubernetes-services/work-scheduler/app/util/k8s.ts
```diff
@@ -0,0 +1,21 @@
...
+/**
+ * Get the number of pods running for a service
+ * @param serviceId - The service ID for which to get the number of pods
+ * @param namespace - The namespace in which to look for pods
+ * @returns The number of pods running for the service
+ * @throws An error if the Kubernetes API call fails
+ **/
+export async function getPodsCountForService(serviceId: string, namespace = 'harmony'): Promise<number> {
+  // Get all pods in the namespace
+  const allPodsResponse = await k8sApi.listNamespacedPod(namespace);
```
chris-durbin, line 12: I'm curious how often this is called and what the performance of this function is - I think I'll want to look at that while testing - what do you think of adding a debug log line that logs the duration?

kubernetes-services/work-scheduler/bin/push-image
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
```
chris-durbin, line 1: I think @vinnyinverso added a script that works for either AWS CLI 1 or 2 - we should replace these two scripts with that one modified for this image. This would be another good file to add to the templates directory I mentioned in another comment.

kubernetes-services/work-scheduler/config/service-template.yaml
```diff
@@ -0,0 +1,48 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: harmony-work-scheduler
+  labels:
+    name: harmony
+spec:
+  replicas: 1
```
chris-durbin, line 8: Is there any reason we might want to run more than one? Or maybe this is only used locally so it's fine to hardcode and not use a variable here.

kubernetes-services/work-scheduler/env-defaults
```diff
@@ -0,0 +1,61 @@
...
+QUEUE_LONG_POLLING_WAIT_TIME_SEC=20
+
+# page size to use with CMR calls - only used for logging in this service
+CMR_MAX_PAGE_SIZE=2000
+
+# SQS queue used to request scheduling of work items for a service
+WORK_ITEM_SCHEDULER_QUEUE_URL=http://localstack:4566/queue/work-item-scheduler-queue.fifo
+SCHEDULER_QUEUE_BATCH_SIZE=10
+# Used to allow slightly more work items to be queued than the number of available workers
+# to avoid queue starvation
```
chris-durbin, line 44: "1.1 for example means to queue 10% more work items than there are running pods for the service."  Assuming that's what that means.

kubernetes-services/work-scheduler/test/helpers/env.ts
```diff
@@ -0,0 +1,3 @@
+process.env.SHARED_SECRET_KEY = '_THIS_IS_MY_32_CHARS_SECRET_KEY_';
```
chris-durbin, line 1: Shared secret key was for EDL token passing - why do we need it here?

kubernetes-services/work-scheduler/test/helpers/env.ts
```diff
@@ -0,0 +1,3 @@
...
+process.env.NODE_ENV = 'test';
+process.env.ARTIFACT_BUCKET = 'stac-catalogs';
```
chris-durbin, line 3: Why are we using the artifact bucket?

kubernetes-services/work-scheduler/env-defaults
```diff
@@ -0,0 +1,61 @@
+# Port the health check server listens on
+PORT=5000
+
+# Fake credentials for localstack
+AWS_ACCESS_KEY_ID=LOCALSTACK
+AWS_SECRET_ACCESS_KEY=LOCALSTACK
+
+AWS_DEFAULT_REGION=us-west-2
+
+NODE_ENV=production
```
chris-durbin, line 10: Normally we default things for running locally - why do we set it to production? Note to self that if this is just an indicator to use Postgres we should probably stop using NODE_ENV for that.

kubernetes-services/work-scheduler/test/helpers/hooks.ts
```diff
@@ -0,0 +1,11 @@
+import { before } from 'mocha';
```
chris-durbin, line 1: another potential templates/example file - or better yet - just define this in one place and in the `.mocharc.yml` use it directly: `- ../../../test/helpers/test-hooks.ts`

bin/create-k8s-config-maps-and-secrets
```diff
@@ -18,6 +18,8 @@ HARMONY_VERSION=$(grep version ./package.json \
 USER_AGENT="harmony/${HARMONY_VERSION} (${CLIENT_ID})"
 
 SERVICES_YML=$(< config/services.yml)
+# needed for k8s services running locally that access the dB
+DATABASE_URL=${DATABASE_URL/localhost/postgres}
```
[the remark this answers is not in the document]
indiejames: This is a string substitution, so as long as "localhost" does not appear in the DATABASE_URL it will not be affected. I don't think it would for any deployed environments.

bin/start-postgres-localstack
```diff
@@ -108,6 +108,9 @@ kubectl -n harmony create configmap localstack-config --from-literal=startup.py=
 
 kubectl apply -n harmony -f ./config/local-postgres-localstack-deployment.yml
 
+# run the database migrations for postgres
+NODE_ENV=production DATABASE_URL=postgresql://postgres:password@localhost:5432  npx knex --cwd db migrate:latest
```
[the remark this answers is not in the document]
indiejames: No, I still have to change HIAB to use postgres instead of sqlite, but I just have to do that before we make queues the default.

kubernetes-services/work-scheduler/app/server.ts
```diff
@@ -0,0 +1,34 @@
...
+import router from './routers/router';
+import { Server } from 'http';
+import Scheduler from './workers/scheduler';
+
+/**
+ *
+ * @param config - The configuration Record from the environment variables
+ * @returns An object containing the running components
+ */
+export default function start(_config: Record<string, string>): Server {
```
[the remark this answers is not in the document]
indiejames: Done.

kubernetes-services/work-scheduler/app/routers/router.ts
indiejames: Fixed

kubernetes-services/work-scheduler/.eslintrc.yml
indiejames: I'll add the README, but not the template directory. The README will basically say "copy the work-scheduler" directory and modify it, so the work-scheduler directory IS the template.

indiejames: Done.

kubernetes-services/work-scheduler/test/helpers/env.ts
indiejames: Because I copied this from the `service-runner` directory. Removed.

kubernetes-services/work-scheduler/env-defaults
indiejames: Yes, this is so the work-scheduler will default to using postgres. We actually default things to work with Harmony in a Box (HIAB), so this needs to use postgres by default since HIAB is not set up to share a file system between pods (and therefore can't use sqlite). Your point about not using NODE_ENV for this is a good one, though. Not sure if we want to change that now or later.

lerna.json
```diff
@@ -1,6 +1,7 @@
 {
   "packages": [
```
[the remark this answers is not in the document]
indiejames: I think github actions use lerna - not sure though. Doesn't it just run `npm run test` (which uses lerna)?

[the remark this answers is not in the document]
chris-durbin: Ok I checked the build for this PR and it looks like Github is good to go:

lerna info Executing command in 4 packages: "npm run test"
lerna info run Ran npm script 'test' in 'harmony-work-scheduler' in 42.2s:

> harmony-work-scheduler@1.0.0 test
> eslint --ext .ts . && nyc mocha --timeout 5000 && better-npm-audit audit

Although we shouldn't default the timeout to 5 seconds for this one - I'm guessing that was an artifact of copying from service-runner .mocharc.yml.

We will likely still need to change Bamboo for building and deploying the images since those are not done through `lerna`.

ygliuvt (approved):

ygliuvt (approved):

ygliuvt (approved):

kubernetes-services/work-scheduler/config/service-template.yaml
indiejames: this is only used locally

kubernetes-services/work-scheduler/test/helpers/env.ts
indiejames: we don't

indiejames: removed

lerna.json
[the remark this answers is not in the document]
indiejames: Removed the 5 second default timeout.

kubernetes-services/work-scheduler/test/helpers/hooks.ts
indiejames: Not going to add a separate template project - will use this project as the template.

kubernetes-services/work-scheduler/env-defaults
indiejames: Added your sentence to the comment for clarification.

kubernetes-services/work-scheduler/config/service-template.yaml
```diff
@@ -0,0 +1,48 @@
+apiVersion: apps/v1
```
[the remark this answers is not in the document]
indiejames: Not going to add a separate template project - will use this project as the template.

kubernetes-services/work-scheduler/app/util/env.ts
```diff
@@ -0,0 +1,100 @@
+import camelCase from 'lodash.camelcase';
```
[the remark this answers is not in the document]
indiejames: I think this should be worked separately as you mentioned. Along with your changes I think that ticket should look to pull common code into a a library (under the harmony repository) that has its own tests and can be used across the sub-projects, instead of the way the sub-projects get built now, which results in a weird file structure.

kubernetes-services/work-scheduler/app/util/k8s.ts
indiejames: Added timing/logging

kubernetes-services/work-scheduler/bin/push-image
indiejames: Changed to use Vinny's script.

bin/start-postgres-localstack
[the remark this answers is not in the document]
chris-durbin: Please sanity check HIAB with the default queues disabled before merging this in.

lerna.json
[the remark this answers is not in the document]
chris-durbin: Make sure to update the Bamboo deployments to build and deploy the images like we do with the tasks when you merge this in.

bin/start-postgres-localstack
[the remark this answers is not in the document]
indiejames: I removed this line for now since there is a race condition waiting for postgres to become ready to accept SQL commands. This line was just intended as a convenience and has nothing to do with queueing.

PR #405, opened 2023-06-01
HARMONY-1432: As a harmony service provider, I want to ensure there is exactly one UMM-S record for each harmony service chain.

ygliuvt:
## Jira Issue ID
HARMONY-1432

## Description
As a harmony service provider, I want to ensure there is exactly one UMM-S record for each harmony service chain so that it is simpler to configure harmony service related information

## Local Test Steps
1. Change a service chain configuration in services.yml to without `umm_s` configuration, verify harmony prints out a proper error message similar to `Error: There must be one and only one umm_s record configured as a string for harmony service: podaac/l2-subsetter` and harmony service does not start up.
2. Change a service chain configuration in services.yml to `umm_s` configured as a list, verify verify harmony prints out a proper error message similar to `Error: There must be one and only one umm_s record configured as a string for harmony service: podaac/l2-subsetter` and harmony service does not start up.

## PR Acceptance Checklist
* [x] Acceptance criteria met


test/models/index.ts
```diff
@@ -0,0 +1,26 @@
+import { expect } from 'chai';
+import { loadServiceConfigsFromFile, getServiceConfigs, validateServiceConfig } from '../../app/models/services';
+import _ from 'lodash';
+
+describe('Services.yml validation', function () {
+
+  describe('harmony services.yml configuration is valid', function () {
```
chris-durbin, line 7: This is such a great test that I wish we had added a long time ago. Can you change it to make sure it validates that all of the configuration for both Prod and UAT are correct?

It will require refactoring `loadServiceConfigs()` to take in the CMR environment URL as an argument and will need to return the service configurations from it, but then we'll be guaranteed that services.yml is correct in all environments.

test/resources/services_no_umm_s.yml
```diff
@@ -0,0 +1,955 @@
+# Order for each CMR endpoint in this file will reflect precedence of the service when
```
chris-durbin, line 1: Can you make both of these test files more targeted to only a couple of services and make the names for the services match what we're trying to test? Something like `Missing UMM-S` instead of gesdisc/giovanni. 

So maybe one bad service and one good one in each file and only one CMR environment.

app/models/services/index.ts
```diff
@@ -83,6 +93,18 @@ function validateServiceConfig(config: ServiceConfig<unknown>): void {
   }
 }
 
+/**
+ * Throws an error if the configuration is invalid. Logs a warning if configuration will be ignored.
+ * @param config - The service configuration to validate
+ */
+export function validateServiceConfig(config: ServiceConfig<unknown>): void {
+  if (config.umm_s === undefined || typeof config.umm_s !== 'string') {
+    throw new RequestValidationError(`There must be one and only one umm_s record configured as a string for harmony service: ${config.name}`);
```
chris-durbin, line 102: RequestValidationError is meant to be thrown from an HTTP call and return a 400 status code to a user along with a message.

I'm not sure if there's an existing exception more suited for this - maybe an InternalServerError is defined? If not you can just add a new Exception class in `errors.ts`.

chris-durbin (commented):
Changes look good - just a couple of requests to changes to the tests and a different exception class.

test/resources/services_no_umm_s.yml
ygliuvt: I thought about doing this, but decided not to because I wanted to demonstrate that we can pick out the wrong service among a list of good ones with realistic configurations. 

I can see test files with customized service name and less configuration provides more clarity though. So I am OK (not thrilled) with your approach. I will make the change.

app/models/services/index.ts
ygliuvt: I guess we could use the existing `ServerError`, but in this case, the server is not up, so the 500 status code will not be returned to the client.

chris-durbin (approved):
Tested this out locally and when the umm-s field is not configured correctly the server fails to start. I think this is better than just logging a warning message especially since we have the test that will fail the build if it is incorrect for either UAT or prod configurations.

vinnyinverso (approved):

PR #415, opened 2023-06-13
HARMONY-1396: As a harmony user, I want collection capabilities to include a link to the UMM-S record for any services connected to my collection via Harmony.

ygliuvt:
## Jira Issue ID
HARMONY-1396

## Description
As a harmony user, I want collection capabilities to include a link to the UMM-S record for any services connected to my collection via Harmony.

## Local Test Steps
Run harmony locally,

1) Verify `http://localhost:3000/capabilities?collectionId=C1233800302-EEDTEST` returns response with a capabilitiesVersion of '2' and services have a new `href` field with value of the umm-s CMR concept url.

2) Verify `http://localhost:3000/capabilities?collectionId=C1233800302-EEDTEST&version=2` returns the same result as above.

3) Verify `http://localhost:3000/capabilities?collectionId=C1233800302-EEDTEST&version=1` returns capabilitiesVersion of '1' and services without the `href` field.

## PR Acceptance Checklist
* [x] Acceptance criteria met

### DIFF
--- a/test/collection-capabilities.ts
+++ b/test/collection-capabilities.ts
@@ -1,6 +1,6 @@
 import _ from 'lodash';
-import { currentApiVersion } from '../app/frontends/capabilities';
 import { expect } from 'chai';
+import { currentApiVersion } from '../app/frontends/capabilities';
 import { hookGetCollectionCapabilities } from './helpers/capabilities';
 import hookServersStartStop from './helpers/servers';
 
--- a/test/collection-capabilities.ts
+++ b/test/collection-capabilities.ts
@@ -1,5 +1,6 @@
-import { expect } from 'chai';
+import _ from 'lodash';
 import { currentApiVersion } from '../app/frontends/capabilities';
+import { expect } from 'chai';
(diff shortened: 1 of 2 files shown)


vinnyinverso (approved):

test/collection-capabilities.ts
```diff
@@ -1,5 +1,6 @@
-import { expect } from 'chai';
+import _ from 'lodash';
 import { currentApiVersion } from '../app/frontends/capabilities';
+import { expect } from 'chai';
```
indiejames, line 3: nit: It looks like you are sorting the imports in alphabetical order (based on the thing imported, not the module name?). The loose convention we seem to have is to list npm modules at the top and our own modules below those. So 'chai' would come before '.../capabilities'. Not a big deal and I don't really care which convention we use, we should just be consistent.

indiejames (approved):
Tested locally

test/collection-capabilities.ts
ygliuvt: Good point. I will update the imports based on harmony convention.

PR #422, opened 2023-06-23
HARMONY-1426: Lock gdal-async back to 3.5.0

chris-durbin:
## Jira Issue ID
HARMONY-1426

## Description
I made some library updates and inadvertently updated the gdal version to a version that does not work on our EC2 instances. This moves it back to 3.5.0 which was the version that was working.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3797,7 +3797,8 @@
     "node_modules/@gar/promisify": {
       "version": "1.1.3",
       "resolved": "https://registry.npmjs.org/@gar/promisify/-/promisify-1.1.3.tgz",
-      "integrity": "sha512-k2Ty1JcVojjJFwrg/ThKi2ujJ7XNLYaFGNB/bWT9wGR+oSMJHMa5w+CUq6p/pVrKeNNgA7pCqEcjSnHVoqJQFw=="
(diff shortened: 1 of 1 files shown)


package-lock.json
```diff
@@ -12727,22 +12739,20 @@
       }
     },
     "node_modules/gdal-async": {
-      "version": "3.7.0",
-      "resolved": "https://registry.npmjs.org/gdal-async/-/gdal-async-3.7.0.tgz",
-      "integrity": "sha512-UeG7RPw3xDJa759QIcZ5LM0OhsIqqDZo39WSCYZfve53oC9wjjdUD+qmXauB9dpbDCYE+MGKO4QKCVDzX/FaDQ==",
+      "version": "3.5.0",
```
chris-durbin, line 12742: Here's the change back to 3.5.0

vinnyinverso (approved):

indiejames (approved):

package-lock.json
ygliuvt: This change should be made in package.json. We don't want to edit the package-lock.json directly.

ygliuvt (changes requested):
change package.json: `"gdal-async": "~3.5.0",`

PR #423, opened 2023-06-23
Harmony 900 - Determine whether my results are shareable via jobs endpoint

vinnyinverso:
## Jira Issue ID
HARMONY-900

## Description
This adds a boolean flag to the /jobs endpoint that indicates whether a job's results can be shared. In practice, this is accomplished by checking EULA tags and permissions in CMR.

The second acceptance criteria on this ticket is OBE due to a previous ticket that lifted sharing restrictions on the STAC endpoints.

You may notice that there is little change to the tests. This is because  this ticket simply adds a single field, the computation of which was already implemented. There are also tests that check for the existence of certain fields returned from /jobs. I added the new field to those checks.

## Local Test Steps

Submit two different requests. One with a collection that has no restrictions (no CMR EULA tags and guest read = true), and one request that does have the aforementioned restrictions. E.g.:

http://localhost:3000/C1233800302-EEDTEST/ogc-api-coverages/1.0.0/collections/all/coverage/rangeset?granuleId=G1233800343-EEDTEST&format=image%2Ftiff

http://localhost:3000/C1234724470-POCLOUD/ogc-api-coverages/1.0.0/collections/all/coverage/rangeset?subset=lon(-160%3A160)&subset=lat(-80%3A80)&maxResults=4&concatenate=true&skipPreview=true

Take a look at the /jobs endpoint. The second request's JSON should have `shareable: false`, and the first request's JSON should have `shareable: true`.

## PR Acceptance Checklist
* [x] Acceptance criteria met


app/util/env.ts
```diff
@@ -21,11 +21,13 @@ try {
 }
 
 let envOverrides = {};
-try {
-  envOverrides = dotenv.parse(fs.readFileSync('.env'));
-} catch (e) {
-  winston.warn('Could not parse environment overrides from .env file');
-  winston.warn(e.message);
+if (process.env.NODE_ENV !== 'test') {
```
vinnyinverso, line 24: This is unrelated to the current ticket, but should improve our tests by making sure that any custom .env values do not interfere.

chris-durbin (commented):
Tested out locally. Can you sanity check harmony-py is fine prior to merging? I think adding a new field to the job status endpoint should be fine, but am not 100% sure.

I think it would also be good if in `job-sharing.ts` we call the job status endpoint for each of the jobs and verify that the shareable value is what we expect for each of the jobs (pull back the job status as an admin so they can see the status for each one).

indiejames (approved):

vinnyinverso:
> Tested out locally. Can you sanity check harmony-py is fine prior to merging? I think adding a new field to the job status endpoint should be fine, but am not 100% sure.
> 
> I think it would also be good if in `job-sharing.ts` we call the job status endpoint for each of the jobs and verify that the shareable value is what we expect for each of the jobs (pull back the job status as an admin so they can see the status for each one).

Added those tests and sanity checked harmony-py.

chris-durbin (approved):

PR #425, opened 2023-06-30
Harmony 1492 - Set static position for important headers, links and buttons in workflow user interface pages

vinnyinverso:
## Jira Issue ID
harmony-1492

## Description
This sets a static position for important elements (e.g. pause + resume links, paging buttons, table headers) when the user is scrolling a large page in the workflow user interface. This prevents these elements from disappearing from the viewport. This change was applied to the jobs page and work items page.

Screenshot 2023-07-03 at 11 42 48 AM

## Local Test Steps
Submit a large number of job requests so that the jobs page of the workflow user interface can display lots of jobs at once. Make sure that at least one of these requests produces a lot of work items or granules so that you can also test the job details page with a large number of work item rows.
When you scroll, all important actions / links and table headers should be viewable in the jobs and work items pages.
- table header
- pause, cancel, resume, skip preview
- table filters
- paging buttons

## PR Acceptance Checklist
* [x] Acceptance criteria met


chris-durbin (approved):
Tested out successfully in Chrome, Firefox, and Safari. The only thing that I wasn't sure about was the paging buttons over top of the body of the table. I'm not sure if it might be better to have the body of the table scrollable instead. In any case this is definitely an improvement and works well.

One note I had is that for another ticket to display the currently viewed work items (100-199 out of 8,522) we wouldn't want that to appear next to the sticky paging buttons - probably want it in the Jobs breadcrumb header row.

vinnyinverso:
Closing in favor of https://github.com/nasa/harmony/pull/428

PR #427, opened 2023-07-03
Harmony 1493 - A page limit input in the Workflow UI

vinnyinverso:
## Jira Issue ID
HARMONY-1493

## Description
This ticket adds a page limit input to the workflow user interface on both of the jobs and job details page. It also sets the default page limit for the job details page to 100 work items (previously 1000).

Screenshot 2023-07-03 at 2 48 10 PM

## Local Test Steps
- set the page limit via the page limit input and click apply
- check that the appropriate number of records is returned
- also check that `limit=` is set properly in the query string of the URL

## PR Acceptance Checklist
* [x] Acceptance criteria met


app/frontends/workflow-ui.ts
```diff
@@ -295,7 +295,7 @@ export async function getJob(
   try {
     const isAdmin = await isAdminUser(req);
     const job = await getJobIfAllowed(jobID, req.user, isAdmin, req.accessToken, true);
-    const { page, limit } = getPagingParams(req, 1000);
+    const { page, limit } = getPagingParams(req, 100);
```
chris-durbin, line 298: I don't think we have to make these configuration parameters, but rather than using numbers directly can you define constants for them and then use those constants - e.g. `const defaultWorkItemPageSize = 100` just as a documentation thing.

test/workflow-ui/job.ts
```diff
@@ -75,6 +75,10 @@ describe('Workflow UI job route', function () {
           const listing = this.res.text;
           expect(listing).to.contain(mustache.render('<a href="/workflow-ui">Jobs</a>', {}));
         });
+        it('sets the page limit input to the expected value', function () {
+          const listing = this.res.text;
+          expect(listing).to.contain('<input name="limit" type="number" class="form-control" value="1">');
```
chris-durbin, line 80: We should add some tests with values larger than the max allowed size (which looks to be 2000 based on testing).

app/views/workflow-ui/job/index.mustache.html
```diff
@@ -70,6 +69,10 @@
                             negate statuses
                         </label>
                     </div>
+                    <div class="input-group mt-2">
+                      <span class="input-group-text">limit</span>
```
indiejames, line 73: I think "page limit" might be a bit clearer about what you are limiting here.

chris-durbin (changes requested):
There is one behavior change I'd like to see - on the jobs page if you pass in a max that's over 2000 it reloads the page and changes whatever your value was to 2000. I notice in the URL it still keeps whatever value you had set, but seems to still limit to 2000.

On the work items page though if you put a value in over 2000 you get a 400 error page with JSON telling you the max allowed value is 2000. I think it would be better if the behavior was like the jobs page and just changed your value to 2000.

One other thought is that an information "i" tooltip next to the limit that you could hover over or maybe click would be nice to tell users what the limit is for and that the max is 2000. I don't think we do tooltips elsewhere so probably a bigger ask and feel free to consider this one a new ticket enhancement request, not something we need to do prior to merging in this PR.

app/views/workflow-ui/job/index.mustache.html
indiejames: Actually, the word "limit" kind of implies you won't go past that page, so now I'm confused about the best choice of words here. Maybe "page size" ?

indiejames (approved):
Works well. Just had the one comment about using the word "limit". Not sure what the best choice is there, though.

app/views/workflow-ui/job/index.mustache.html
vinnyinverso: Yeah "page size" sounds great to me. I was trying to match the naming of the query parameter but definitely doesn't seem to be the best choice.

test/workflow-ui/job.ts
vinnyinverso: Added the test.

app/frontends/workflow-ui.ts
vinnyinverso: Added the const.

vinnyinverso: Also added "...changed your value to 2000." by passing in useMaxWhenExceeded=true to getPagingParams.

chris-durbin (approved):

ygliuvt (approved):

PR #434, opened 2023-07-14
HARMONY-900-2: Revert addition of shareable field to job JSON

vinnyinverso:
## Jira Issue ID
HARMONY-900

## Description
After discussing with Chris, we realize that the field that I added for https://github.com/nasa/harmony/pull/423 will have little future value now that we have access to a new EULA acceptance validation endpoint (EDL `verify_user_eula`) which will change how we check for shareability.

Context:
The future plan is to allow sharing based on user EULA acceptance and user data permissions rather than what we do now which is based on EULA presence and whether or not the collections are public. This should increase access to job results and sharing once the new logic is implemented. This flag will be a lot less meaningful once we implement those changes, hence the desire to remove it now before it appears in prod.

## Local Test Steps
- submit a request with > 1 granule and view the job endpoint JSON
- the shareable field should no longer show up on the jobs endpoint

## PR Acceptance Checklist
* [x] Acceptance criteria met

### DIFF
--- a/app/frontends/jobs.ts
+++ b/app/frontends/jobs.ts
@@ -188,7 +188,6 @@ export async function getJobStatus(
     const pagingLinks = getPagingLinks(req, pagination).map((link) => new JobLink(link));
     job.links = job.links.concat(pagingLinks);
     const jobForDisplay = getJobForDisplay(job, urlRoot, linkType, errors);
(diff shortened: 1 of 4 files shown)


ygliuvt (approved):
Tested locally.

chris-durbin (approved):

PR #438, opened 2023-07-19
HARMONY-1499: Added ability to use conditional step in services.yml based on native file format info in umm-c.

ygliuvt:
## Jira Issue ID
HARMONY-1499

## Description
Added ability to use conditional step in services.yml based on native file format info in umm-c. Here is the proposed conditional step that is requested by @owenlittlejohns :
```
- name: sds/HyBIG
    description: |
      The Harmony Browse Image Generator (HyBIG) supports the conversion of
      GeoTIFF inputs to Global Imagery Browse Services (GIBS) compatible PNG or
      JPEG outputs. This includes, where necessary, conversion to a GIBS
      supported Coordinate Reference System (CRS). Other projections can be
      requested, but will not be compatible with GIBS.
    data_operation_version: '0.17.0'
    type:
      <<: *default-turbo-config
      params:
        <<: *default-turbo-params
        env:
          <<: *default-turbo-env
          STAGING_PATH: public/sds/hybig
    umm_s: S1257776354-EEDTEST
    capabilities:
      subsetting:
        bbox: false
        shape: false
        temporal: false
      output_formats:
        - image/png
        - image/jpeg
      reprojection: true
    steps:
      - image: !Env ${QUERY_CMR_IMAGE}
      - image: !Env ${NET_2_COG_IMAGE}
        conditional:
          umm_c:
            native_format: ['netcdf-4']
      - image: !Env ${HYBIG_IMAGE}
```

The `umm_c` conditional with `native_format` compares with the value of the collection UMM-C field: `ArchiveAndDistributionInformationType.FileArchiveInformation.Format` when the sibling `FormatType = 'Native'`.

Since the NetCDF to COG service is still in development, we can't test the desired configuration in Harmony. But tests have been run with stubbed setup to prove that once the new service is available, the setup will work.
## Local Test Steps
Run Harmony locally and verify a regular harmony request works.
Manually set up a service chain with the new conditional umm_c native_format step (see description for an example) in services.yml.
For example, I set up this configuration with a conditional bad step for harmony/service-example service:
```
- name: harmony/service-example
    description: |
      Reference service that can be used to perform most operations supported by Harmony.
      Useful for testing new API features end-to-end on example data but not meant for production use.
    data_operation_version: '0.17.0'
    type:
      <<: *default-turbo-config
      params:
        <<: *default-turbo-params
        env:
          <<: *default-turbo-env
          STAGING_PATH: public/harmony/service-example
    umm_s: S1257851197-EEDTEST
    collections:
      - id: C1243747507-EEDTEST
        granule_limit: 100 # added to test collection granule limits for HARMONY-795
    capabilities:
      subsetting:
        bbox: true
        variable: true
        multiple_variable: true
      output_formats:
        - image/tiff
        - image/png
        - image/gif
      reprojection: true
    steps:
      - image: !Env ${QUERY_CMR_IMAGE}
      - image: !Env ${HARMONY_NETCDF_TO_ZARR_IMAGE}
        conditional:
          umm_c:
            native_format: ['netcdf-4']
      - image: !Env ${HARMONY_SERVICE_EXAMPLE_IMAGE}
```
Verify the conditional step is not scheduled when the collection's native-format is not `netcdf-4`, the request has two workflow steps and the request runs successfully.
Verify the conditional step is scheduled when the collection's native-format is `netcdf-4`, the request has three workflow steps and the request fails on the bad netcdf to zarr step. (I replaced line 16 in cmr-umm-collection-reader.ts with `req.operation.ummcollections = await getUmmCollectionsByIds(['C1239324659-POCUMULUS'], req.accessToken);` to fake the collection native_format to netcdf-4 for the test.)

You don't have to follow exactly what I did to create the fake condition to test this PR. You could add the umm_c native_format conditional step to a real service chain and with real collection if you want. 

## PR Acceptance Checklist
* [x] Acceptance criteria met


app/models/data-operation.ts
```diff
@@ -852,6 +852,24 @@ export default class DataOperation {
     this.model.stagingLocation = value;
   }
 
+  /**
```
chris-durbin, line 855: If you want to store the UMM collections to use later on, there are generally two ways:

1.) Add it to RequestContext
2.) Add it as a top level property of the Data operation class.

Only fields that get serialized should be added to the model. Also a nit - call it `ummCollections`.

app/models/services/base-service.ts
```diff
@@ -159,6 +162,16 @@ function stepRequired(step: ServiceStep, operation: DataOperation): boolean {
       required = true;
     }
   }
+  if (required && step.conditional?.umm_c) {
+    required = false;
```
chris-durbin, line 166: I think the default should be the step is required - so move this line to set it to false below the `if (step.conditional.umm_c.native_format)` check.

app/util/cmr.ts
```diff
@@ -254,6 +254,22 @@ export interface CmrPermissionsResponse extends CmrResponse {
   data: CmrPermissionsMap;
 }
 
+export interface CmrUmmCollection {
+  meta: {
+    // eslint-disable-next-line @typescript-eslint/naming-convention
+    'concept-id': string;
+  };
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  umm: any;
```
chris-durbin, line 263: It's useful to list out the fields that we use for documentation reasons. So this wouldn't be the full list of UMM-C fields, just the ones harmony reads.

So it would include something like `ArchiveAndDistributionInformation { FileArchiveInformation:  { FormatType: string; Format: string;}`

test/models/services.ts
```diff
@@ -1033,4 +1054,54 @@ describe('createWorkflowSteps', function () {
...
+          'Name': 'UMM-C',
+          'URL': 'https://cdn.earthdata.nasa.gov/umm/collection/v1.17.3',
+          'Version': '1.17.3',
+        },
+        'ArchiveAndDistributionInformation': {
+          'FileArchiveInformation': [ {
+            'Format': 'netCDF-4',
+            'FormatType': 'Native',
+          } ],
+          'FileDistributionInformation': [ {
```
chris-durbin, line 1076: Are we making use of this field? If not when you change the type for CmrUmmCollection you may need to drop this field in your test.

chris-durbin (changes requested):
Design and implementation looks good - only a couple of minor changes requested.

app/util/cmr.ts
```diff
@@ -255,6 +255,29 @@ export interface CmrPermissionsResponse extends CmrResponse {
   data: CmrPermissionsMap;
 }
 
+export interface CmrUmmCollection {
+  meta: {
+    // eslint-disable-next-line @typescript-eslint/naming-convention
+    'concept-id': string;
+  };
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
chris-durbin, line 263: Get rid of this directive since you changed to not use `any`

chris-durbin (approved):

indiejames (approved):

PR #444, opened 2023-08-04
Harmony 1503 - Add extend parameter

vinnyinverso:
## Jira Issue ID
HARMONY-1503

## PR Update
See https://github.com/nasa/harmony/pull/444#discussion_r1300471581
Test by issuing a request like http://localhost:3000/C1234724470-POCLOUD/ogc-api-coverages/1.0.0/collections/all/coverage/rangeset?subset=lon(-160%3A160)&subset=lat(-80%3A80)&maxResults=2&skipPreview=true&extend=dimA,dimB and making sure it doesn't get rejected.

## Description
From the ticket: "As a Tempo user, I want to be able to request data for one dimension be stitched together from multiple files into a single file." 

This pull request is the first step in supporting the aforementioned stitching by adding `extend` to the list of valid query parameters. When a user requests to extend a particular variable, this code validates that the variable exists (as far as CMR is concerned), and is one of the variables that was requested -- e.g. `/collections/blue_var/coverage/rangeset?extend=green_var` would be invalid since `blue_var` was requested, but `extend=green_var`.

The bulk of that logic is here: https://github.com/nasa/harmony/pull/444/files#diff-0190623dd051025f9e4603855805344ee376dfbe378fb80a3aaaf2e36725de0b

**Some Questions:**
- Whether the user should be able to pass in a comma separated list of dimensions for extend?

- Do we need another request option? See comment from Daniel K.:
>Not all variables would be indexed by this requested dimension, and in that case we will either have to decide what the default is, or have another request option that says what to do with those variables which aren't indexed by that dimension.

I also cleaned up some logging that I had used to build some dashboards. I realized that the approach is pretty brittle and hasn't been particularly useful for a while now, so I'm removing the code. 

## Local Test Steps
You can test this out with requests like:

should succeed:
 http://localhost:3000/C1233800302-EEDTEST/ogc-api-coverages/1.0.0/collections/all/coverage/rangeset?maxResults=2&extend=green_var

should fail:
http://localhost:3000/C1233800302-EEDTEST/ogc-api-coverages/1.0.0/collections/blue_var/coverage/rangeset?maxResults=2&extend=green_var

## PR Acceptance Checklist
* [x] Acceptance criteria met

### DIFF
--- a/app/backends/service-metrics.ts
+++ b/app/backends/service-metrics.ts
@@ -30,7 +30,7 @@ export async function getEligibleWorkItemCountForServiceID(
       workItemCount = await getQueuedOrRunningCountForService(tx, serviceID);
     });
     if (!workItemCount) workItemCount = 0;
-    logger.info('Got num_ready_work_items metric.');
+    logger.debug('Got num_ready_work_items metric.');
     const response = {
       availableWorkItems: workItemCount,
     };
--- a/app/backends/service-metrics.ts
+++ b/app/backends/service-metrics.ts
@@ -1,10 +1,8 @@
 import { Response, Request, NextFunction } from 'express';
-import { WorkItemMeta } from '../models/work-item-interface';
 import { getQueuedOrRunningCountForService } from '../models/user-work';
 import db from '../util/db';
 import logger from '../util/log';
 import { RequestValidationError } from '../util/errors';
-import { sanitizeImage } from '../util/string';
 
 /**
  * Express.js handler that returns the number of work items in the 'READY' or 'RUNNING' state for the given serviceID
@@ -32,9 +30,7 @@ export async function getEligibleWorkItemCountForServiceID(
       workItemCount = await getQueuedOrRunningCountForService(tx, serviceID);
     });
     if (!workItemCount) workItemCount = 0;
-    const itemMeta: WorkItemMeta = { workItemAmount: workItemCount, workItemService: sanitizeImage(serviceID),
-      workItemEvent: 'readyMetric' };
-    logger.info('Got num_ready_work_items metric.', itemMeta);
+    logger.info('Got num_ready_work_items metric.');
     const response = {
       availableWorkItems: workItemCount,
     };
--- a/app/frontends/ogc-coverages/get-coverage-rangeset.ts
+++ b/app/frontends/ogc-coverages/get-coverage-rangeset.ts
@@ -8,7 +8,7 @@ import env from '../../util/env';
 import { RequestValidationError } from '../../util/errors';
 import { keysToLowerCase } from '../../util/object';
 import { ParameterParseError } from '../../util/parameter-parsing-helpers';
-import { doesPathMatch, parseVariables } from '../../util/variables';
+import { parseVariables, validateExtend } from '../../util/variables';
 import { parsePointParam, parseSubsetParams, subsetParamsToBbox, subsetParamsToTemporal } from './util/subset-parameter-parsing';
 /**
  * Express middleware that responds to OGC API - Coverages coverage
@@ -90,17 +90,8 @@ export default function getCoverageRangeset(
       varInfo.variables, varInfo.coordinateVariables);
   }
   if (query.extend) {
-    let extendVarValid = false; 
-    for (const varInfo of varInfos) {
-      extendVarValid = varInfo.variables.some((v) => doesPathMatch(v, query.extend));
-      if (extendVarValid) break;
-    }
-    if (!extendVarValid) {
-      throw new RequestValidationError(`${query.extend} was not found in the requested variables`);
-    }
+    validateExtend(query.extend, req.params.collectionId, req.collections, varInfos);
   }
   req.operation = operation;
   next();
 }
-
-
--- a/app/frontends/ogc-coverages/get-coverage-rangeset.ts
+++ b/app/frontends/ogc-coverages/get-coverage-rangeset.ts
@@ -97,3 +97,4 @@ export default function getCoverageRangeset(
   next();
 }
 
+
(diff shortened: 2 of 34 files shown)


app/frontends/ogc-coverages/get-coverage-rangeset.ts
```diff
@@ -90,6 +90,9 @@ export default function getCoverageRangeset(
       varInfo.variables, varInfo.coordinateVariables);
   }
 
+  if (query.extend) {
+    validateExtend(query.extend, req.params.collectionId, req.collections, varInfos);
```
chris-durbin, line 94: Since we're only updating the OGC coverages API that means the parameter will not be supported from our WMS AMI - I think that's fine, just making note of it.

app/util/job.ts
```diff
@@ -74,11 +74,10 @@ export async function completeJob(
     job.updateStatus(finalStatus, message);
     await job.save(tx);
     if (failed) {
-      const numUpdated = await updateWorkItemStatusesByJobId(
+      await updateWorkItemStatusesByJobId(
         tx, job.jobID, [WorkItemStatus.READY, WorkItemStatus.RUNNING, WorkItemStatus.QUEUED], WorkItemStatus.CANCELED,
       );
-      const itemMeta: WorkItemMeta = { workItemStatus: WorkItemStatus.CANCELED, workItemAmount: numUpdated, workItemEvent: 'statusUpdate' };
-      logger.info(`Updated work items to ${WorkItemStatus.CANCELED} for completed job.`, itemMeta);
+      logger.info(`Updated work items to ${WorkItemStatus.CANCELED} for completed job.`);
```
chris-durbin, line 80: I would keep the number of work items updated in this log statement even if you don't want to keep the rest of the WorkItemMeta fields.

app/util/variables.ts
```diff
@@ -204,4 +204,33 @@ export function getVariablesForCollection(
       versionId: collection.version_id, variables, coordinateVariables });
   }
   return variableInfo;
+}
+
+/**
+ * Validate that the extend query parameter is both a valid variable name or concept ID
+ * and falls within the requested variable(s) (OGC collectionId) and collection(s).
+ * Assumes that the extend variable only needs to be found within one of the requested collections.
+ * @param extendParam - The name or concept ID of the variable to be extended (extend query parameter)
```
chris-durbin, line 213: I tested it out and we're expecting only a single value for the extend parameter - that means we can't allow extending across both lat and lon for example. Is that fine?

chris-durbin (approved):
I tested this out successfully locally. I see we have separate tickets for capabilities matching on `extend`, and updates to the data operation schema to pass in the extend information to backend services. We should continue to address feedback from the Tempo team as we work those tickets, but this should be fine to merge in prior to getting feedback.

app/backends/service-metrics.ts
```diff
@@ -32,9 +30,7 @@ export async function getEligibleWorkItemCountForServiceID(
       workItemCount = await getQueuedOrRunningCountForService(tx, serviceID);
     });
     if (!workItemCount) workItemCount = 0;
-    const itemMeta: WorkItemMeta = { workItemAmount: workItemCount, workItemService: sanitizeImage(serviceID),
-      workItemEvent: 'readyMetric' };
-    logger.info('Got num_ready_work_items metric.', itemMeta);
+    logger.info('Got num_ready_work_items metric.');
```
indiejames, line 33: Should this be 'debug' level?

indiejames (approved):
Looks good. Not sure if all the logging needs to be at the 'info' level, though.

ygliuvt:
We should add description of the `extend` parameter in the apis.md file.

ygliuvt (approved):
Approved assuming the new `extend` parameter is documented in the API doc.

indiejames:
> Approved assuming the new `extend` parameter is documented in the API doc.

I don't think we want to document this parameter yet as it doesn't do anything (yet).

ygliuvt:
> > Approved assuming the new `extend` parameter is documented in the API doc.
> 
> I don't think we want to document this parameter yet as it doesn't do anything (yet).

Right. I agree we can wait until it is fully supported to add the parameter in API doc.

app/util/variables.ts
owenlittlejohns: I think that's a use case that extend will likely want to support in the future.

app/util/variables.ts
```diff
@@ -204,4 +204,33 @@ export function getVariablesForCollection(
...
+ * @param collectionIdParam - The OGC collectionId
+ * @param eosdisCollections - An array of CMR collections
+ * @param varInfos - The variables requested by the user
+ * @throws RequestValidationError if the variable cannot be extended for the user's request
+ */
+export function validateExtend(extendParam: string, collectionIdParam: string, eosdisCollections: CmrCollection[], varInfos: VariableInfo[]): void {
+  if (collectionIdParam === 'all') {
+    for (const collection of eosdisCollections) { // Look through all of the variables
+      const foundVar = collection.variables.some((v) => doesPathMatch(v, extendParam));
```
owenlittlejohns, line 222: It might be worth specifically limiting the variables checked to those that are dimensions of others. The successful example in the ticket is a bit confusing - `green_var` isn't a dimension variable, it is a 3-D science variable, so I'm not sure it makes sense to be able to request to extend it.

Some possible options might be:

* Limiting to the `coordinateVariables`. Potential issues include: non 1-D coordinate variables (e.g., 2-D lon or lat variables in swath data). Whether all extendable dimensions have `VariableType = "COORDINATE"`.
* Check the length of the `Dimensions` attribute of the UMM-Var record. If it has 1 element, then it is 1-D, and dimension variables are 1-D. Downsides: some collections have 1-D science variables (e.g., trajectory data), and it relies on metadata curators populating the `dimensions` attribute for UMM-Var records.
* When all the variables are pulled from CMR, iterate through all of them and create a set of all the `Name` values of the elements of `Dimensions` for each variable. Downside: Again relies on well populated UMM-Var records.

I think the three options get more complicated as you go down the list, but I'm tempted to say that the last option is most likely to get the right set of variables that are dimensions.

vinnyinverso: Thanks for the feedback Owen. I'm going to get this merged ASAP after making the change to allow possibly multiple values for `extend`. We'll implement the validation checks after things have solidified a bit. That way we're not doing anything prematurely but we also won't be blocking anything on the service side of things.

ygliuvt (approved):

chris-durbin (approved):

PR #445, opened 2023-08-07
HARMONY-1548: Update Harmony service.yml gesdisc/giovanni service configuration to match Giovanni created UMM-S records

ygliuvt:
## Jira Issue ID
HARMONY-1548

## Description
Update Harmony service.yml gesdisc/giovanni service configuration to match Giovanni created UMM-S records

## Local Test Steps
Run `npm run compare-services` and verify no discrepancies are reported.

## PR Acceptance Checklist
* [x] Acceptance criteria met

### DIFF
--- a/config/services.yml
+++ b/config/services.yml
@@ -38,7 +38,7 @@ https://cmr.earthdata.nasa.gov:
         env:
           <<: *default-turbo-env
           STAGING_PATH: public/gesdisc/giovanni
-    umm_s: S2701049693-XYZ_PROV
+    umm_s: S2739607260-GES_DISC
     collections:
       - id: C1598621093-GES_DISC
         variables:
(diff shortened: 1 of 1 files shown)


chris-durbin (commented):
I think in the end we need to:
1. Add support for point subsetting capabilities matching.
2. Update UMM-S/services.yml comparison to consider point subsetting.
3. Figure out whether a point subset can also be turned into a bbox subset for other services.
4. Perform a successful ETE test with the Giovanni service through harmony.

Those may be performed as a couple of separate tickets. Rather than holding up changing over the UMM-S record, maybe for this ticket we only change the `umm_s` entry and remove `bbox` from the subsetting capabilities?

chris-durbin (approved):
Discussed and only change to this PR prior to merging should be to remove the `point: true` lines from services.yml.

We'll write two separate tickets to run an end to end test with Giovanni and add support for point subsetting capabilities matching to choose services.

vinnyinverso (approved):

PR #448, opened 2023-08-10
Harmony 1537 env vars

indiejames:
## Jira Issue ID
HARMONY-1537

## Description
Refactor env vars and add validation on startup to make sure the various environments are set up correctly / have all the env vars they need.

## Local Test Steps
1. check out this branch
2. Run `npm install`
3. Run `npm test`
4. Test Harmony in a Box by running `bin/bootstrap-harmony` and trying a query
5. Optionally do a sandbox deployment to and make sure things work

## PR Acceptance Checklist
* [x] Acceptance criteria met


test/helpers/stac-item.ts
```diff
@@ -12,12 +12,12 @@ import { it } from 'mocha';
 function expectAssets(assetsEntry: any, linkType?: string): void {
   let expectedType: string;
   switch (linkType) {
-    case 'https':
-    case 'http':
-      expectedType = 'http';
-      break;
-    default:
-      expectedType = 's3';
+  case 'https':
```
indiejames, line 15: This looks wrong to me, but the latest version of the linter insisted.

Dockerfile-notebooks.dockerignore
```diff
@@ -3,7 +3,7 @@ services
 .env
 db/*.sqlite3
 logs
-tasks
+#tasks
```
chris-durbin, line 6: For the notebooks image we don't need the tasks - can we put this back in?

app/backends/service-metrics.ts
```diff
@@ -4,7 +4,7 @@ import { getQueuedOrRunningCountForService } from '../models/user-work';
 import db from '../util/db';
 import logger from '../util/log';
 import { RequestValidationError } from '../util/errors';
-import { sanitizeImage } from '../util/string';
+import { sanitizeImage } from '@harmony/util/string';
```
chris-durbin, line 7: We purposely avoided this in the past - will relative paths no longer work?

env-defaults
```diff
@@ -39,59 +34,21 @@ HOST_BINDING='0.0.0.0'
...
-# For testing: Whether to use Localstack instead of AWS S3.  Options are "true" or "false".
-USE_LOCALSTACK=true
-
-# Hostname for harmony to use to connect to Localstack
-LOCALSTACK_HOST=localstack
-
-# Hostname for K8S pods to connect to Localstack
-LOCALSTACK_K8S_HOST=localstack
-
+# TODO - Can't find where this is used in the code - maybe just and env var used indirectly
```
chris-durbin, line 43: Address TODO

kubernetes-services/work-scheduler/package.json
```diff
@@ -11,8 +11,8 @@
     "test-fast": "TS_NODE_TRANSPILE_ONLY=true mocha",
     "lint": "eslint --ext .ts .",
     "coverage": "nyc mocha",
-    "build": "tsc && docker build -f Dockerfile.mac --tag harmonyservices/work-scheduler:${VERSION:-latest} .",
-    "build-m1": "tsc && docker build -f Dockerfile.mac --platform linux/amd64 --tag harmonyservices/work-scheduler:${VERSION:-latest} .",
+    "build": "tsc && cd ../.. && docker build -f kubernetes-services/work-scheduler/Dockerfile.mac --tag harmonyservices/work-scheduler:${VERSION:-latest} .",
```
chris-durbin, line 14: Won't this change your directory after running? If so can we put the`cd` in a subshell - ie `(cd ...)`?

service-env-defaults
```diff
@@ -0,0 +1,224 @@
+###########################################################################
```
chris-durbin, line 1: I think splitting out these from `env-defaults` makes it harder to deal with environment variables and harder for people to add new harmony services. I think my preference would be to just have a single `env-defaults` file at the top level - even for all the tasks and submodules we end up creating. You'd still define your expected environment for each task and module in an `env.ts` file but we wouldn't splinter the defaults all throughout the repo - just one place.

You still have the benefit of seeing which environment variables are used by which with the separate `env.ts` files, and we can still have separate modules to minimize the dependencies required for each new task/service we create.

tasks/giovanni-adapter/Dockerfile
```diff
@@ -1,9 +1,9 @@
 FROM node:16-alpine
 
 WORKDIR /app
-COPY built package.json package-lock.json ./
+COPY tasks/giovanni-adapter/built tasks/giovanni-adpater/package.json tasks/giovanni-adapter/package-lock.json ./
```
chris-durbin, line 4: This is really confusing that the Dockerfile doesn't work from the directory it is defined in. Instead of the `cd` stuff, can we use symlinks for anything that it needs from the top harmony level?

app/backends/service-metrics.ts
indiejames: This is not the same as path aliases that Patrick didn't want to use before. The '@' symbol here is not referring to a path alias, it is referring to a namespace. This is using a namespace (@harmony) to group together libraries from a provider (in this case `harmony`). So we don't have to give them names like `harmony_env`, `harmony_string`, etc. This is just a normal library import now - so no, we won't use relative imports for the libraries we create as separate modules now. There is an entry for it in `package.json` `dependencies` like any other library, only it uses a local file path with no version string.

kubernetes-services/work-scheduler/package.json
indiejames: `npm run` is already in a subshell, right?

tasks/giovanni-adapter/Dockerfile
indiejames: I don't like it either, but no, COPY will not follow symlinks (by design). There are some weird work-arounds using `tar` to pipe to the build command, but they seem even hackier to me.

service-env-defaults
indiejames: Reverted back to the main env-defaults

env-defaults
indiejames: Done.

Dockerfile-notebooks.dockerignore
indiejames: Done

package.json
```diff
@@ -150,10 +153,10 @@
     "tsconfig-paths": "^4.0.0",
     "tweetnacl": "^1.0.3",
     "tweetnacl-util": "^0.15.1",
-    "typescript": "^4.6.3",
+    "typescript": "^5.1.6",
```
ygliuvt, line 156: Typescript 5.0! Now I can use decorators.

tasks/giovanni-adapter/Dockerfile
indiejames: I changed this so that it now runs properly from the directory by doing most of the copying in a script before the Docker build runs

app/middleware/shapefile-converter.ts
```diff
@@ -30,6 +30,7 @@ async function _esriToGeoJson(filename: string): Promise<string> {
     const buffer = await fs.readFile(filename);
     const geojson = rewind(await shpjs.parseZip(buffer));
     await fs.writeFile(geoJsonFile.path, JSON.stringify(geojson), 'utf8');
+    await fs.writeFile('/tmp/test.geo.json', JSON.stringify(geojson, null, 2), 'utf8');
```
chris-durbin, line 33: Looks like debug code to remove?

app/util/cookies.ts
```diff
@@ -14,7 +14,7 @@ import HarmonyRequest from '../models/harmony-request';
 import { get } from 'lodash';
 import { randomBytes } from 'crypto';
 
-export const cookieOptions = { signed: true, sameSite: 'Lax' };
+export const cookieOptions = { signed: true, sameSite: 'lax' as const };
```
chris-durbin, line 17: What does this change do?

chris-durbin (approved):

app/middleware/shapefile-converter.ts
indiejames: Done.

app/middleware/shapefile-converter.ts
```diff
@@ -119,7 +119,7 @@ export default async function shapefileConverter(req, res, next: NextFunction):
         convertedFile = await converter.geoJsonConverter(originalFile, req.context.logger);
         operation.geojson = await store.uploadFile(convertedFile, `${url}.geojson`);
       } finally {
-        await fs.unlink(originalFile);
+        // await fs.unlink(originalFile);
```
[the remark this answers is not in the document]
indiejames: Done.

vinnyinverso (approved):

app/util/cookies.ts
indiejames: It forces 'lax' to be a specific type instead of just a string to match the declaration in the node module. TypeScript was complaining about it.

PR #449, opened 2023-08-28
HARMONY-1549: Fix to ensure service runner does not get stuck in an infinite loop thinking it needs to terminate

chris-durbin:
## Jira Issue ID
HARMONY-1549

## Description
Locally almost every time I run `bin/restart-services` I have some pods stuck in a state where they get restarted but just print out RECEIVED TERMINATION REQUEST forever. I changed things so that when a pod first sees the request to be terminated it cleans up that TERMINATING file allowing the restart to succeed. It will still not attempt to do any work when it sees the TERMINATING request, and the pre-stop hook will prevent a pod from being killed while it is actively WORKING on a request.

For my 50 granule request I ran `bin/restart-services` a dozen times and every granule was processed successfully and there were no retries as would happen if a pod was killed mid request.

## Local Test Steps

1. Build the service-runner image (npm run build in tasks/service-runner).
2. Run `bin/restart-services` to update all of your services to use the new image. 
3. Submit a request for 50 granules: http://localhost:3000/C1233800302-EEDTEST/ogc-api-coverages/1.0.0/collections/blue_var/coverage/rangeset?subset=lat(20%3A80)&subset=lon(-140%3A0)&outputCrs=EPSG%3A4326&format=image%2Fpng&ignoreErrors=true&maxResults=50
4. Run `bin/restart-services` while tailing logs for the service-example pod.
5. Verify that each time you run `bin/restart-services` the service-example pod gets terminated and a new one is started.
6. Verify that the new service-example pod always works.
7. Depending on when the service-example pod is terminated in its polling cycle you will see the last line logged:  `[warn]: Received TERMINATION request, no longer processing work`

## PR Acceptance Checklist
* [X] Acceptance criteria met

### DIFF
--- a/kubernetes-services/work-scheduler/package-lock.json
+++ b/kubernetes-services/work-scheduler/package-lock.json
@@ -28,7 +28,7 @@
         "pg": "^8.7.3",
         "pg-native": "^3.0.1",
         "tmp": "^0.2.1",
-        "winston": "^3.7.2",
+        "winston": "^3.10.0",
         "xstate": "^4.31.0"
       },
       "devDependencies": {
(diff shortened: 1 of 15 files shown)


vinnyinverso (approved):

ygliuvt (approved):

indiejames (approved):
Looks good

PR #454, opened 2023-09-06
HARMONY-1496&1498: Make job row and request more readable in Workflow UI; Add button for copying job request to clipboard

vinnyinverso:
## Jira Issue ID
HARMONY-1496/8

## Description
The goal of this PR is to make the job's request URL text and the entire job row (in the jobs page of the Workflow UI) more readable. It also adds a button for copying the request.

Below shows a comparison before/after. They're not the same jobs so it's not a perfect comparison but it should give you a good idea of what the change is meant to do.

Before:
Screenshot 2023-09-13 at 1 47 41 PM

After:
Screenshot 2023-09-13 at 1 47 53 PM

## Local Test Steps
Submit some jobs. Try resizing the page and copying a request URL. Try it out in multiple browsers if you are feeling adventurous.

## PR Acceptance Checklist
* [x] Acceptance criteria met


ygliuvt (approved):
Tested locally. Love the new layout and the copy feature!

app/frontends/workflow-ui.ts
```diff
@@ -138,11 +139,25 @@ function jobRenderingFunctions(logger: Logger, requestQuery: Record<string, any>
...
         const url = new URL(this.request);
-        const path = url.pathname + url.search;
-        return truncateString(path, 100);
+        let { pathname } = url;
+        if (pathname.indexOf('ogc-api-coverages') > -1) {
+          pathname = pathname.replace(`${ogcVersion}/collections`, '...');
+          pathname = pathname.replace('coverage/rangeset', '...');
+        }
+        const path = pathname + decodeURIComponent(url.search);
+        return truncateString(path, 315);
```
chris-durbin, line 160: I can't tell if this is the cause, but the copy functionality is broken for large requests. We should definitely truncate on display, but not on the copy to clipboard.

Note that there are issues with really large query strings where only a POST will work, but ideally they would get an error back about the URL being too long for a GET than an error about an id with '...' not found in CMR.

vinnyinverso: Great find. Made an update here: https://github.com/nasa/harmony/pull/454/commits/d2cffd6f151c63c3d2824f8c547723b5d897b7d0

vinnyinverso: I'm trying to tweak the wording so that people who see a warning about the truncation don't think that there is something wrong with their job.

vinnyinverso: If you can think of a better way to word the warning that let me know.

chris-durbin: Tested out successfully. I don't have a better recommendation, I usually go James for wordsmithing.

chris-durbin (approved):
Looks great.

PR #462, opened 2023-09-22
Harmony 945 - Improved error messaging for users

vinnyinverso:
## Jira Issue ID
HARMONY-945

## Description
- Removes application-specific details  like app logs and exception messages from user-facing job messages. This is in order to  comply with OSA.
- Adds failure details to job messages for some cases that were previously resulting in "unknown error"--see `getFinalStatusAndMessageForJob`.
- Also snuck in a small Workflow UI fix to ensure that when clicking the "copy request button" job URLs are encoded (see `jobs-table.mustache.html`).

I tried my best to keep messages generic but useful. Please let me know if you have any feedback/suggestions. I should check regression tests before merging this to make sure any tests are updated if necessary.

## Local Test Steps
1. There are a lot of different error scenarios to test. In general, you can throw an error in `runServiceFromPull` or `runQueryCmrFromPull`, e.g.:
```javascript
// eslint-disable-next-line @typescript-eslint/no-throw-literal
throw { response: { data : { description: 'error description' }, status: 500, statusText: 'server error' } };
```
2. Rebuild the work item updater
3. Rebuild the service runner (after any change like the one mentioned in step 1)
4. Redeploy/restart services and dev services
5. Submit a request with some combination of ignoreErrors=true/false
6. Inspect the job message in the Workflow UI once the job fails

## PR Acceptance Checklist
* [x] Acceptance criteria met

### DIFF
--- a/app/backends/workflow-orchestration/work-item-updates.ts
+++ b/app/backends/workflow-orchestration/work-item-updates.ts
@@ -6,7 +6,7 @@ import WorkflowStep, { decrementFutureWorkItemCount, getWorkflowStepByJobIdStepI
 import { Logger } from 'winston';
 import _, { ceil, range, sum } from 'lodash';
 import { JobStatus, Job } from '../../models/job';
-import JobError, { getErrorCountForJob } from '../../models/job-error';
+import JobError, { getErrorCountForJob, getNErrorsForJob } from '../../models/job-error';
 import JobLink, { getJobDataLinkCount } from '../../models/job-link';
 import { incrementReadyCount, deleteUserWorkForJob, incrementReadyAndDecrementRunningCounts, decrementRunningCount } from '../../models/user-work';
 import WorkItem, { maxSortIndexForJobService, workItemCountForStep, getWorkItemsByJobIdAndStepIndex, getWorkItemById, updateWorkItemStatus, getJobIdForWorkItem } from '../../models/work-item';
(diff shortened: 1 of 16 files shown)


app/backends/workflow-orchestration/work-item-updates.ts
```diff
@@ -161,11 +173,11 @@ async function handleFailedWorkItems(
   // If the response is an error then set the job status to 'failed'
   if (status === WorkItemStatus.FAILED) {
     continueProcessing = job.ignoreErrors;
-    if (!job.isComplete()) {
+    if (!job.hasTerminalStatus()) {
       let jobMessage;
 
       if (errorMessage) {
-        jobMessage = `WorkItem [${workItem.id}] failed with error: ${errorMessage}`;
```
vinnyinverso, line 168: I removed "with error" since a lot of our error messages already contain the word error or are evidently errors.

tasks/service-runner/app/service/service-runner.ts
```diff
@@ -285,28 +271,27 @@ export async function runServiceFromPull(workItem: WorkItemRecord, workItemLogge
               resolve({ batchCatalogs: catalogs });
             } else {
               clearTimeout(timeout);
-              const logErr = await _getErrorMessage(status, stdOut.aggregateLogStr, catalogDir, workItemLogger);
-              const errMsg = `${sanitizeImage(env.harmonyService)}: ${logErr}`;
+              const logErr = await _getErrorMessage(status, catalogDir, workItemLogger);
+              const errMsg = `${serviceName} - ${logErr}`;
```
vinnyinverso, line 275: Changed this to a `-` in between the message and the service name, since another `:` gets added to the error message upstream. (The message ends up looking pretty awkward with two colons.)

vinnyinverso: E.g.:
"WorkItem [3583761] failed with error: nasa/harmony-gdal-adapter:1.2.9: Incompatible variables"
will now look like:
"WorkItem [3583761] failed: nasa/harmony-gdal-adapter:1.2.9 - Incompatible variables"

abrahamcuenca (commented):
This looks good; the messages seem generic enough. I noticed some checks failed, though. Are these failures expected?

chris-durbin (approved):
Changes look ok to me. I did some testing and received only generic error messages back for my case, but I think that's because the service did not provide an error message back.

You will definitely want to run the HGA regression test notebook which tests error messages.

app/models/job-error.ts
```diff
@@ -75,6 +75,29 @@ export async function getErrorsForJob(
...
+/**
+ * Returns the first n errors for a given job
+ *
+ * @param tx - the transaction to use for querying
+ * @param jobID - the UUID associated with the job
+ * @param n - the limit for errors to retrieve
+ *
+ * @returns A promise that resolves to an array of job errors
+ */
+export async function getNErrorsForJob(
```
indiejames, line 87: I would probably just call this `getErrorsForJob` and make `n` an optional parameter that defaults to -1, meaning "get all the errors". Not necessary for now, though.

indiejames: Also, it would be good to validate `n`

indiejames (approved):

PR #472, opened 2023-10-04
Harmony 398/1491 - Bulk job status change

vinnyinverso:
## Jira Issue ID
HARMONY-398/1491

## Description
Adds bulk job status change endpoints (cancel, pause, etc.), and links to the Workflow UI (jobs page) to utilize the endpoints.
Screenshot 2023-10-04 at 4 04 04 PM

## Local Test Steps
Submit jobs and use the check boxes and links in the top right corner of the web page to change their statuses. If you have multiple EarthData Login users, you can test out cases where an admin cancels or modifies someone else's job.

## PR Acceptance Checklist
* [x] Acceptance criteria met


vinnyinverso:
Will write a ticket for getting the jobs page of the workflow UI to refresh automatically if not already done.

vinnyinverso:
Added HARMONY-1610 for the table refresh after status change.

ygliuvt (approved):
Tested successfully locally.

chris-durbin (changes requested):
Tested out successfully. I think we need one change to prevent potential mistakes since canceling a job is irreversible.

When a user selects to cancel we should provide a clear message to the user with the number of jobs they are canceling (potentially with the list of IDs), and in that dialog message give them option buttons to proceed or not. For consistency it probably makes sense to do that for pause as well even though pause is reversible.

chris-durbin (approved):
Tested everything out successfully. I ran into some minor issues with the pagination when I was applying filters that caused my jobs to disappear from the page (as they should). For example I ended up on a blank page of jobs that had first greyed out and only allowed clicking next and had the old number of total items listed (I canceled 10 out of 12 so there were only 2 left, but the navigation links were still acting like there were 12). So obviously clicking next resulted in a blank page, but then the navigation links were correct on that page and clicking back on the first page I saw the two remaining uncanceled jobs.

I don't think we have to address updating the pagination links as part of this PR, we can do that in a separate ticket.

vinnyinverso:
@chris-durbin I have a pagination bug ticket that I can add that to.

PR #481, opened 2023-10-20
Harmony 1482 - Paging fixes for Workflow UI

vinnyinverso:
## Jira Issue ID
HARMONY-1482

## Description
Fixes two things:
- "first" and "last" links are now usable when there is a first and last page to navigate to (they were previously disabled)
- pagination links are updated after batch status changes

---------------------

There is a known bug where you may end up on a page with no jobs when a new filter is applied or a batch action is taken against all of the jobs on the current page (this is due to the page always remaining the same). I've created a ticket to address that in the future.

## Local Test Steps
- When you're on the second to last page and the 2nd page of the jobs table of the Workflow UI, there should be a usable "last" and "first" paging link, respectively.
- After applying a filter like "running" and pausing a job from the jobs page of the Workflow UI, the paging links should update as expected (showing one less job in the counts, e.g. "1-3 of 3 => 1-2 of 2").

## PR Acceptance Checklist
* [x] Acceptance criteria met


indiejames (approved):
Tested locally

chris-durbin (approved):

PR #489, opened 2023-10-27
HARMONY-1617: Updates to work-failer to process database updates instead of using the message queue

chris-durbin:
## Jira Issue ID
HARMONY-1617

## Description
This changes the work-failer and the workflow-ui to no longer put messages on the work item update queue for the work-item-updater to process retries. Instead the database updates will be directly performed by the components. This is to address an issue we were seeing in production with the message queue getting overwhelmed by retry messages.

## Local Test Steps

Build the work-failer image.

Set FAILABLE_WORK_AGE_MINUTES to 0 and WORK_FAILER_PERIOD_SEC to 2 so that it constantly tries to fail work items. Verify that the retries are processed in the work-failer by looking at the logs.

Also test retry from the workflow-ui and make sure the work item is moved back to the READY state and retry count incremented.

I tested this in the sandbox and found that when submitting a large request it identified 160 items to be retried less than 5 minutes into the request. I figured out this is because things get queued based on the number of expected pods, but the pods are in the pending state for a little while before they get to running. The items are then marked as needing retry within one minute of being queued even though they haven't been worked on.

To account for this I changed our defaults from 1 minute being considered a failure to 5 minutes. I also changed the work failer to run every 5 minutes instead of 2 minutes with the end result being we may retry something every 5-10 minutes instead of 1-3 minutes.

## PR Acceptance Checklist
* [X] Acceptance criteria met


indiejames (approved):
Tested locally. Had to remove the validation on FAILABLE_WORK_AGE_MINUTES and change the code to the threshold check to force the work-failer to retry items, but I was able to see from the logs that it was doing the work itself instead of queueing. Also verified that retry from the workflow-ui moved items into the 'ready' state and updated the retry count.

ygliuvt (approved):

PR #492, opened 2023-10-31
Harmony 535 - Add documentation regarding exception handling in backend services

vinnyinverso:
## Jira Issue ID
HARMONY-535

## Description
Add documentation regarding exception handling in backend services.

After chatting with James we decided to remove the guidance on developing services without the service library. The rationale is that we want to push people toward using the service library and address needs that fall outside of this on a case-by-case basis.  We also have not encountered any use cases that fall in this realm so keeping that part of the documentation up to date seems like unnecessary toil. I could be convinced otherwise but I couldn't see a reason to keep it given how untested that case is.

## Local Test Steps
n/a

## PR Acceptance Checklist
* [ ] Acceptance criteria met


indiejames (approved):
Looks good to me.

chris-durbin (approved):
Nice job - I've been afraid to delete Appendix A for some reason - thanks for finally getting rid of it!

PR #500, opened 2023-11-08
Harmony 1469 - Add concatenation to UMM-S / services.yml comparison script

vinnyinverso:
## Jira Issue ID
HARMONY-1469

## Description
Adds concatenation to UMM-S / services.yml comparison script. Errors are returned from the script output when there is a mismatch between the concatenation property of the Harmony service config and the UMM-S record.

## Local Test Steps
`npm run compare-services`
In my case I also console.logged() the UMM-S record and service config when there is an error to see if the error reported is accurate.

## PR Acceptance Checklist
* [x] Acceptance criteria met

### DIFF
--- a/scripts/service-comparison.ts
+++ b/scripts/service-comparison.ts
@@ -71,18 +71,18 @@ function validateVariableSubsetting(
  * @returns validation failure message or '' if validation succeeds
  */
 function validateConcatenation(
-  _ummRecord: CmrUmmService, harmonyConfig: ServiceConfig<unknown>,
+  ummRecord: CmrUmmService, harmonyConfig: ServiceConfig<unknown>,
 ): string {
   const errors = [];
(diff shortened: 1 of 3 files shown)


chris-durbin (approved):
Looks good.

One thing that I was thinking is the harmony configuration should probably be concatenate_by_default for any of the zarr service chains. So the service providers might want to update these two UAT services to set default to true:

Validation failures for harmony/podaac-l2-subsetter-netcdf-to-zarr and S1239265626-EEDTEST:
    - Concatenation mismatch: harmony is true and UMM-S is false.
Validation failures for harmony/swot-repr-netcdf-to-zarr and S1240999847-EEDTEST:
    - Concatenation mismatch: harmony is true and UMM-S is false.

indiejames (approved):

ygliuvt (approved):

PR #509, opened 2023-11-30
Harmony 1642

indiejames:
## Jira Issue ID
HARMONY-1642

## Description
Add indices to dB tables to support more efficient querying when processing work-item updates, as well as an extra query to allow us to usually skip a more expensive query when doing work-item updates.

## Local Test Steps
1. `export VERSION=<some version\>
2. build the work-updater image using `npm run build-m1` in the `services/work-updater` directory
3. push the image to the sandbox using `bin/push-images $VERSION`
4. set WORK_ITEM_UPDATER_IMAGE=266454154769.dkr.ecr.us-west-2.amazonaws.com/harmonyservices/work-updater:<your version> in the .env file in your harmony-ci-cd repo
5. do a sandbox deployment
6. try some queries

## PR Acceptance Checklist
* [x] Acceptance criteria met

### DIFF
--- a/services/harmony/app/backends/workflow-orchestration/work-item-updates.ts
+++ b/services/harmony/app/backends/workflow-orchestration/work-item-updates.ts
@@ -427,9 +427,14 @@ async function maybeQueueQueryCmrWorkItem(
  *
  * @param logger - the logger to use for logging timing statements
  * @param tx - the database transaction to use for making queries
- * @param workItem - the work-item o
+ * @param workItem - the work-item associated with the step
+ * @param thisStep - the step to check
+ *
+ * @returns true if the step is complete
  */
 async function isStepComplete(logger: Logger, tx: Transaction, workItem: WorkItem, thisStep: WorkflowStep): Promise<boolean> {
+  // this first check is much faster than the second, but only conclusive if notComplete is true.
+  // if it is false we must do the second (slower check)
   let notComplete = await (await logAsyncExecutionTime(
     isDefinitelyNotComplete,
     'HWIJUWJI.isDefinitelyNotComplete',


services/harmony/app/backends/workflow-orchestration/work-item-updates.ts
```diff
@@ -421,6 +421,35 @@ async function maybeQueueQueryCmrWorkItem(
...
+ */
+async function isStepComplete(logger: Logger, tx: Transaction, workItem: WorkItem, thisStep: WorkflowStep): Promise<boolean> {
+  let notComplete = await (await logAsyncExecutionTime(
+    isDefinitelyNotComplete,
+    'HWIJUWJI.isDefinitelyNotComplete',
+    logger))(
+    tx, workItem.jobID, workItem.workflowStepIndex,
+  );
+
+  if (!notComplete) {
```
chris-durbin, line 440: if we negated the other function this would be `if (possiblyComplete)` - I can understand `!notComplete` so think it's good to leave it this way

services/harmony/app/backends/workflow-orchestration/work-item-updates.ts
```diff
@@ -421,6 +421,35 @@ async function maybeQueueQueryCmrWorkItem(
   }
 }
 
+
+/**
+ * Check to see if a workflow step for the given job is completed
+ *
+ * @param logger - the logger to use for logging timing statements
+ * @param tx - the database transaction to use for making queries
+ * @param workItem - the work-item o
```
chris-durbin, line 430: ChatGPT wasn't finished writing your docstring - I have no idea what word might start with 'o' that fits  and need @param thisStep

services/harmony/app/backends/workflow-orchestration/work-item-updates.ts
```diff
@@ -421,6 +421,35 @@ async function maybeQueueQueryCmrWorkItem(
...
+ */
+async function isStepComplete(logger: Logger, tx: Transaction, workItem: WorkItem, thisStep: WorkflowStep): Promise<boolean> {
+  let notComplete = await (await logAsyncExecutionTime(
```
chris-durbin, line 433: It's probably worth mentioning why we are doing this (either as a comment or in the docstring) - it's because this first check is much faster, but there are cases where it won't find any incomplete work items, but we're not done with the step. So we do the fast check, then if it thinks we might be done, we do the guaranteed check.

chris-durbin (approved):
Reviewed the logs from James' sandbox run with a couple of million granule queries and verified that processing was keeping up and no slow queries were logged for checking if a workflow step was complete like we were seeing in production.

services/harmony/app/backends/workflow-orchestration/work-item-updates.ts
indiejames: fixed

services/harmony/app/backends/workflow-orchestration/work-item-updates.ts
indiejames: Added additional clarifying comment

services/harmony/app/backends/workflow-orchestration/work-item-updates.ts
indiejames: ok

PR #511, opened 2023-12-07
HARMONY-1651: Queue fewer items to ensure faster processing of new requests and better fair queueing

chris-durbin:
## Jira Issue ID
HARMONY-1651

## Description
Queues fewer items to resolve an issue where synchronous requests were timing out when there were other requests in the system for many granules for the same service and the processing of those granules was extremely slow.

I also sped up the docker build for the work-scheduler (we need to do this for all the images) by reordering the npm ci command so that it happens before installing the code. So when there are only code changes the build and pushing the image to sandbox take seconds due to the layer caching.

## Local Test Steps
This is best tested in a sandbox environment since you'll have more than one worker pod and multiple schedulers. I tested in sandbox and saw that with 125 workers anywhere from 109 to 161 items were in the queued or running state at any point in time. That means we were fairly balanced. We were not allowing too many workers to be idle due to queue starvation (lowering overall system throughput) and we were not overqueueing (hindering the response time of synchronous requests while processing the large request). I submitted single granule requests throughout the run and they all completed in under 30 seconds.

## PR Acceptance Checklist
* [X] Acceptance criteria met

### DIFF
--- a/services/work-scheduler/app/util/k8s.ts
+++ b/services/work-scheduler/app/util/k8s.ts
@@ -21,5 +21,28 @@ export async function getPodsCountForService(serviceId: string, namespace = 'har
     });
   });
 
-  return pods.length;
+  const runningPods = pods.filter((pod) => pod.status.phase === 'Running');
+  return runningPods.length;
+}
+
+/**
+ * Get the number of pods running for a given pod name
+ * @param podName - The pod name for which to get the number of pods
+ * @param namespace - The namespace in which to look for pods
+ * @returns The number of pods running for the service
+ **/
+export async function getPodsCountForPodName(podName: string, namespace = 'harmony'): Promise<number> {
+  try {
+    const labelSelector = `name=${podName}`;
+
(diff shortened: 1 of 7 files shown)


services/work-scheduler/app/util/k8s.ts
```diff
@@ -21,5 +21,28 @@ export async function getPodsCountForService(serviceId: string, namespace = 'har
...
+
+    const res = await k8sApi.listNamespacedPod(namespace, undefined, undefined, undefined, undefined, labelSelector);
+    const pods = res.body.items;
+    const runningPods = pods.filter(pod => pod.status.phase === 'Running');
+
+    return runningPods.length;
+  } catch (e) {
+    logger.error(`Error getting the number of running ${podName} pods`);
+    logger.error(e);
+    return 1;
```
indiejames, line 46: `return 1`?

indiejames (approved):

services/work-scheduler/app/util/k8s.ts
chris-durbin: Sanity checked 1 scheduler is what I want in this case, also added one more test for when 0 schedulers are reported as running.

PR #512, opened 2023-12-08
HARMONY-1618: Speed up building all images when only a code change is being made.

chris-durbin:
## Jira Issue ID
HARMONY-1618

## Description
Building all docker images takes only seconds when making a code change due to caching of layers. Previously on every code change we were running `npm ci` or `npm install` commands due to the ordering in the Dockerfile.

## Local Test Steps
1. Build all the images with `npm run build`, `npm run build-notebooks-image`, `npm run build-all`, and `npm run build-all-m1` to make sure all of the images are working as expected.
2. Test out harmony in a box to verify the harmony container
3. Run `npm run run-notebooks-image` to test out the container for running notebooks

I tested a deployment to sandbox and verified everything was working.

## PR Acceptance Checklist
* [X] Acceptance criteria met

### DIFF
--- a/Dockerfile-notebooks
+++ b/Dockerfile-notebooks
@@ -6,12 +6,13 @@ RUN pip install jupyter
 RUN pip install rasterio OWSLib GDAL matplotlib netCDF4 numpy
 RUN pip install s3fs zarr
 
-RUN mkdir -p notebooks
-RUN mkdir -p notebooks/tmp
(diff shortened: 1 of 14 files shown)


indiejames (approved):
Tested locally. Verified that adding a dependency to a package.json file would prevent the cached layer from being used so that the new dependency was included in the new image. The cached layer is used if the dependencies have not changed.

PR #515, opened 2023-12-18
Harmony 1653 - Improve env logic

vinnyinverso:
## Jira Issue ID
HARMONY-1653

## Description
This is an effort to build upon some of the env refactoring that was done. 

## Local Test Steps
1. run `bin/build-service-images`
2. make a Harmony request

## PR Acceptance Checklist
* [x] Acceptance criteria met


bin/build-service-images
```diff
@@ -47,12 +47,13 @@ function build_all {
   parallel=$1
   pids=()
   for dir in services/*; do
+    [[ $dir == "services/README.md" ]] && continue
```
indiejames, line 50: Might be better to use `if [ -d "$dir" ]` here to only process `dir` if it actually is a directory. That way if we add another file besides README.md at some point things won't break.

packages/util/env-defaults
```diff
@@ -43,7 +43,7 @@ USE_LOCALSTACK=true
 LOCALSTACK_HOST=localstack
 
 # Identifier so backends know which Harmony client submitted the request
-CLIENT_ID=harmony-local
+CLIENT_ID=harmony-unknown
```
chris-durbin, line 46: It might make sense to set `CLIENT_ID=harmony-in-a-box` in the create-dotenv script.

packages/util/env.ts
```diff
@@ -30,64 +30,127 @@ if (Object.prototype.hasOwnProperty.call(process.env, 'GDAL_DATA')) {
...
+        }
+      } catch (e) {
+        logger.error(`Could not parse value ${value} for ${k} as JSON`);
+      }
+    }
   }
+  return serviceQueueUrls;
 }
 
-export interface IHarmonyEnv {
```
chris-durbin, line 46: Does dropping these definitions mean we do not have type information for the environment variables or a list of the valid variables that can be used from the `env` object? I think it's ok and we can still add validations, but just sanity checking this is desired.

chris-durbin (commented):
What are some things that we can test that were broken before, but are now working?

packages/util/env.ts
vinnyinverso: You'll still get both.

vinnyinverso:
> What are some things that we can test that were broken before, but are now working?

The one thing that was broken got fixed already.

chris-durbin (approved):
Successfully verified harmony in a box and a sandbox environment.

indiejames (approved):

PR #523, opened 2024-01-16
HARMONY-1564: Test Giovanni adapter service in github workflow.

ygliuvt:
## Jira Issue ID
HARMONY-1564

## Description
Test Giovanni adapter service in github workflow.

## Local Test Steps
Verify Giovanni adapter service request is tested in `harmony-in-a-box` github workflow.

## PR Acceptance Checklist
* [x] Acceptance criteria met


.github/workflows/harmony-in-a-box.yml
```diff
@@ -23,24 +23,38 @@ jobs:
...
+        node-version: 18
+        cache: 'npm'
+        cache-dependency-path: '**/package-lock.json'
 
-    - name: Pull down basic harmony docker images
-      run: bin/pull-harmony-images
+    - name: Create netrc
+      run: bin/create-netrc
+      env:
+        EDL_USER: ${{ secrets.LEO_EDL_USER }}
```
chris-durbin, line 41: I was thinking we have EDL_USER and EDL_PASSWORD secrets already - but not sure where you check that.

chris-durbin (approved):

.github/workflows/harmony-in-a-box.yml
ygliuvt: I kept running into harmony request failures and that made me think if it is because something is messed up with Patrick's EDL credential either with EDL login or access to the collection. In order to control/test it better, I changed it to use my credential and it actually helped.

.github/workflows/harmony-in-a-box.yml
```diff
@@ -84,5 +107,60 @@ jobs:
...
+    - name: Check Giovanni request final status
+      run: |
+        for i in {1..30}; do
+          request_status==$(curl -Ln -bj "${{ steps.giovanni.outputs.status_url }}" | jq -r '.status')
+          # remove the leading = that shows up only in github actions
+          GIOVANNI_STATUS="${request_status#=}"
+          if [[ "$GIOVANNI_STATUS" != "running" ]]; then
+            break
+          else
+            echo "Giovanni reqeust running, wait..."
```
indiejames, line 155: nit: 'request' typo

indiejames (approved):
I can't tell from the code, but does the Giovanni test require accessing Giovanni? Doesn't seem like we want to hit there service as part of our test.

ygliuvt:
Looks like we only return a link to the real Giovanni service in the request status, not directly invoke the service.
> I can't tell from the code, but does the Giovanni test require accessing Giovanni? Doesn't seem like we want to hit there service as part of our test.

PR #532, opened 2024-02-07
Harmony 1677 - Set up EDL group for users allowed to deploy new service versions

vinnyinverso:
## Jira Issue ID
HARMONY-1677

## Description
Sets up an EDL group for users allowed to deploy new service versions. Added code to check for membership in the new group as well as unit tests.

## Local Test Steps
n/a

## PR Acceptance Checklist
* [x] Acceptance criteria met


services/harmony/fixtures/uat.urs.earthdata.nasa.gov-443/log-viewer-eve-groups
```diff
@@ -23,4 +23,4 @@ X-Runtime: 0.011620
 Strict-Transport-Security: max-age=31536000
 Transfer-Encoding: chunked
 
-{"user_groups":[{"group_id":"0cf2a427-96cc-453e-ae44-e28dd0958738","name":"Harmony Log Viewers","tag":null,"shared_user_group":false,"created_by":"harmony_uat","app_uid":"harmony_uat","client_id":"abcd"}]}
```
chris-durbin, line 26: I would recommend adding at least one new user (and fixture). We should have one that is only service-deployer and one that is log-viewer-and-service-deployer.

The name of the fixture should explain exactly what it is - for example `log-viewer-eve-groups` - Eve should not also be in the service-deployers group - if she is then we should name the file `log-viewer-and-service-deployer-eve-groups`

services/harmony/test/util/edl-api.ts
```diff
@@ -0,0 +1,34 @@
...
+      stubEdlRequest(
+        '/oauth/token',
+        { grant_type: 'client_credentials' },
+        token({ accessToken: 'fake_access' }),
+      );
+    });
+    after(function () {
+      unstubEdlRequest();
+    });
+    describe('when the user is not part of the service deployers group', function () {
```
chris-durbin, line 21: I'd sanity check a few tests:
1.) only in service-deployer group
2.) in service-deployer and another group
3.) not in any group

You have 2 and 3 covered it looks like, so I'd probably add one more.

chris-durbin (approved):
Looks good - I just have comments about naming of fixtures and possibly adding one more test.

services/harmony/test/util/edl-api.ts
```diff
@@ -0,0 +1,34 @@
...
+      it('returns isServiceDeployer:false', async function () {
+        const groups = await getEdlGroupInformation('joe', logger);
```
indiejames, line 23: I think it would be good to use more descriptive user names here. Instead of 'joe' and 'eve', maybe 'non-service-deployer' and 'service-deployer', etc.

indiejames (approved):
Looks good. Just a minor comment about readability.

ygliuvt (approved):

services/harmony/test/util/edl-api.ts
chris-durbin: Probably out of scope for this ticket (there are a lot of these tests that rely on users in groups). We should start coming up with the usernames roy-the-deployer, sue-the-viewer, bradman-the-admin.

PR #533, opened 2024-02-13
HARMONY-1696: Ensure that STAC catalogs are returned in proper order

indiejames:
## Jira Issue ID
HARMONY-1696

## Description
This PR returns the STAC catalogs in proper numerical order rather than in string order. In other words catalog2.json
comes before catalog10.json. This is needed by Batchee to preserve the order it creates when it batches granules.

## Local Test Steps
Not easy to test locally. You will need to add Batchee and Stitchee to config.yml  as well as env vars for their queue URLS to lib/util/env-defaults. OR you can look at the following screen shots. The first one is our current code and shows that the Stitchee work-items were created in order of the "string sorted" STAC catalog file names. The Second one is from this branch and shows the Stichee work items were created in the order of the STAC catalogs created by Batchee, which is what we want. Note the rows from the dB are sorted by work-item ID in both cases.

Screenshot 2024-02-13 at 3 46 09 PM
Screenshot 2024-02-13 at 3 24 49 PM

## PR Acceptance Checklist
* [x] Acceptance criteria met

### DIFF
--- a/services/harmony/.nsprc
+++ b/services/harmony/.nsprc
@@ -26,7 +26,17 @@
   },
   "1096460": {
     "active": true,
-    "notes": "Will fix in HARMONY-",
+    "notes": "Will fix in HARMONY-1698",
+    "expiry": "2024-08-01"
+  },
+  "1096482": {
+    "active": true,
+    "notes": "Will fix in HARMONY-1700",
+    "expiry": "2024-08-01"
+  },
+  "1096484": {
+    "active": true,
+    "notes": "Will fix in HARMONY-1700",
     "expiry": "2024-08-01"
   }
 }
--- a/services/harmony/.nsprc
+++ b/services/harmony/.nsprc
@@ -23,5 +23,10 @@
     "active": true,
     "notes": "Will fix in HARMONY-1650",
     "expiry": "2024-05-01"
+  },
+  " 1096460": {
+    "active": true,,
+    "notes": "Will fix in HARMONY-",
+    "expiry": "2024-08-01"
   }
 }
--- a/services/harmony/.nsprc
+++ b/services/harmony/.nsprc
@@ -24,8 +24,8 @@
     "notes": "Will fix in HARMONY-1650",
     "expiry": "2024-05-01"
   },
-  " 1096460": {
-    "active": true,,
+  "1096460": {
+    "active": true,
     "notes": "Will fix in HARMONY-",
     "expiry": "2024-08-01"
   }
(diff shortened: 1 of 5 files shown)


services/service-runner/app/service/service-runner.ts
```diff
@@ -92,9 +92,27 @@ export class LogStream extends stream.Writable {
...
+    const result =  batchCatalogs.map(filename => `${dir}${filename}`);
+    return result;
+  }
+
+  // otherwise retrieve the keys from the bucket that are of the form catalog*.json,
+  // and sort them by index number
+  const urls = (await s3.listObjectKeys(dir))
+    .filter((fileKey) => fileKey.match(/catalog\d*.json$/))
+    .map((fileKey) => `${dir}${fileKey}`);
+  const fileNumRegex = /.*catalog(\d+)\.json$/;
```
chris-durbin, line 108: If there's only one catalog the file is named catalog.json (no digits in it). If this doesn't match anything will we still return the right thing?

I'm not sure if this code will get reached because I thought we might have a batchCatalogs.json file in all cases so maybe it doesn't matter, but we should test this logic.

services/service-runner/test/service-runner.ts
```diff
@@ -57,12 +77,114 @@ describe('Service Runner', function () {
...
+        ]);
+      });
+    });
+
+    describe('when there is no batch-catalogs.json file associated with the WorkItem', async function () {
+      describe('when there is just one catalog', async function () {
+        before(async function () {
+          const s3 = objectStoreForProtocol('s3');
+          const catalogContent = JSON.stringify(dummyCatalog);
+          const stacCatalogUrl = resolve(workItemWithOneCatalog, 'catalog.json');
```
chris-durbin, line 137: You are testing the scenario I was worried about so looks like it works.

services/harmony/.nsprc
```diff
@@ -23,5 +23,10 @@
     "active": true,
     "notes": "Will fix in HARMONY-1650",
     "expiry": "2024-05-01"
+  },
+  "1096460": {
+    "active": true,
+    "notes": "Will fix in HARMONY-",
```
chris-durbin, line 29: Missing the ticket number.

chris-durbin (approved):

chris-durbin (changes requested):
Ran into an issue testing in sandbox - the catalog location is getting set incorrectly to `<jobID>/<workItemId>/outputs/<jobID>/<workItemId>/outputs/catalog.json` when it should be just:
`<jobID>/<workItemId>/outputs/catalog.json`

services/service-runner/app/service/service-runner.ts
indiejames: as you noted below I have a test for this

services/service-runner/test/service-runner.ts
indiejames: yep

services/harmony/.nsprc
indiejames: fixed

ygliuvt (approved):

PR #540, opened 2024-02-28
modify start-postgres-localstack by adding wait time and start-harmon…

cirrusasf:
…y by adding check if harmony pod is ready

## Description
When I install harmony on my ubuntu20.04, sometimes it is not successful. The reason is that port forward fails because the harmony pod is not ready yet. I add statement in start-harmony to check if the harmony pod is ready before doing port forward.

## Local Test Steps
After I delete minikube, do fresh run ./bin/bootstrap-harmony. it works very time now.

## PR Acceptance Checklist
* [x] Acceptance criteria met


chris-durbin (commented):
I'm glad you were able to find a way to get the scripts to work consistently for you. I think we can keep the 30 second sleep you added after starting up minikube, and then remove the other sleeps and just update line 51 in `bin/port-forward` to wait for up to 120 seconds for a pod to reach the ready state. That way it should work for everyone without adding any extra time to the startup.

PR #542, opened 2024-03-04
HARMONY-1708: Update frontend and backend to be able to use TLS

chris-durbin:
## Jira Issue ID
HARMONY-1708

## Description
With the USE_HTTPS variable set to true both the harmony frontend and backend listeners will use TLS. We will set this to true in the deployed environments. See the ci-cd repo for instructions on how to test there.

## Local Test Steps
Verify that locally everything continues to work as expected by default (and does not use TLS). Verify in the sandbox that setting USE_HTTPS to true works correctly.

## PR Acceptance Checklist
* [X] Acceptance criteria met

### DIFF
--- a/services/harmony/app/server.ts
+++ b/services/harmony/app/server.ts
@@ -66,7 +66,7 @@ function addRequestId(appLogger: Logger): RequestHandler {
  * @param hostBinding - The host network interface to bind against
  * @returns The running express application
  */
-function buildBackendServer(port: number, hostBinding: string): http.Server | https.Server {
+function buildBackendServer(port: number, hostBinding: string, useHttps: string): http.Server | https.Server {
   const appLogger = logger.child({ application: 'backend' });
   const setRequestId = (req: HarmonyRequest, res: Response, next: NextFunction): void => {
     const { requestId } = req.params;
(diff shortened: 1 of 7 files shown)


services/harmony/app/server.ts
```diff
@@ -91,7 +93,19 @@ function buildBackendServer(port: number, hostBinding: string): Server {
   app.get('/', ((req, res) => res.send('OK')));
   app.use(errorHandler);
 
-  return app.listen(port, hostBinding, () => appLogger.info(`Application backend listening on ${hostBinding} on port ${port}`));
+  let listener;
+  if (useHttps === 'true') {
```
indiejames, line 97: For safety sake maybe change this to
```
if (useHttps.toLowerCase() === 'true') {
```

services/harmony/app/util/url.ts
```diff
@@ -7,6 +7,9 @@ import * as url from 'url';
  * @returns The protocol (http or https) to use for public Harmony URLs
  */
 function _getProtocol(req): string {
+  if (process.env.USE_HTTPS === 'true') {
```
indiejames, line 10: Same comment here about `toLowerCase`

services/harmony/app/server.ts
```diff
@@ -130,7 +144,19 @@ function buildFrontendServer(port: number, hostBinding: string, config: RouterCo
   ogcCoveragesApi.handleOpenApiErrors(app);
   app.use(errorHandler);
 
-  return app.listen(port, hostBinding, () => appLogger.info(`Application frontend listening on ${hostBinding} on port ${port}`));
+  let listener;
+  if (config.USE_HTTPS === 'true') {
```
indiejames, line 148: Same comment here about `toLowerCase`

services/harmony/app/server.ts
indiejames: Or maybe just convert it to lowercase when you first set up the config from the env var

indiejames (commented):
Looks good. I will test then approve pending outcome.

indiejames (approved):
Tested locally

indiejames (approved):
tested locally

services/harmony/app/server.ts
chris-durbin: I prefer checking for the exact string match - this is a config parameter we directly control so we don't need to be flexible on what we support unlike something like a query parameter from a user.

ygliuvt (approved):

PR #543, opened 2024-03-06
HARMONY-1706: Allow all aggregating steps to proeceed when a work-item fails, not just batched steps.

indiejames:
## Jira Issue ID
HARMONY-1706

## Description
This PR changes the logic that creates a work item for the next step so that even if the current work-item fails we might 
create a work item for the next step if it is an aggregating step. Previously we only did this if the step used batching. This
prevented Batchee (which is not batched,  but does aggregate) from running if a work-item fails. This change allows any
aggregating step to proceed.

## Local Test Steps
Since we don't have a reliable way to reproduce the original Batchee chain failures (the problem is transient), the best
we can do is make sure our other queries still work as expected. I am running the following query in the sandbox to make
sure it completes even though it always has a few failures:

```
https://internal-harmony-jn-1682-frontend-1034601994.us-west-2.elb.amazonaws.com/C1234724470-POCLOUD/ogc-api-coverages/1.0.0/collections/all/coverage/rangeset?subset=lat(-45%3A45)&subset=lon(0%3A180)&skipPreview=true&maxResults=100000&ignoreErrors=true
```

(in case you are wondering, the url uses jn-1682 because I did not do a full deployment from scratch - I just updated the deployment from my previous ticket, 1682)

## PR Acceptance Checklist
* [x] Acceptance criteria met

### DIFF
--- a/services/harmony/app/backends/workflow-orchestration/work-item-updates.ts
+++ b/services/harmony/app/backends/workflow-orchestration/work-item-updates.ts
@@ -266,14 +266,16 @@ async function getItemLinksFromCatalog(catalogPath: string): Promise<StacItemLin
  * @param nextStep - the next step in the workflow
  * @param results - an array of paths to STAC catalogs from the last worked item
  * @param logger - the logger to use
+ * @returns true if a new work item was created, false otherwise
  */
 async function createAggregatingWorkItem(
   tx: Transaction, currentWorkItem: WorkItem, nextStep: WorkflowStep, logger: Logger,
-): Promise<void> {
+): Promise<boolean> {
   const itemLinks: StacItemLink[] = [];
   const s3 = objectStoreForProtocol('s3');
   // get all the previous results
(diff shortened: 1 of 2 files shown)


chris-durbin (approved):
The changes make sense to me. Before merging just sanity check all of these test cases are covered:

* 2 work items, both fail, the job is failed and no work item gets queued for the aggregating step
* 2 work items, the first one fails and the second one succeeds and it queues one work item
* 2 work items, both succeed, 2 work items queued for the aggregating step (most likely this is already tested)

indiejames:
> The changes make sense to me. Before merging just sanity check all of these test cases are covered:
> 
> * 2 work items, both fail, the job is failed and no work item gets queued for the aggregating step
> * 2 work items, the first one fails and the second one succeeds and it queues one work item
> * 2 work items, both succeed, 2 work items queued for the aggregating step (most likely this is already tested)

Done.

ygliuvt (approved):

PR #544, opened 2024-03-06
HARMONY_1702: As a harmony operator, I want to ensure harmony does not need to be restarted when a backend service version is updated

ygliuvt:
## Jira Issue ID
HARMONY_1702

## Description
Add `deployment-callback` endpoint to allow updating harmony server with the new service image. This endpoint is supposed to be called only by the ci-cd `bin/deploy-service` script as part of the service self-deployment workflow.

## Local Test Steps
Tested in sandbox https://internal-harmony-yliu10-frontend-1209292950.us-west-2.elb.amazonaws.com,
Pick a harmony service to test, e.g. giovanni-adapter,
1. Submit a giovanni request, note down the service image tag of the successful request in workflow ui. 
https://internal-harmony-yliu10-frontend-1209292950.us-west-2.elb.amazonaws.com/C1225808238-GES_DISC/ogc-api-coverages/1.0.0/collections/Grid%2FprecipitationCal/coverage/rangeset?format=text%2Fcsv&point=0.76,-3.8&subset=time(%222020-01-06T12%3A00%3A00Z%22%3A%222020-01-06T16%3A00%3A00Z%22)

2. Call {{ _.root }}/service-image-tag/giovanni-adapter endpoint to update the giovanni-adapter to a different image with request body: `{"tag": "yliu10"}`

3. Wait one minute for the service self-deployment to run to completion. Then submit another giovanni request, verify that the request runs successfully and it is processed by the new giovanni service image with tag `yliu10`.

## PR Acceptance Checklist
* [x] Acceptance criteria met

### DIFF
--- a/services/harmony/app/frontends/deployment-callback.ts
+++ b/services/harmony/app/frontends/deployment-callback.ts
@@ -8,6 +8,14 @@ import _ from 'lodash';
  * Express.js handler that handls deployment callback message for deploying a service.
  * This function will only be called when the service deployment is successful.
  * Update the harmony server configuration for the new service image and service queue urls in the request.
+ * The callback_message in request body is in JSON. It has three fields:
+ *  deployService - name of the harmony service
+ *  image - full image url with tag for the deployed service
+ *  serviceQueueUrls - a list contains the service image to queue url mapping as defined in HarmonyEnv class.
+ * Here is an example:
+ *  \{"deployService":"giovanni-adapter",
+ *   "image":"123456.dkr.ecr.us-west-2.amazonaws.com/harmonyservices/giovanni-adapter:latest",
+ *   "serviceQueueUrls":"[\"123456.dkr.ecr.us-west-2.amazonaws.com/harmonyservices/giovanni-adapter:latest.....\"]"\}
  * @param req - The request sent by the client
  * @param res - The response to send to the client
  */
@@ -28,7 +36,7 @@ export default async function handleCallbackMessage(req: HarmonyRequest, res: Re
     services.resetServiceConfigs();
   } else {
     logger.error('You do not have permission to call deployment-callback endpoint');
-    res.statusCode = 400;
+    res.statusCode = 401;
     res.send('You do not have permission to call deployment-callback endpoint');
     return;
   }
--- a/services/harmony/app/frontends/deployment-callback.ts
+++ b/services/harmony/app/frontends/deployment-callback.ts
@@ -1,0 +1,29 @@
+import { Response } from 'express';
+import HarmonyRequest from '../models/harmony-request';
+import * as services from '../models/services/index';
+import env from '../util/env';
+import _, { get as getIn, partial } from 'lodash';
+
+
+/**
+ * Express.js handler that handls deployment callback message for deploying a service.
+ * This function will only be called when the service deployment is successful.
+ * Update the harmony server configuration for the new service image and service queue urls in the request.
+ * @param req - The request sent by the client
+ * @param res - The response to send to the client
+ */
+export default async function handleCallbackMessage(req: HarmonyRequest, res: Response): Promise<void> {
+  const logger = req.context.logger.child({ component: 'snsHandler.handleCallbackMessage' });
+  const { deployService, tag, image, serviceQueueUrls, status } = req.body;
+  logger.info(`handleCallbackMessage: ${JSON.stringify(req.body)}`);
+
+  const serviceImageEnv = _.camelCase(`${deployService.replace(/-/g, '_')}_image`);
+  const serviceQueueUrlsEnv = _.camelCase(`${deployService.replace(/-/g, '_')}_service_queue_urls`);
+
+  env[serviceImageEnv]=image;
+  env[serviceQueueUrlsEnv]=serviceQueueUrls;
+  env.servicesYml=undefined;
+  services.resetServiceConfigs();
+
+  res.send('OK');
+}
--- a/services/harmony/app/frontends/deployment-callback.ts
+++ b/services/harmony/app/frontends/deployment-callback.ts
@@ -2,8 +2,7 @@ import { Response } from 'express';
 import HarmonyRequest from '../models/harmony-request';
 import * as services from '../models/services/index';
 import env from '../util/env';
-import _, { get as getIn, partial } from 'lodash';
-
+import _ from 'lodash';
 
 /**
  * Express.js handler that handls deployment callback message for deploying a service.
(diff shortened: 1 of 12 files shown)


services/harmony/app/routers/router.ts
```diff
@@ -289,6 +290,7 @@ export default function router({ skipEarthdataLogin = 'false' }: RouterConfig):
   result.get('/service-image-tag', asyncHandler(getServiceImageTags));
   result.get('/service-image-tag/:service', asyncHandler(getServiceImageTag));
   result.put('/service-image-tag/:service', jsonParser, asyncHandler(updateServiceImageTag));
+  result.post('/deployment-callback', jsonParser, asyncHandler(getSnsMessage));
```
indiejames, line 293: Might it make more sense for this callback route to be on the back-end instead of the front-end?

services/harmony/app/frontends/service-image-tags.ts
```diff
@@ -279,11 +277,10 @@ export async function getServiceImageTag(
  * @param res  - The response object
  * @param service  - The name of the service to deploy
  * @param tag  - The service image tag to deploy
- * @returns a Promise containing the deployment status
  */
 export async function execDeployScript(
```
indiejames, line 281: Is there any reason for this function to be `async`? The`exec` call uses a callback, so you won't be waiting for it. I think you can remove the `async` keyword and the return type.

Also, I don't think `res` is used here, unless I'm missing it.

services/harmony/app/frontends/service-image-tags.ts
```diff
@@ -335,9 +332,7 @@ export async function updateServiceImageTag(
   const { service } = req.params;
   const { tag } = req.body;
 
-  const status = await module.exports.execDeployScript(req, res, service, tag);
-  if (status == successfulStatus) {
-    res.statusCode = 201;
-    res.send({ 'tag': tag });
-  }
+  await module.exports.execDeployScript(req, res, service, tag);
```
indiejames, line 335: As I mentioned above I think you can remove the `async` from the `execDeployScript` definition and then remove the `await` here.

services/harmony/app/routers/router.ts
```diff
@@ -26,6 +26,7 @@ import { cloudAccessJson, cloudAccessSh } from '../frontends/cloud-access';
 import landingPage from '../frontends/landing-page';
 import { setLogLevel } from '../frontends/configuration';
 import getVersions from '../frontends/versions';
+import getSnsMessage from '../frontends/deployment-callback';
```
indiejames, line 29: Should probably rename this now that it doesn't use SNS.

services/harmony/app/frontends/deployment-callback.ts
```diff
@@ -0,0 +1,37 @@
...
+    const serviceImageEnv = _.camelCase(`${deployService.replace(/-/g, '_')}_image`);
+    const serviceQueueUrlsEnv = _.camelCase(`${deployService.replace(/-/g, '_')}_service_queue_urls`);
+
+    env[serviceImageEnv] = image;
+    env[serviceQueueUrlsEnv] = serviceQueueUrls;
+    env.servicesYml = undefined;
+    services.resetServiceConfigs();
+  } else {
+    logger.error('You do not have permission to call deployment-callback endpoint');
+    res.statusCode = 400;
```
indiejames, line 31: Should probably be a 403

services/harmony/app/frontends/deployment-callback.ts
```diff
@@ -0,0 +1,37 @@
...
+ * @param res - The response to send to the client
+ */
+export default async function handleCallbackMessage(req: HarmonyRequest, res: Response): Promise<void> {
+  const logger = req.context.logger.child({ component: 'snsHandler.handleCallbackMessage' });
+  const headerSecret = req.headers['cookie-secret'];
+  const secret = process.env.COOKIE_SECRET;
+  logger.info(`handleCallbackMessage: ${JSON.stringify(req.body)}`);
+
+  if (headerSecret === secret) {
+    const { deployService, image, serviceQueueUrls } = req.body;
```
indiejames, line 21: I think it would be good to document what fields are in the message here. Not like JSON schema or anything, but maybe just a comment or define an interface above that documents the fields and use that here. Whichever, as long as there is some place that documents what the message fields are.

services/harmony/app/frontends/deployment-callback.ts
```diff
@@ -0,0 +1,37 @@
...
+    env[serviceQueueUrlsEnv] = serviceQueueUrls;
+    env.servicesYml = undefined;
+    services.resetServiceConfigs();
+  } else {
+    logger.error('You do not have permission to call deployment-callback endpoint');
+    res.statusCode = 400;
+    res.send('You do not have permission to call deployment-callback endpoint');
+    return;
+  }
+
```
indiejames, line 35: I think you want to set the status as well (201, I think?).

services/harmony/app/frontends/deployment-callback.ts
```diff
@@ -0,0 +1,37 @@
...
+  res.send('OK');
```
indiejames, line 36: I think you want to set the status to 201 here as well.

services/harmony/app/frontends/deployment-callback.ts
```diff
@@ -0,0 +1,37 @@
...
+    const { deployService, image, serviceQueueUrls } = req.body;
+    const serviceImageEnv = _.camelCase(`${deployService.replace(/-/g, '_')}_image`);
+    const serviceQueueUrlsEnv = _.camelCase(`${deployService.replace(/-/g, '_')}_service_queue_urls`);
+
+    env[serviceImageEnv] = image;
+    env[serviceQueueUrlsEnv] = serviceQueueUrls;
+    env.servicesYml = undefined;
+    services.resetServiceConfigs();
+  } else {
+    logger.error('You do not have permission to call deployment-callback endpoint');
```
indiejames, line 30: If you are going to log this error maybe you should log the message to aid in debugging.

services/harmony/app/frontends/service-image-tags.ts
ygliuvt: Good point! I am going to remove the async and res parameter.

services/harmony/app/frontends/deployment-callback.ts
ygliuvt: I thought we use 400 in Harmony for the Unauthorized (401) error. If I can put anything I want here. I would use 401.

chris-durbin: Agree with 403 (they have logged in but do not have permission). 401 would be if they failed to login.

services/harmony/test/deployment-callback.ts
```diff
@@ -0,0 +1,126 @@
...
+      it('rejects the request', async function () {
+        expect(this.res.status).to.equal(401);
+      });
+
+      it('returns a meaningful error message', async function () {
+        expect(this.res.text).to.equal('You do not have permission to call deployment-callback endpoint');
+      });
+    });
+
+    describe('when no cookie-secret header is provided', async function () {
```
indiejames, line 72: I think this should read 'when the wrong cookie-secret header is provided'

services/harmony/test/deployment-callback.ts
```diff
@@ -0,0 +1,126 @@
...
+      }
+    }
+  }
+  return undefined;
+}
+
+describe('Deployment callback endpoint', async function () {
+  hookServersStartStop();
+  let originalEnv;
+  before(async function () {
```
indiejames, line 38: Doesn't need to be `async`

indiejames (commented):
Once you make your changes I will re-review

chris-durbin (approved):

indiejames (approved):

PR #549, opened 2024-03-18
HARMONY-1703: Do not schedule additional work when there is a backlog on the work item update queue

chris-durbin:
## Jira Issue ID
HARMONY-1703

## Description
Prevents scheduling work when the small work item update queue is backed up since any new work is just going to end up at the end of that queue. This prevents issues with work being retried because it has been in the running state for too long and also helps ensure that fair queueing works as expected with new small requests able to be worked off quickly instead of being stuck behind a long queue to mark the work item as completed.

## Local Test Steps
It's tricky to test locally because things do not work in parallel like in an AWS environment. In sandbox I set `MAX_WORK_ITEMS_ON_UPDATE_QUEUE=10` and then ran a test with 250 concurrent pods with a 110K granule request. I saw the log message throughout the test: 

> Work item update queue is too large with 124 items, will not schedule more work`

The large request still completed in under 3 hours so even with pausing the scheduler throughout the throughput was still in line with performance before the change.

I set the default value to 1200 because we can work off about 20 work items per second which would allow up to a one minute backlog.

## PR Acceptance Checklist
* [X] Acceptance criteria met

### DIFF
--- a/services/work-scheduler/app/workers/scheduler.ts
+++ b/services/work-scheduler/app/workers/scheduler.ts
@@ -4,10 +4,11 @@ import env from '../util/env';
 import logger from '../../../harmony/app/util/log';
 import { logAsyncExecutionTime } from '../../../harmony/app/util/log-execution';
 import { Logger } from 'winston';
-import { getQueueUrlForService, getQueueForUrl, getWorkSchedulerQueue } from '../../../harmony/app/util/queue/queue-factory';
+import { getQueueUrlForService, getQueueForUrl, getWorkSchedulerQueue, getQueueForType } from '../../../harmony/app/util/queue/queue-factory';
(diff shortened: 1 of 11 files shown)


services/work-scheduler/app/workers/scheduler.ts
```diff
@@ -100,11 +101,15 @@ export function calculateNumItemsToQueue(
  * Read the scheduler queue and process any items in it
  *
  * @param reqLogger - a logger instance
+ * @param schedulerDisabledDelay - how long to sleep when the scheduler is disabled because
```
indiejames, line 104: I'm guessing this is in msec?

indiejames (approved):

ygliuvt (approved):

PR #556, opened 2024-03-29
HARMONY-1732: As a harmony user, I want to check my service self deployment status.

ygliuvt:
## Jira Issue ID
HARMONY-1732

## Description
This PR has the following changes:
- adds service_deployments table in harmony to track all service self-deployments. 
- adds `status_link` field in service image tag update endpoint response to list the url to track service self-deployment status
- adds new route `GET <harmony-root>/service-image-tag/deployment/<deployment-id>` and support to show the deployment details including status for the given `deployment-id`.

## Local Test Steps
Modify the command line in `execDeployScript` to simulate successful and failed service deployments. e.g. 
```
//const command = `./bin/exec-deploy-service ${service} ${tag}`;
const command = 'ls';
```
Verify:
1) when update a service tag, a `status_link` field is returned in the response. e.g.
curl -XPUT -H "Authorization: Bearer <token>" -H 'Content-type: application/json' http://localhost:3000/service-image-tag/harmony-service-example -d '{"tag": "latest"}'

2) verify following the `status_link`, user can see the deployment detail of the service tag update. It shows the service deployment status, service, tag, username who performed the service deployment, `createdAt` is the deployment start time, `updatedAt` is the deployment end time.

## PR Acceptance Checklist
* [x] Acceptance criteria met

### DIFF
--- a/docs/guides/managing-existing-services.md
+++ b/docs/guides/managing-existing-services.md
@@ -77,12 +77,12 @@ example above:
 ```
 **Example 6** - Harmony `/service-image-tags` request body for updating a tag
 
-The returned JSON response has a tag field indicating the new tag value and a status_link field with the url for getting the status of the service image deployment.
+The returned JSON response has a tag field indicating the new tag value and a statusLink field with the url for getting the status of the service image deployment.
 
 ```JSON
 {
   "tag": "new-version",
-  "status_link": "https://harmony.uat.earthdat.nasa.gov/service-image-tag/deployment/<deployment-id>"
+  "statusLink": "https://harmony.uat.earthdat.nasa.gov/service-image-tag/deployment/<deployment-id>"
 }
 ```
 **Example 7** - Harmony `/service-image-tags` response for a updating a single service
@@ -96,7 +96,7 @@ Harmony validates that the image and tag are reachable - an error will be return
 
 ## Get backend service image tag update status
 
-You can get the status of backend service tag update by following the `status_link` returned in the backend service tag update response.
+You can get the status of backend service tag update by following the `statusLink` returned in the backend service tag update response.
 
 For example:
 
@@ -110,12 +110,12 @@ The returned JSON response has the fields indicating the current status of the s
 ```JSON
 {
   id: 1,
-  deployment_id: "befb50e0-e467-4776-86c8-e7218f1123cc",
+  deploymentId: "befb50e0-e467-4776-86c8-e7218f1123cc",
   username: "yliu10",
   service: "giovanni-adapter",
   tag: "new-version",
   status: "successful",
-  message: null,
+  message: "Deployment successful",
   createdAt: "2024-03-29T14:56:29.151Z",
   updatedAt: "2024-03-29T14:56:29.273Z"
 }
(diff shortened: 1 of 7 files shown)


docs/guides/managing-existing-services.md
```diff
@@ -93,14 +94,42 @@ Harmony validates that the image and tag are reachable - an error will be return
...
+
+```JSON
+{
+  id: 1,
+  deployment_id: "befb50e0-e467-4776-86c8-e7218f1123cc",
+  username: "yliu10",
+  service: "giovanni-adapter",
+  tag: "new-version",
+  status: "successful",
+  message: null,
```
chris-durbin, line 118: I noticed the message is going to be null unless there is an issue. I think we should change the message to something generic like "Deployment in progress" when it first starts and "Deployment successful" when it is successful.

docs/guides/managing-existing-services.md
```diff
@@ -93,14 +94,42 @@ Harmony validates that the image and tag are reachable - an error will be return
...
+curl -XGET -Ln -bj https://harmony.uat.earthdat.nasa.gov/service-image-tag/deployment/<deployment-id>
+```
+**Example 8** - Get service image tag update status
+
+The returned JSON response has the fields indicating the current status of the service image deployment.
+
+```JSON
+{
+  id: 1,
+  deployment_id: "befb50e0-e467-4776-86c8-e7218f1123cc",
```
chris-durbin, line 113: I noticed both deployment_id and status_link are snake case, but other things like createdAt and updatedAt are camel case. On the jobs API we use camel case for everything so maybe use camel case everywhere for consistency?

chris-durbin (approved):

docs/guides/managing-existing-services.md
ygliuvt: I thought about doing it, but decided not to because the camel cases are from the db schema, given that we are changing all db column names to snake case, we probably want to get all the fields to snake case as well in the future. Looks like it bothers you, I can change it to camel case for now and we can change it back to snake case when we eventually change the db schemas.

services/harmony/test/service-image-tags.ts
```diff
@@ -830,11 +857,190 @@ describe('Service image endpoint', async function () {
...
+      execStub = sinon.stub(serviceImageTags, 'asyncExec').callsFake(() => Promise.resolve({}));
+
+      // Stub out the exec function to simulate successful execution
+      // eslint-disable-next-line @typescript-eslint/no-var-requires
+      execDeployScriptStub = sinon.stub(require('child_process'), 'exec');
+      execDeployScriptStub.callsArgWith(2, null, 'Success output', '');
+
+      this.res = await request(this.frontend).put('/service-image-tag/harmony-service-example').use(auth({ username: 'buzz' })).send({ tag: 'foo' });
+      // add delay for async db update to finish
+      await new Promise((res) => {
```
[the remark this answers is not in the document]
chris-durbin: I mean we shouldn't use sleeps in the test code (leads to brittle tests). Generally we would mock something in the tests to cause whatever event we wanted to trigger.

docs/guides/managing-existing-services.md
chris-durbin: The API and the database fields should be treated separately. We shouldn't have the API try to match the database, and we may not necessarily want to include all database fields in API responses.

indiejames (approved):

PR #567, opened 2024-04-26
Harmony 1757 - Redact specific headers

vinnyinverso:
## Jira Issue ID
HARMONY-1757

## Description
Gets rid of specific header values like auth, so that they don't show up in our logs.

## Local Test Steps
1. set `TEXT_LOGGER=false` in `.env`
2. Make a request to Harmony like `curl -H "Authorization: Bearer <my-bearer-token>" http://localhost:3000/C123/ogc-api-coverages/1.0.0/collections/bathymetry/coverage/rangeset`
3.  `"authorization":"<redacted>"` should show up in the logs

## PR Acceptance Checklist
* [x] Acceptance criteria met


services/harmony/app/server.ts
```diff
@@ -20,19 +20,28 @@ import * as exampleBackend from '../example/http-backend';
 import cmrCollectionReader from './middleware/cmr-collection-reader';
 import * as fs from 'fs';
 
+// redact specific values from a request log
+export const requestFilter: expressWinston.RequestFilter = (req, propName) => {
```
indiejames, line 24: nit: should probably use a named function here instead of an arrow function for consistency.

services/harmony/app/server.ts
```diff
@@ -20,19 +20,28 @@ import * as exampleBackend from '../example/http-backend';
...
+  if (propName === 'headers') {
+    const redacteProp = req[propName];
+    if (redacteProp.cookie) {
+      redacteProp.cookie = '<redacted>';
```
indiejames, line 28: nit: better to create a constant for 'redacted' since you are using it in more than one place

indiejames (approved):
Tested locally

ygliuvt (approved):

PR #571, opened 2024-04-30
HARMONY-1766 - Handle numerical parameters in POST requests.

owenlittlejohns:
## Jira Issue ID

HARMONY-1766

## Description

Running `harmony-py~=0.4.12` I noticed that requests with `height`, `width`, `scale_size` or `scale_extent` were causing 500 errors. It looks like this is due to the new use of the POST method for requests in this version of the package onwards. Looking at Harmony itself, these parameters are either numbers or lists of numbers, but are being passed around as strings, which has cause schema validation errors within Harmony.

This PR might not be the best way to fix it (I'm kind of just wading in trying to make a fix), but maybe it's okay? Totally happy if there's a better way to fix the issue, and this PR gets throw away, though.

## Local Test Steps

* Build a local Docker image of the Harmony Regridding service (it's _almost_ in open-source land, but for now, [use this script](https://git.earthdata.nasa.gov/projects/SITC/repos/harmony-regridding-service/browse/bin/build-image)).
* Pull this branch.
* Add `harmony-regridder` to your `.env` under `LOCALLY_DEPLOYED_SERVICES`.
* Build and run a local Harmony instance from this branch.
* Create a new Python environment and run `pip install harmony-py` (you'll probably get v0.4.13, which is the latest, and affected by this issue).
* Run the following command:

```
from harmony import Collection, Environment, Client, Request


harmony_client = Client(env=Environment.LOCAL)

merra_request = Request(
    Collection(id='C1245662776-EEDTEST'),
    granule_id='G1245662793-EEDTEST',
    interpolation='Elliptical Weighted Averaging',
    scale_size=(2.5, 2.5),
    scale_extent=(-180, -90, 180, 90),
    crs='+proj=latlong +datum=WGS84 +no_defs',
    height=72,
    width=144
)                       
merra_job_id = harmony_client.submit(merra_request)
harmony_client.wait_for_processing(merra_job_id, show_progress=True) 
```

This request should successfully trigger a job.

To see the issue, you can run the same snippet of code, but against UAT:

```
harmony_client = Client(env=Environment.UAT)
```

## PR Acceptance Checklist
* [x] Acceptance criteria met

### DIFF
--- a/services/harmony/app/util/parameter-parsers.ts
+++ b/services/harmony/app/util/parameter-parsers.ts
@@ -4,9 +4,30 @@
 
 import DataOperation from '../models/data-operation';
 import parseCRS from './crs';
-import { parseMultiValueParameter, parseNumber } from './parameter-parsing-helpers';
+import { ParameterParseError, parseMultiValueParameter, parseNumber } from './parameter-parsing-helpers';
 import HarmonyRequest from '../models/harmony-request';
 import { parseAcceptHeader } from './content-negotiation';
+import { RequestValidationError } from './errors';
+
+/**
+ * Helper function to convert parameter parsing errors into 400 errors for an end
+ * user that identifies the query parameter that was a problem and the message
+ * when failing to parse the parameter. If any other type of exception was thrown
+ * that exception is passed through unchanged.
+ *
+ * @param e - the exception that was thrown
+ * @param parameterName - the name of the query parameter that was being parsed
+ *
+ * @throws RequestValidationError if there was a specific parsing error message otherwise
+ * the original exception is thrown.
+ */
+function convertParameterParsingError(e: Error, parameterName: string): void {
+  if (e instanceof ParameterParseError) {
+    throw new RequestValidationError(`parsing query parameter '${parameterName}', value ${e.message}`);
+  } else {
+    throw (e);
+  }
+}
 
 /**
  * Handle the granuleName parameter in a Harmony query, adding it to the DataOperation
@@ -79,24 +100,22 @@ export function handleCrs(
  */
 export function handleScaleExtent(
   operation: DataOperation,
-  query: Record<string, number[]>): void;
-export function handleScaleExtent(
-  operation: DataOperation,
-  query: Record<string, string>): void;
-export function handleScaleExtent(
-  operation: DataOperation,
   query: Record<string, number[] | string>): void {
   if (query.scaleextent) {
-    if (typeof query.scaleextent === 'string') {
-      const scaleExtentString = query.scaleextent.replace('(', '').replace(')', '');
-      const [xMinString, yMinString, xMaxString, yMaxString] = scaleExtentString.split(/,\s*/);
-      operation.scaleExtent = {
-        x: { min: parseNumber(xMinString), max: parseNumber(xMaxString) },
-        y: { min: parseNumber(yMinString), max: parseNumber(yMaxString) },
-      };
-    } else {
-      const [xMin, yMin, xMax, yMax] = query.scaleextent;
-      operation.scaleExtent = { x: { min: xMin, max: xMax }, y: { min: yMin, max: yMax } };
+    try {
+      if (typeof query.scaleextent === 'string') {
+        const scaleExtentString = query.scaleextent.replace('(', '').replace(')', '');
+        const [xMinString, yMinString, xMaxString, yMaxString] = scaleExtentString.split(/,\s*/);
+        operation.scaleExtent = {
+          x: { min: parseNumber(xMinString), max: parseNumber(xMaxString) },
+          y: { min: parseNumber(yMinString), max: parseNumber(yMaxString) },
+        };
+      } else {
+        const [xMin, yMin, xMax, yMax] = query.scaleextent;
+        operation.scaleExtent = { x: { min: xMin, max: xMax }, y: { min: yMin, max: yMax } };
+      }
+    } catch (e) {
+      convertParameterParsingError(e, 'scaleExtent');
     }
   }
 }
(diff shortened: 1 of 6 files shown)


services/harmony/app/util/parameter-parsers.ts
```diff
@@ -79,10 +79,25 @@ export function handleCrs(
  */
 export function handleScaleExtent(
```
owenlittlejohns, line 80: TypeScript didn't like just using a union of types for the `query` unless I overloaded the function.

services/harmony/app/util/parameter-parsers.ts
```diff
@@ -79,10 +79,25 @@ export function handleCrs(
...
+  operation: DataOperation,
+  query: Record<string, string>): void;
+export function handleScaleExtent(
+  operation: DataOperation,
+  query: Record<string, number[] | string>): void {
   if (query.scaleextent) {
-    const [xMin, yMin, xMax, yMax] = query.scaleextent;
-    operation.scaleExtent = { x: { min: xMin, max: xMax }, y: { min: yMin, max: yMax } };
+    if (typeof query.scaleextent === 'string') {
+      const scaleExtentString = query.scaleextent.replace('(', '').replace(')', '');
```
owenlittlejohns, line 91: The string being passed around here was e.g., `(-180, -90, 180, 90)`, so I wanted to strip off the outer brackets before splitting the string.

I did also consider `query.scaleextent.slice(1, -1)`, but that seemed a little less clear as to what I was doing.

services/harmony/app/util/parameter-parsers.ts
```diff
@@ -79,10 +79,25 @@ export function handleCrs(
...
+      const [xMinString, yMinString, xMaxString, yMaxString] = scaleExtentString.split(/,\s*/);
```
owenlittlejohns, line 92: The value I found passed around within my local Harmony instance had numbers separated by ", " (so including a single space), but the tests were just comma-delimited. To be safe, I just opted to make the delimiter be a comma with an optional amount of whitespace after it. (Might be a bodge, but works in both my local instance and the tests)

services/harmony/app/util/parameter-parsers.ts
```diff
@@ -95,10 +110,22 @@ export function handleScaleExtent(
...
+  operation: DataOperation,
+  query: Record<string, string>): void;
+export function handleScaleSize(
+  operation: DataOperation,
+  query: Record<string, number[] | string>): void {
   if (query.scalesize) {
-    const [x, y] = query.scalesize;
-    operation.scaleSize = { x, y };
+    if (typeof query.scalesize === 'string') {
+      const scaleSizeString: string = query.scalesize.replace('(', '').replace(')', '');
```
owenlittlejohns, line 122: Similar story to `scaleextent`; the string passed around for `scalesize` was something like `(123,456)`.

services/harmony/app/util/parameter-parsers.ts
chris-durbin: I was able to just use the last definition without issue - maybe a VS code problem? Can you check what I pushed up to the branch to make sure you don't have any issues?

chris-durbin (approved):
Thanks Owen I tested successfully. I pushed up a change to make sure that failing to parse any of these values gets reported back to the end user with a 400 error with a good message instead of a generic 500 error. I'll also close out https://bugs.earthdata.nasa.gov/browse/HARMONY-1723 - we lost track of this bug, thanks for fixing it.

services/harmony/app/util/parameter-parsers.ts
owenlittlejohns: I pulled the new commit and my editor (vim :wink:)  seemed to like it, as did running the tests/type-checking. I think at the same time I implemented the overloads, I also cut over from `instanceof` to `typeof`, so that's probably really what quashed my type errors (and seems pretty obvious now I write it down 🙄).

services/harmony/app/util/parameter-parsers.ts
```diff
@@ -121,4 +165,44 @@ export function handleFormat(
...
\ No newline at end of file
+}
+
+/**
+ * Handle the height parameter in a Harmony query, adding it to the DataOperation
+ * if necessary.
+ *
+ * @param operation - the DataOperation for the request
+ * @param query - the query for the request
+ * @param req - the request
```
indiejames, line 176: There is no `req` param

services/harmony/app/util/parameter-parsers.ts
```diff
@@ -121,4 +165,44 @@ export function handleFormat(
...
+  }
+}
+
+/**
+ * Handle the width parameter in a Harmony query, adding it to the DataOperation
+ * if necessary.
+ *
+ * @param operation - the DataOperation for the request
+ * @param query - the query for the request
+ * @param req - the request
```
indiejames, line 196: No `req` param

owenlittlejohns: Good catch. I originally had one, then realised it wasn't being used. Will clean up now.

services/harmony/app/util/parameter-parsers.ts
```diff
@@ -79,10 +100,23 @@ export function handleCrs(
...
   operation: DataOperation,
-  query: Record<string, string>): void {
+  query: Record<string, number[] | string>): void {
   if (query.scaleextent) {
-    const [xMin, yMin, xMax, yMax] = query.scaleextent;
-    operation.scaleExtent = { x: { min: xMin, max: xMax }, y: { min: yMin, max: yMax } };
+    try {
+      if (typeof query.scaleextent === 'string') {
+        const scaleExtentString = query.scaleextent.replace('(', '').replace(')', '');
+        const [xMinString, yMinString, xMaxString, yMaxString] = scaleExtentString.split(/,\s*/);
```
indiejames, line 108: nit: I think it can be slightly more precise:
```
const [xMin, yMin, xMax, yMax] = scaleExtentString.split(/,\s*/).map(parseNumber);
operation.scaleExtent = {
          x: { min: xMin, max: xMax },
          y: { min: yMin, max: yMax },
        };
```

indiejames (approved):

PR #586, opened 2024-05-31
HARMONY-1787: Force queryVars to an array before validating and

indiejames:
fix bug in parsing item size as an int

## Jira Issue ID
HARMONY-1787

## Description
Fixes a bug in the way we validate variables that was triggered when using a POST for the query with variables that have 'all' in their names.

## Local Test Steps
I don't know of a collection that has 'all' in the variable names outside of production, so it is hard to do a full end-to-end test for this. There is a new unit test that passes queryVars as a string containing a variable with `all` in the name. This test fails without the changes to the variable validation code, as expected.

## PR Acceptance Checklist
* [x] Acceptance criteria met

### DIFF
--- a/services/harmony/app/util/aggregation-batch.ts
+++ b/services/harmony/app/util/aggregation-batch.ts
@@ -113,7 +113,7 @@ Promise<number[]> {
       outputItemSizes = outputItemSizes.concat(sizes);
     }
   }
-  return outputItemSizes;
+  return outputItemSizes.map(Math.floor);
 }
 
 /**
--- a/services/harmony/app/util/parameter-parsing-helpers.ts
+++ b/services/harmony/app/util/parameter-parsing-helpers.ts
@@ -40,10 +40,14 @@ export function parseNumber(valueStr: string | number): number {
 
 /**
  * Returns the parameter as parsed as an array of comma-separated values if
- * it was a string, or just returns the array if it's already parsed
+ * it was a string, or just returns the array if it's already parsed. Returns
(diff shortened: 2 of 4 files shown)


services/harmony/app/util/aggregation-batch.ts
```diff
@@ -113,7 +113,7 @@ Promise<number[]> {
       outputItemSizes = outputItemSizes.concat(sizes);
     }
   }
-  return outputItemSizes;
+  return outputItemSizes.map(Math.floor);
```
ygliuvt, line 116: We convert `number` to `integer`. Are we not afraid of losing precision?

indiejames: The item sizes are supposed to be the number of bytes, which should be an integer

indiejames: This is a fix for cases where a weird rounding error was giving us values like 97.0000000001, which were breaking the dB insertion

ygliuvt (approved):

chris-durbin (approved):
Verified when I include a variable `not_all` in my request to UAT it returned:
```
{
	"code": "harmony.RequestValidationError",
	"description": "Error: \"all\" cannot be specified alongside other variables"
}
```
With this branch it gets beyond that validation and returns:
``` 
{
	"code": "harmony.RequestValidationError",
	"description": "Error: Coverages were not found for the provided variables: not_all"
}
```

PR #589, opened 2024-06-05
Harmony 1779 - Adds the provider ids field to every job request

vinnyinverso:
## Jira Issue ID
HARMONY-1779

## Description
This PR adds a `provider_ids` `string[]` column for the jobs model/table, which stores a provider identifier (parsed from a collection `xyz-PROV123`) for each collection that is referenced in a request.

I also added a GIN index on that column https://www.postgresql.org/docs/current/indexes-types.html#INDEXES-TYPES-GIN so that we can eventually execute efficient searches for jobs based on provider(s).

## Local Test Steps
Submit a job and check the DB to see if the provider_ids column has been populated. 

`SELECT * FROM jobs WHERE provider_ids && '{"EEDTEST","OTHERPROV"}';` will pull back any job that has any of the provider ids specified in the array. (Update statements look like: `update jobs set provider_ids = '{prov,xyz}' WHERE id = 1;`.)

Run `SET enable_seqscan = OFF;` first (if you have a small number of jobs in the DB) and then `EXPLAIN ...` to see the index being used. You should see Bitmap Heap Scan followed by Bitmap Index Scan. 
> When postgres uses the GIN index, you’ll notice in the EXPLAIN that it’s always doing a bitmap heap scan. It means that the search algorithm returns a map of tids (pointers to rows) in physical memory order.

https://www.louisemeta.com/blog/indexes-gin-algorithms/#:~:text=When%20postgres%20uses%20the%20GIN,rows)%20in%20physical%20memory%20order.

## PR Acceptance Checklist
* [x] Acceptance criteria met

### DIFF
--- a/.nsprc
+++ b/.nsprc
@@ -26,7 +26,7 @@
   },
   "1097346": {
     "active": true,
-    "notes": "Will fix in HARMONY-1790",
+    "notes": "ignored because it doesn't affect us and there is not current patch",
     "expiry": "2024-08-01"
   }
 }
--- a/.nsprc
+++ b/.nsprc
@@ -23,5 +23,10 @@
     "active": true,
     "notes": "Will not fix. Only moderate vulnerability for functionality we don't use. Upgrade would break gdal-async dependency which is fixed to a specific version.",
     "expiry": "2024-08-01"
+  },
+  "1097346": {
+    "active": true,
+    "notes": "Will fix in HARMONY-1790",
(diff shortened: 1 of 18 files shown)


chris-durbin (approved):
Good find with the representation and best way to create the index. I tested locally and verified the provider_ids column is being populated correctly and the down migration works as expected.

indiejames (approved):

PR #596, opened 2024-06-19
HARMONY-1794: Default to keeping work items for 14 days after job completes

chris-durbin:
## Jira Issue ID
HARMONY-1794

## Description
Changes the default to keep work items in the database for 14 days after a job completes. NSIDC asked if we could keep them for at least 7 days and I took a look and found that the number of work items from old paused jobs often dwarfs the number that would be kept by increasing the retention so it will not have a significant impact to the size of the work items table in the database. We were already keeping them around for 3 days in both UAT and prod (we override the default with a variable in Bamboo - we will want to delete that override in Bamboo after deploying this change).

## Local Test Steps
There's no code change - just sanity check that with the new default value the work reaper does not throw any errors when running.

## PR Acceptance Checklist
* [X] Acceptance criteria met

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -21413,9 +21413,9 @@
       }
     },
     "node_modules/ws": {
-      "version": "8.14.2",
-      "resolved": "https://registry.npmjs.org/ws/-/ws-8.14.2.tgz",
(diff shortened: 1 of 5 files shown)


indiejames (approved):
I did not test this locally, but the number looks right for 14 days

ygliuvt (approved):

PR #597, opened 2024-06-21
HARMONY-1772: As a harmony service provider, I want to know the reason service deployments are disabled when I am unable to deploy

ygliuvt:
## Jira Issue ID
HARMONY-1772

## Description
As a harmony service provider, I want to know the reason service deployments are disabled when I am unable to deploy

## Local Test Steps
Manually change the `execDeployScript` command in service-image-tags.ts to something like a `sleep 30` that can be executed locally, e.g.
```
// const command = `./bin/exec-deploy-service ${service} ${tag}`;
const command = 'sleep 30';
```

Then, try to enable/disable service deployment state with and without an optional message field and verify when getting the service deployment state, the result reflect the correct state and message used in the enable/disable call.

See below for a couple of the example requests:
```
curl -H "Authorization: Bearer <token>" http://localhost:3000/service-deployments-state

curl -i -XPUT -H "Authorization: Bearer <token>" -H 'Content-type: application/json' http://localhost:3000/service-deployments-state -d '{"enabled": true, "message": "some custom message"}'
```

Then, run something similar to the following simulating common enable/disable and self deployment scenarios:

### Manually disable service deployment
curl -i -XPUT -H "Authorization: Bearer <token>" -H 'Content-type: application/json' http://localhost:3000/service-deployments-state -d '{"enabled": false, "message": "Disable for local testing"}'

### Try to deploy a service, and verify the deployment failed with status code 423 and message indicating service deployment is locked for local testing.
`Service deployment is disabled. Reason: Disable for local testing.`

### Manually enable service deployment
curl -i -XPUT -H "Authorization: Bearer <token>" -H 'Content-type: application/json' http://localhost:3000/service-deployments-state -d '{"enabled": true}'

### Deploy a service with a specified tag
curl -i -XPUT -H "Authorization: Bearer <token>" -H 'Content-type: application/json' http://localhost:3000/service-image-tag/harmony-service-example -d '{"tag": "latest"}'

### Get service-deployments-state, it should show service deployment is disabled and the message indicating it is locked for service deployment with a link to the service deployment status url. e.g.
{"enabled":false,"message":"Locked for service deployment: http://localhost:3000/service-deployment/e791620b-c676-4163-84fe-f225ca14d355"}

### After the sleep time expires, get the service-deployments-state again, it should show the service-deployments-state is enabled. e.g.
{"enabled":true,"message":"Re-enable service deployment after successful deployment: e791620b-c676-4163-84fe-f225ca14d355"}

And there is a lot of other scenarios that you can test with.

## PR Acceptance Checklist
* [x] Acceptance criteria met

### DIFF
--- a/services/harmony/app/frontends/service-image-tags.ts
+++ b/services/harmony/app/frontends/service-image-tags.ts
@@ -5,6 +5,7 @@ import { getEdlGroupInformation, validateUserIsInCoreGroup } from '../util/edl-a
 import { exec } from 'child_process';
 import * as path from 'path';
 import util from 'util';
+import { truncateString } from '@harmony/util/string';
 import db from '../util/db';
 import env from '../util/env';
 import { getRequestRoot } from '../util/url';
@@ -65,7 +66,7 @@ export function ecrImageNameToComponents(name: string): EcrImageNameComponents {
   const match = name.match(componentRegex);
   if (!match) return null;
 
-  const [ _, host, region, repository, tag ] = match;
+  const [_, host, region, repository, tag] = match;
   const registryId = host.split('.')[0]; // the AWS account ID
 
   return {
@@ -155,7 +156,8 @@ async function getEnabledAndMessage(): Promise<{ enabled: boolean, message: stri
  */
 export async function enableServiceDeployment(message: string): Promise<void> {
   await db.transaction(async (tx) => {
-    await tx('service_deployment').update({ enabled: true, message: message });
+    await tx('service_deployment')
+      .update({ enabled: true, message: truncateString(message, 4096) });
   });
 }
 
@@ -345,7 +347,7 @@ async function validateStatusForListServiceRequest(
 async function validateDeploymentState(
   req: HarmonyRequest, res: Response,
 ): Promise<boolean> {
-  if ( (!req.body || req.body.enabled === undefined)) {
+  if ((!req.body || req.body.enabled === undefined)) {
     res.statusCode = 400;
     res.send('\'enabled\' is a required body parameter');
     return false;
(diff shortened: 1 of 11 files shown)


services/harmony/app/frontends/service-image-tags.ts
```diff
@@ -130,29 +130,32 @@ async function validateTaggedImageIsReachable(
...
 }
 
 /**
  * Set the enabled to true in service_deployment table
  */
-export async function enableServiceDeployment(): Promise<void> {
+export async function enableServiceDeployment(message: string): Promise<void> {
   await db.transaction(async (tx) => {
-    await tx('service_deployment').update({ enabled: true });
+    await tx('service_deployment').update({ enabled: true, message: message });
```
chris-durbin, line 158: We need to truncate to less than 4096 characters (we have a truncate function we use for other database fields with a fixed length). Otherwise we return a 500 error to the end user with an HTML page.

I think there's something in `error.ts` to make sure we return JSON for some endpoints instead of HTML - might need to add the service-deployments-state route to that.

chris-durbin (commented):
Tested in sandbox - looked good and I liked the default messages.

Only 2 issues I saw were needing to truncate the message if longer than 4096 characters and when there's no message it returns `{"enabled":true,"message":null}`. I wouldn't worry about changing anything for that null case though because after the first enable/disable after deploying this code there should always be a message.

chris-durbin (approved):
Tested successfully. I see the message that gets returned from the call to enable/disable has the full string when it's more than 4096 characters, but when calling the endpoint to get the current state it is showing the truncated message.

vinnyinverso (approved):

PR #599, opened 2024-06-25
Harmony 1784

indiejames:
## Jira Issue ID
HARMONY-1784

## Description
Adds code to split shapefile geometry at the antimeridian as well as normalizing longitudes to the [-180,180] range
## Local Test Steps
Run the following `curl`:
```
curl -Ln -bj  "http://localhost:3000/C1256535487-NSIDC_CUAT/ogc-api-coverages/1.0.0/collections/all/coverage/rangeset/?maxResults=1" -F "shapefile=@shapefile.geojson;type=application/geo+json"
```

where the shapefile is as follows:
```

  "type": "FeatureCollection",
  "features": [
    {
      "type": "Feature",
      "properties": {},
      "geometry": {
        "coordinates": [[[-189.39835546840183, -71.30184718454765], [-196.93423129853957, -74.24875148066187], [-199.01126455296102, -76.99500221847319], [-196.01615081260698, -78.29244034236493], [-162.51718990801453, -79.03063047059774], [-148.37161275139798, -76.90896062169273], [-142.83828400653184, -75.05376361670923], [-189.39835546840183, -71.30184718454765]]],
        "type": "Polygon"
      }
    }]
}
```

The job should make it past query-cmr (trajectory-subsetter will fail, currently).

## PR Acceptance Checklist
* [x] Acceptance criteria met


services/harmony/app/middleware/shapefile-converter.ts
```diff
@@ -1,6 +1,9 @@
+/* eslint-disable require-jsdoc */
```
chris-durbin, line 1: We should fill in the JS docs for the functions.

services/harmony/app/middleware/shapefile-converter.ts
```diff
@@ -81,6 +84,103 @@ const contentTypesToConverters = {
...
+export function normalizeGeoJsonCoords(geojson: any): any {
+  function normalizeCoordinates(coordinates: any): any {
+    if (Array.isArray(coordinates[0])) {
+      return coordinates.map(normalizeCoordinates);
+    } else {
+      return [normalizeLongitude(coordinates[0]), coordinates[1]];
+    }
+  }
+
+  function normalizeGeometry(geometry: any): any {
```
chris-durbin, line 112: I think you might be able to specify a type here from the GeoJSON libraries. Worst case we can define an interface with `type`, `coordinates`, and `geometries`.

services/harmony/app/middleware/shapefile-converter.ts
```diff
@@ -81,6 +84,103 @@ const contentTypesToConverters = {
...
+  return newGeoJson;
+}
+
+/**
+ * Handle any weird cases like splitting geometry that crosses the antimeridian
+ * @param url - the url of the geojson file
+ * @param isLocal - whether the url is a downloaded file (true) or needs to be downloaded (false)
+ */
+async function normalizeGeoJsonFile(url: string, isLocal: boolean, logger: Logger): Promise<string> {
+  logger.debug('Normalizing geojson');
```
chris-durbin, line 168: This message doesn't look to useful unless we're going to log some info about the geometry.

services/harmony/app/middleware/shapefile-converter.ts
```diff
@@ -81,6 +84,103 @@ const contentTypesToConverters = {
...
+    newGeoJson['features'][index] = normalizedGeoJson;
+  }
+
+  // force ccw winding
+  newGeoJson = rewind(newGeoJson, false);
+  return newGeoJson;
+}
+
+/**
+ * Handle any weird cases like splitting geometry that crosses the antimeridian
```
chris-durbin, line 163: The docstring doesn't make it obvious that this is doing a download and an upload of a file. This is going to cause a delay anytime we receive a shapefile. Is there a way we can rewrite things to not store the original JSON to S3, only store the normalized GeoJSON file?

services/harmony/app/middleware/shapefile-converter.ts
```diff
@@ -81,6 +84,103 @@ const contentTypesToConverters = {
...
+  let originalGeoJson: object;
+  const localFile = url;
+  if (!isLocal) {
+    originalGeoJson = await store.getObjectJson(url);
+  } else {
+    originalGeoJson = (await fs.readFile(localFile)).toJSON();
+  }
+  const normalizedGeoJson = normalizeGeoJson(originalGeoJson);
+  const uploadUrl = `${url}-normalized.geojson`;
+  await store.upload(JSON.stringify(normalizedGeoJson), uploadUrl);
```
chris-durbin, line 179: If there is no change to the GeoJSON we shouldn't need to upload the file. So if we can't figure out how to prevent the extra read of the file we could at least prevent uploading it.

services/harmony/app/middleware/shapefile-converter.ts
indiejames: I disagree. This disable is here because there are a couple of utility functions that are defined internally to another function (something I wouldn't do myself, but chatGPT did). These functions don't represent any sort of API that needs documenting.

services/harmony/app/middleware/shapefile-converter.ts
indiejames: This is chatGPT black box wizardry that I'm not really wanting to mess around with. I definitely don't see any benefit to adding types for it, but we can discuss if you prefer to add types.

services/harmony/app/middleware/shapefile-converter.ts
indiejames: Removed

services/harmony/app/middleware/shapefile-converter.ts
chris-durbin: The internal to a function ones are fine, but we should fill out the JS docs for the top level functions.

services/harmony/app/middleware/shapefile-converter.ts
chris-durbin: Types are just documentation - we try to limit the use of `any` as much as possible. I don't see a reason for ignoring types here when we know what type it supports and returns. If we really want to use `any` here, I think we should get rid of the ignore types at the top of the file and put it above each line we want to ignore types for.

services/harmony/app/middleware/shapefile-converter.ts
chris-durbin: Similar comment for the other lint ignores, I'd rather see explicit ignores above each line that at the top level of the file because we don't want to ignore JS docs for all the functions in this file, just the internal to a function ones if you want to keep them defined inside the function.

services/harmony/app/middleware/shapefile-converter.ts
indiejames: I changed this to only do the upload of the file has changed.

services/harmony/app/middleware/shapefile-converter.ts
indiejames: I changed this to only do the upload of the file has changed.

services/harmony/app/middleware/shapefile-converter.ts
indiejames: Ok, I removed the top level linter ignore and added specific ones for the internal functions, along with JSDocs for the top-level ones missing them. I do with there was a way to tell VS Code to not underline everything in yellow just because I haven't written a JSDoc yet. It's annoying that it happens immediately when you create a function.

services/harmony/app/middleware/shapefile-converter.ts
indiejames: Regarding types, if we were going to use them somewhere else, then sure, I would add them. But in this case, we are just  passing the geojson as an `object` and checking various fields. Creating a type that captures all of geojson  seems like overkill. This function is not exported so is not part of some external API. We could change `any` to `object` if you prefer - again, this is just what chatGPT came up with. 

Regarding the single line vs entire file linter disable, I had to use the entire file disable, because I'm already using a "disable for next line" linter disable comment for JSDocs for the internal function. ESLint gets confused if you have two consecutive "disable for next line" comments.

ygliuvt (approved):

chris-durbin (approved):

PR #601, opened 2024-06-26
Harmony 1791 -  filter the jobs on the /admin/workflow-ui to only those for requests for collections for a given provider

vinnyinverso:
## Jira Issue ID
HARMONY-1791

## Description
Adds a provider id column for jobs, that is searchable via the Workflow UI.

## Local Test Steps
Submit some jobs for collections that have different providers. Or change values in the provider_id column manually to simulate different providers. Search for these jobs in the admin Workflow UI (/admin/workflow-ui) using `prov: provider_id` syntax and verify that the filtering is working as expected. (Adding multiple prov constraints performs an OR, _not_ AND.)

Screenshot 2024-06-26 at 1 34 20 PM

## PR Acceptance Checklist
* [x] Acceptance criteria met


services/harmony/app/frontends/workflow-ui.ts
```diff
@@ -261,6 +271,12 @@ function tableQueryToJobQuery(tableQuery: TableQuery, isAdmin: boolean, user: st
       in: tableQuery.allowUsers,
     };
   }
+  if (tableQuery.providerValues.length) {
```
indiejames, line 274: nit: For consistency maybe change this to `if (tableQuery.providerValues.length > 0)` since you use that style in `data-operation.ts` below.

indiejames (approved):
Works for me. Just one small style comment. I do think it would be nice to implement autocomplete for the provider IDs in the UI, though that can be a separate ticket if we decide to do it.

services/harmony/app/frontends/workflow-ui.ts
vinnyinverso: Good point.

ygliuvt:
Why is this filter only applicable in `admin/workflow-ui`? why not `workflow-ui`?

ygliuvt (approved):
Tested successfully locally.

vinnyinverso:
@ygliuvt - good question. I think the rationale is that for a normal user, they only see their own jobs, and probably don't care much about provider ids.

PR #603, opened 2024-07-03
HARMONY-1806: Update error logging when handling STAC parsing exception.

ygliuvt:
## Jira Issue ID
HARMONY-1806

## Description
Update error logging when handling STAC parsing exception.

## Local Test Steps
Submit a harmony request that will cause STAC parsing exception (due to a bug in l2ss-py) and verify that Harmony logs the stacktrace properly in the log. e.g. 
http://localhost:3000/C1261072645-POCLOUD/ogc-api-coverages/1.0.0/collections/all/coverage/rangeset?forceAsync=true&subset=lat(34.84931%3A35.83342)&subset=lon(-125.57812%3A-124.73438)&subset=time(%222023-11-15T18%3A00%3A00%2B00%3A00%22%3A%222023-11-18T07%3A00%3A00%2B00%3A00%22)&skipPreview=true&concatenate=true

## PR Acceptance Checklist
* [x] Acceptance criteria met

### DIFF
--- a/services/harmony/app/backends/workflow-orchestration/work-item-updates.ts
+++ b/services/harmony/app/backends/workflow-orchestration/work-item-updates.ts
@@ -564,13 +564,14 @@ export async function preprocessWorkItem(
     durationMs = new Date().getTime() - resultStartTime;
     logger.debug('timing.HWIUWJI.getResultItemSize.end', { durationMs });
   } catch (e) {
-    errorMessage = 'Could not get result item file size from output STAC';
+    if (catalogItems) {
+      errorMessage = 'Cannot get result item file size from output STAC';
+    } else {
+      errorMessage = 'Cannot parse STAC catalog output from service';
+    }
+
     logger.error(errorMessage);
-    logger.error('Caught exception:', {
-      message: e.message,
-      name: e.name,
-      stack: e.stack,
-    });
+    logger.error(`Caught exception: ${e.message}`, { name: e.name, stack: e.stack });
     status = WorkItemStatus.FAILED;
   }
   const result: WorkItemPreprocessInfo = {
--- a/services/harmony/app/backends/workflow-orchestration/work-item-updates.ts
+++ b/services/harmony/app/backends/workflow-orchestration/work-item-updates.ts
@@ -564,9 +564,13 @@ export async function preprocessWorkItem(
     durationMs = new Date().getTime() - resultStartTime;
     logger.debug('timing.HWIUWJI.getResultItemSize.end', { durationMs });
   } catch (e) {
-    errorMessage = 'Could not get result item file size';
+    errorMessage = 'Could not get result item file size from output STAC';
     logger.error(errorMessage);
-    logger.error(e);
+    logger.error('Caught exception:', {
+      message: e.message,
+      name: e.name,
+      stack: e.stack,
+    });
     status = WorkItemStatus.FAILED;
   }
   const result: WorkItemPreprocessInfo = {


services/harmony/app/backends/workflow-orchestration/work-item-updates.ts
```diff
@@ -564,9 +564,13 @@ export async function preprocessWorkItem(
     durationMs = new Date().getTime() - resultStartTime;
     logger.debug('timing.HWIUWJI.getResultItemSize.end', { durationMs });
   } catch (e) {
-    errorMessage = 'Could not get result item file size';
+    errorMessage = 'Could not get result item file size from output STAC';
```
ygliuvt, line 567: Couldn't come up with a better error message.

indiejames: I think maybe something like `Service ${service} failed to produce usable output` might work better for end users.

indiejames (commented):
Testling locally, but I think we should update the error message.

services/harmony/app/backends/workflow-orchestration/work-item-updates.ts
ygliuvt: Agree the message with service name is better, but the `service` is not available in this function and I am not sure if it is worth it to somehow get the service name in this function for logging only.

services/harmony/app/backends/workflow-orchestration/work-item-updates.ts
```diff
@@ -564,9 +564,13 @@ export async function preprocessWorkItem(
...
     logger.error(errorMessage);
-    logger.error(e);
+    logger.error('Caught exception:', {
+      message: e.message,
```
chris-durbin, line 570: With JSON logging and cloud metrics they can't have a field called message so it will likely break parsing or get converted to something else. Is there something that was missing with the call to just `logger.error(e)`? I thought that should be showing the stack trace.

chris-durbin: Maybe `Caught exception: ${e.message}\`, {name: e.name, stack: e.stack}`

services/harmony/app/backends/workflow-orchestration/work-item-updates.ts
chris-durbin: Is there a way we can have different messages for different issues? For this particular one I think I'd say something like "Cannot parse STAC catalog output from service." and expect it would get thrown from trying to read the catalog and not a call to get the file size.

services/harmony/app/backends/workflow-orchestration/work-item-updates.ts
ygliuvt: You just raised a great point and pointed out the real cause of this ticket. There is no exception logged in Cloud Metrics is not because the exception is not logged by Harmony. It is actually caused by Cloud Metrics not able to handle the logging with message!

chris-durbin (commented):
I tested this successfully locally and see the error is set appropriately in the job status error field - make sure we come to an agreement on the message before merging.

services/harmony/app/backends/workflow-orchestration/work-item-updates.ts
ygliuvt: Done.

indiejames (approved):
Working for me

chris-durbin (approved):

PR #609, opened 2024-07-26
Harmony 1613 - Ensure user is sent to page 1 whenever filters are applied

vinnyinverso:
## Jira Issue ID
HARMONY-1613

## Description
This fixes an issue where a user applies filters in the Workflow UI and then ends up on a page with unexpected results because the page number was maintained. E.g. you may be on page 5 (with three rows of jobs), you add a filter and click the apply button, and the server returns page 5 again, but this time with no results on it. 

What should happen is that they are sent to page 1 instead. (This is the expected behavior according to @chris-durbin.)

## Local Test Steps
Verify that when a user clicks the "submit" button on the Workflow UI, they should be sent to the first page. 

## PR Acceptance Checklist
* [x] Acceptance criteria met

### DIFF
--- a/services/harmony/app/views/workflow-ui/job/index.mustache.html
+++ b/services/harmony/app/views/workflow-ui/job/index.mustache.html
@@ -59,7 +59,8 @@
             <div class="col-2">
                 <form id="work-items-query-form" action="./{{job.jobID}}" method="get" class="pt-2">
                     <input type="hidden" name="jobID" value="{{job.jobID}}" />
-                    <input type="hidden" name="page" value="{{page}}" />
+                    <input type="hidden" name="currentPage" value="{{page}}" />
+                    <input type="hidden" name="page" value="1" />
                     {{> workflow-ui/date-time-picker}}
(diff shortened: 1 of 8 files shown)


indiejames (commented):
Looks good and I tested it locally successfully, but is there any way to add a test for this?

vinnyinverso:
There are already tests for paging (making sure the page parameter is accepted and handled appropriately) and we don't currently have full end-to-end testing that could simulate the button click and form submission, so I couldn't think of anything to add. 

I could add a test that inspects that `<input type="hidden" name="page" value="1" />` is returned. Do you think that'd be useful?

indiejames:
> There are already tests for paging (making sure the page parameter is accepted and handled appropriately) and we don't currently have full end-to-end testing that could simulate the button click and form submission, so I couldn't think of anything to add.
> 
> I could add a test that inspects that `<input type="hidden" name="page" value="1" />` is returned. Do you think that'd be useful?
Yes

chris-durbin (approved):

indiejames (approved):

PR #618, opened 2024-08-29
HARMONY-1789: Change SAMBAH to only concatenate when the `concatenate` flag is set to true and update the description

indiejames:
## Jira Issue ID
HARMONY-1789

## Description
Changes the logic for SAMBAH wrt the `concatenate` flag. Also enables `shapefile` subsetting and updates the online description of the service chain.

## Local Test Steps
1. Deploy this branch
2. Run a query that will access the chain, with no `concatenate` flag passed, `concatenate` set to `false`, and `concatenate` set to `true`, with or without a shapefile. You should see all three queries invoking SAMBAH, but on the last request it should run all of query-cmr/l2ss/batchee/stitchee/concise (the first two just run query-cmr and l2ss).
```
http://localhost:3000/C1261880450-LARC_CLOUD/ogc-api-coverages/1.0.0/collections/all/coverage/rangeset?subset=lat(-70%3A30)&subset=lon(-157.5%3A-10)&concatenate=true&extend=true&maxResults=4
```
3. Check that the online documentation has a multiline description for SAMBAH now.
```
http://localhost:3000/docs
```
4. Run `npm run compare-services` and you should see a couple of warning messages about SAMBAH (since services.yml is not the same as what is in UMM until they update the record), but the script should still exit with code 0 (success). Try changing the entry for some other service in services.yml and running again (need to restart harmony) and verify that this time an error message is printed and the script return a non-zero exit code.

## PR Acceptance Checklist
* [x] Acceptance criteria met

### DIFF
--- a/scripts/service-comparison.ts
+++ b/scripts/service-comparison.ts
@@ -8,6 +8,7 @@ import { exit } from 'process';
 import { loadServiceConfigs } from '../services/harmony/app/models/services';
 import { CmrUmmService, cmrApiConfig, getServicesByIds } from '../services/harmony/app/util/cmr';
 import { ServiceConfig } from '../services/harmony/app/models/services/base-service';
+import validation from 'ajv/dist/vocabularies/validation';
 
 /**
  * Validates spatial subsetting configuration matches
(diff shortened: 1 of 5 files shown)


scripts/service-comparison.ts
```diff
@@ -167,8 +167,14 @@ async function runComparisons(environments = allEnvironments): Promise<void> {
       const ummRecord = ummRecordsMap[harmonyConfig.umm_s];
       const validationMessages = performValidations(ummRecord, harmonyConfig);
       if (validationMessages.length > 0) {
-        exitCode = 1;
-        console.log(`Validation failures for ${harmonyConfig.name} and ${ummRecord.meta['concept-id']}:\n    - ${validationMessages.join('\n    - ')}`);
+        // TODO this is a temporary check until the UMM records for this service chain are updated
```
chris-durbin, line 170: We're allowing for a difference in `concatenate_by_default` since for API users they want that to be false, but in EDSC they want the default to be to have the box checked. This isn't temporary, but the other shapefile discrepancy is.

I'm thinking instead we should still error out for the shape discrepancy until UMM-S is updated, and we shouldn't do any kind of temporary disable for that, but we should have code in to make the check for concatenate_by_default to always print a warning and not treat as an error.

chris-durbin (commented):
I think we should change the service-comparison check - everything else looks good.

scripts/service-comparison.ts
indiejames: My build won't pass if this script errors, right?

chris-durbin: The compare services check will be marked as failed, but that does not block merging in the PR. It stays marked as failed so we know there is a change to UMM-S records that needs to happen.

indiejames: Fixed.

ygliuvt (approved):

chris-durbin (approved):

PR #626, opened 2024-09-13
HARMONY-1878: As a Harmony user, I want to use OGC EDR API trajectory route to submit a harmony request

ygliuvt:
## Jira Issue ID
HARMONY-1878

## Description
Add OGC EDR trajectory support in Harmony.

### OGC EDR trajectory API documentation:
http://localhost:3000/docs#ogc-edr-trajectory-subsetting-query-parameters

### LINESTRING coords
http://localhost:3000/ogc-api-edr/1.1.0/collections/C1234208438-POCLOUD/trajectory?maxResults=1&parameter-name=all&coords=LINESTRING(-40%2010%2C%2030%2010)

### MULTILINESTRING coords
http://localhost:3000/ogc-api-edr/1.1.0/collections/C1234208438-POCLOUD/trajectory?maxResults=1&parameter-name=all&coords=MULTILINESTRING((-40%2010%2C%2030%2010),(10%201%2C%2010%2030))

## PR Acceptance Checklist
* [x] Acceptance criteria met


services/harmony/app/frontends/ogc-edr/get-data-for-trajectory.ts
```diff
@@ -0,0 +1,280 @@
...
+*
+* @param points - The vertices of the polygon.
+* @returns The vertices in counter-clockwise order.
+*/
+function ensureCounterClockwise(points: Array<{ x: number; y: number }>): Array<{ x: number; y: number }> {
+  const area = calculateSignedArea(points);
+
+  // If the area is negative, the points are in clockwise order, so reverse them
+  if (area < 0) {
+    return points.reverse();
```
indiejames, line 43: `reverse()` does an in-place reverse, so the input array will be changed. Make sure this is what is expected from the caller.

services/harmony/app/frontends/ogc-edr/get-data-for-trajectory.ts
```diff
@@ -0,0 +1,280 @@
...
+
+/**
+ * Converts a two-point segment (line) into a polygon with a buffer.
+ * Ensures that the polygon points are in counter-clockwise order.
+ *
+ * @param segment - The two points defining the line segment.
+ * @param sideLength - The buffer distance (side length) to create around the segment.
+ * @returns The vertices of the polygon in counter-clockwise order.
+ */
+function convertSegmentToPolygon(segment: Array<{ x: number; y: number }>, sideLength: number): Array<{ x: number; y: number }> {
```
indiejames, line 57: You use this type definition in a lot of place. It would probably be good to create a formal definition for it like
```
type PointArray = Array<{x: number, y: number}>;
```
Then this function declaration becomes
```
function convertSegmentToPolygon(segment: PointArray, sideLength: number): PointArray
```
Then you can change all the other functions that use this type to use your new type declaration as well.

indiejames: Or maybe better, just define a Point type, like this
```
type Point =  {x: number, y: number};
```
Then your function becomes this
```
function convertSegmentToPolygon(segment: Point[], sideLength: number): Point[]
```

indiejames: Actually, to enforce the two-point nature of a segment you could do this
```
function convertSegmentToPolygon(segment: [Point, Point], sideLength: number): Point[]
```

services/harmony/app/frontends/ogc-edr/get-data-for-trajectory.ts
```diff
@@ -0,0 +1,280 @@
...
+  // Ensure the polygon is counter-clockwise
+  polygonPoints = ensureCounterClockwise(polygonPoints);
+
+  return polygonPoints;
+}
+
+/**
+* Converts a WKT LineString coords part into a list of polygons by splitting the LineString into
+* two-point segments and converting each segment into a polygon.
+*
```
indiejames, line 94: I'm not sure this works correctly. If I'm reading your segment to polygon code correctly, it creates a rectangle along the segment. When you put a bunch of these rectangles together the corners don't match up, so you have discontinuities. Maybe it's close enough, but I don't know - probably depends on a case-by-case basis.

services/harmony/app/frontends/ogc-edr/get-data-for-trajectory.ts
```diff
@@ -0,0 +1,280 @@
...
+  return wktPolygon;
+}
+
+/**
+ * Removes unnecessary spaces between LineStrings in a WKT MULTILINESTRING.
+ *
+ * @param wktMultiLineString - The WKT MULTILINESTRING to process.
+ * @returns The WKT MULTILINESTRING with the spaces removed.
+ */
+function cleanMultiLineString(wktMultiLineString: string): string {
```
indiejames, line 162: nit: Not sure it's worth having an entire function that just invokes a single regex.

services/harmony/app/frontends/ogc-edr/get-data-for-trajectory.ts
ygliuvt: Updated.

services/harmony/app/frontends/ogc-edr/get-data-for-trajectory.ts
ygliuvt: The LINESTRING to MULTIPOLYGON conversion is a simulation. The surrounding polygon doesn't have to be continuous especially when the segments within the LINESTRING are contained in the MULTIPOLYGON.

services/harmony/app/frontends/ogc-edr/get-data-for-trajectory.ts
ygliuvt: Dropped this function.

indiejames (approved):
Tested locally. Jobs failed due to l2ss issues, but otherwise request was handled correctly.

services/harmony/app/frontends/ogc-edr/get-data-for-trajectory.ts
```diff
@@ -0,0 +1,233 @@
...
+    const polygon = convertWktLineToPolygon(query.coords);
+    try {
+      const geoJson = parseWkt(polygon);
+      if (geoJson) {
+        operation.geojson = JSON.stringify(geoJson);
+      }
+    } catch (e) {
+      if (e instanceof ParameterParseError) {
+        // Turn parsing exceptions into 400 errors pinpointing the source parameter
+        throw new ServerError(`LINESTRING/MULTILINESTRING coverted POLYGON/MULTIPOLYGON is invalid ${e.message}`);
```
chris-durbin, line 204: converted typo

services/harmony/app/frontends/ogc-edr/get-data-for-trajectory.ts
```diff
@@ -0,0 +1,233 @@
...
+  if (query.coords) {
+    const polygon = convertWktLineToPolygon(query.coords);
+    try {
+      const geoJson = parseWkt(polygon);
+      if (geoJson) {
+        operation.geojson = JSON.stringify(geoJson);
+      }
+    } catch (e) {
+      if (e instanceof ParameterParseError) {
+        // Turn parsing exceptions into 400 errors pinpointing the source parameter
```
chris-durbin, line 203: If we're throwing a server error it's a 500 error. If it's a parsing issue I think we want to throw a RequestValidationError which results in a 400 rather than a 500.

chris-durbin (approved):

services/harmony/app/frontends/ogc-edr/get-data-for-trajectory.ts
ygliuvt: I will update it to RequestValidationError.

PR #632, opened 2024-09-30
HARMONY-1895: Add labels to jobs listings and status pages

indiejames:
## Jira Issue ID
HARMONY-1895

## Description
Adds labels to jobs listings and status pages

## Local Test Steps
* run a job and pass in labels using the `label` paramater
* load the jobs listing page (may not show any jobs on localhost due to a bug Chris discovered) and verify that labels are given for the jobs
http://localhost:3000/jobs
* load the admin jobs listing page and verify labels are shown
http://localhost:3000/admin/jobs
* load the job status page and verify labels are shown
http://localhost:3000/jobs/<job ID>
* load the admin job status page and verify labels are shown
http://localhost:3000/admin/jobs<job ID>

## PR Acceptance Checklist
* [x] Acceptance criteria met

### DIFF
--- a/services/harmony/app/frontends/jobs.ts
+++ b/services/harmony/app/frontends/jobs.ts
@@ -15,7 +15,6 @@ import env from '../util/env';
 import JobError, { getErrorsForJob } from '../models/job-error';
 import _ from 'lodash';
 import { isAdminUser } from '../util/edl-api';
-import { getLabelsForJob } from '../models/label';
 
 /**
  * Returns true if the job contains S3 direct access links
@@ -134,10 +133,7 @@ export async function getJobsListing(
     }
     let listing;
     await db.transaction(async (tx) => {
-      listing = await Job.queryAll(tx, query, page, limit);
-      for (const job of listing.data) {
-        job.labels = await getLabelsForJob(tx, job.jobID);
-      }
+      listing = await Job.queryAll(tx, query, page, limit, true);
     });
     const serializedJobs = listing.data.map((j) => getJobForDisplay(j, root, 'none', []));
     const response: JobListing = {
--- a/services/harmony/app/frontends/jobs.ts
+++ b/services/harmony/app/frontends/jobs.ts
@@ -15,6 +15,7 @@ import env from '../util/env';
 import JobError, { getErrorsForJob } from '../models/job-error';
 import _ from 'lodash';
 import { isAdminUser } from '../util/edl-api';
+import { getLabelsForJob } from '../models/label';
 
 /**
  * Returns true if the job contains S3 direct access links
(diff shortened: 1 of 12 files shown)


services/harmony/app/frontends/jobs.ts
```diff
@@ -134,6 +135,9 @@ export async function getJobsListing(
     let listing;
     await db.transaction(async (tx) => {
       listing = await Job.queryAll(tx, query, page, limit);
+      for (const job of listing.data) {
```
vinnyinverso, line 138: I think it'd be nice to have this as an optional return field from queryAll so that we can reuse it (for the workflow ui for example). Also wondering why you decided to use a loop versus doing a join between jobs and labels?

indiejames: I started to add this to `queryAll`, but it would mean adding another parameter to the function call, which already has a lot of optional parameters. I didn't want to deal with having to update the call in all the places it's used. And I wasn't sure if labels would be useful in most places where jobs are pulled from the dB, so I would be adding a parameter just to accommodate a handful of uses. But I could be persuaded otherwise. 

A join would only make sense when retrieving the jobs, so I would have to add the join to the `knex` `select` happening in `queryAll` - may be worth trying. I'm kind of growing weary of the places where we build these queries and optionally add additional `knex` actions (like `join`) based on the function parameters. But maybe that's just me. Would like to hear from others.

indiejames: Alternatively, I could just query for the labels in bulk using the job IDs, instead of iterating over the jobs.

services/harmony/test/jobs/jobs-listing.ts
```diff
@@ -363,13 +381,19 @@ describe('Jobs listing route', function () {
...
+        expect(labels).deep.equal(expectedLabels);
+      });
     });
 
     describe('when the user is not part of the admin group', function () {
       hookAdminJobListing({ username: 'eve' });
       it('returns an error', function () {
         expect(this.res.statusCode).to.equal(403);
-        expect(this.res.text).to.include('You are not permitted to access this resource');
+        expect(this.res.text).to.include('You are not permitted to access this resource');7;
```
chris-durbin, line 396: Cat jumped on the keyboard.

services/harmony/app/frontends/jobs.ts
chris-durbin: I think we allow up to 2000 jobs so we definitely should remove the `for` loop querying one job at a time. I don't really have a strong opinion on whether using `includeLabels` as a parameter vs. a second query to grab the labels when you need them is better. I guess the pros are what Vinny mentioned about the worklow-ui just having to add `includeLabels=true` as well as only issuing one query.

vinnyinverso: > I started to add this to queryAll...
Now that I think about it some more, I may want to query for them separately anyway so that the UI loads faster--but either way it's ok.

indiejames: I changed it to use a double-join to (optionally) get the labels in `queryAll`.

services/harmony/test/jobs/jobs-listing.ts
chris-durbin: looks like this still here

indiejames: Fixed (for real)

PR #636, opened 2024-10-07
Das 1601 remove png jpeg from hga

owenlittlejohns:
## Jira Issue ID

DAS-1601 - preliminary work.

## Description

This PR takes the steps towards removing functionality from HGA in favour of HyBIG. HyBIG is now the preferred service to create browse imagery. DAS-1601 is a ticket to remove similar functionality from HGA. We'd held off doing that for a while, because ideally a chain of net2cog and HyBIG should exist to truly replace what we suspect HGA can do. That said, there are no true users of HGA, and very few collections associated with the service. (I've checked with PO.DAAC and they are not using HGA for browse image generation right now)

This PR removes PNG and GIF outputs from the capabilities of HGA. Once deployed, we can give it a few sprints to check people aren't missing the functionality (and point them at HyBIG if they are) and then actually rip out the relevant browse image code from HGA (the actual work for DAS-1601).

## Local Test Steps

* Pull this branch.
* Start a local Harmony-in-a-Box instance with `harmony-gdal-adapter` in the `LOCALLY_DEPLOYED_SERVICES` environment variable in `.env`.
* Run the latest version of the [HGA regression test suite](https://github.com/nasa/harmony-regression-tests/blob/main/test/hga/HGA_Regression.ipynb). (Because this is being run against your local instance, you'll have to run the tests from a local Jupyter notebook server)
* Note - if you have issues running things on a Mac with an M1 chip, you can use the updates in [this branch of HGA](https://github.com/nasa/harmony-gdal-adapter/pull/32) to build a local version of the HGA image that should run on your machine.
* Second note - both Matt and I were finding the latest version of Harmony-in-a-Box to be a bit flaky.

## PR Acceptance Checklist
* ~~Acceptance criteria met~~ - this is preliminary to ensure we can rip out the code for the ticket.
* [ x] Tests added/updated (if needed) and passing

### DIFF
--- a/services/harmony/test/collection-capabilities.ts
+++ b/services/harmony/test/collection-capabilities.ts
@@ -74,7 +74,7 @@ describe('Testing collection capabilities', function () {
         it('sets the outputFormats field correctly', function () {
           const capabilities = JSON.parse(this.res.text);
           const expectedFormats = [
-            'application/x-netcdf4', 'image/tiff', 'image/png', 'image/gif', 'application/x-zarr',
+            'application/x-netcdf4', 'image/tiff', 'application/x-zarr', 'image/png', 'image/gif',
           ];
           expect(capabilities.outputFormats).to.eql(expectedFormats);
         });
@@ -235,7 +235,7 @@ describe('Testing collection capabilities', function () {
         it('sets the outputFormats field correctly in the version 1 response', function () {
           const capabilities = JSON.parse(this.res.text);
           const expectedFormats = [
-            'application/x-netcdf4', 'image/tiff', 'image/png', 'image/gif', 'application/x-zarr',
+            'application/x-netcdf4', 'image/tiff', 'application/x-zarr', 'image/png', 'image/gif',
           ];
           expect(capabilities.outputFormats).to.eql(expectedFormats);
         });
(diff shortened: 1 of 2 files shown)


services/harmony/test/collection-capabilities.ts
```diff
@@ -74,7 +74,7 @@ describe('Testing collection capabilities', function () {
         it('sets the outputFormats field correctly', function () {
           const capabilities = JSON.parse(this.res.text);
           const expectedFormats = [
-            'application/x-netcdf4', 'image/tiff', 'image/png', 'image/gif', 'application/x-zarr',
+            'application/x-netcdf4', 'image/tiff', 'application/x-zarr', 'image/png', 'image/gif',
```
owenlittlejohns, line 77: Just a quick explanation on these updated tests:

Harmony concatenates all the output formats into a single list. The ordering of the list reflects the first time each output is encountered in `services.yml`. HGA is listed above NetCDF-to-Zarr, which in turn is above harmony-service-example.

Before this change, HGA and harmony-service-example both listed PNGs and GIFs as output options, so those formats were listed before Zarr, because HGA is before NetCDF-to-Zarr. After this change, only harmony-service-example lists PNG and GIF, so these formats are found after NetCDF-to-Zarr and listed after the Zarr output option.

Kind of in-the-weeds, but wanted to explain what looks like a weird test change.

flamingbear (approved):
This looks right to me.

chris-durbin (approved):
Verified in production that it has been over a year since the last time anyone requested one of those formats from the harmony-gdal-adapter service so this change should not impact anyone.

PR #637, opened 2024-10-08
Harmony 1891 - Display job labels in Workflow UI

vinnyinverso:
## Jira Issue ID
HARMONY-1891

## Description
Users will now see labels for jobs on all pages of the Workflow UI.

On the jobs page, each row will contain the job's label(s) on the bottom right corner of the row.

Screenshot 2024-10-08 at 11 12 21 AM

On the individual job page, a similar view will be displayed, which will also include the request URL, which I had intended to add in HARMONY-1497. The look of this may evolve in the future but for now I'm mainly going for consistency.

Screenshot 2024-10-08 at 11 12 03 AM

## Local Test Steps
1. Make a request http://localhost:3000/C1233800302-EEDTEST/ogc-api-coverages/1.0.0/collections/all/coverage/rangeset?granuleId=G1233800343-EEDTEST&format=image%2Ftiff&label=hi&label=there
2. Look at the labels http://localhost:3000/workflow-ui?page=1&limit=10

## PR Acceptance Checklist
* [x] Acceptance criteria met

### DIFF
--- a/services/harmony/app/frontends/workflow-ui.ts
+++ b/services/harmony/app/frontends/workflow-ui.ts
@@ -568,6 +568,7 @@ export async function getWorkItemsTable(
           .replace('/work-items', '')
           .replace(/(&|\?)checkJobStatus=(true|false)/, '') : '');
       },
+      ...jobRenderingFunctions(req.context.logger, requestQuery),
     });
   } catch (e) {
     req.context.logger.error(e);
--- a/services/harmony/app/frontends/workflow-ui.ts
+++ b/services/harmony/app/frontends/workflow-ui.ts
@@ -278,7 +278,7 @@ function tableQueryToJobQuery(tableQuery: TableQuery, isAdmin: boolean, user: st
     };
   }
   if (tableQuery.from || tableQuery.to) {
-    jobQuery.dates = { field: tableQuery.dateKind };
+    jobQuery.dates = { field: `jobs.${tableQuery.dateKind}` };
     jobQuery.dates.from = tableQuery.from;
     jobQuery.dates.to = tableQuery.to;
   }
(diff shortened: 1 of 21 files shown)


services/harmony/app/frontends/workflow-ui.ts
```diff
@@ -198,6 +198,12 @@ function jobRenderingFunctions(logger: Logger, requestQuery: Record<string, any>
         return this.request;
       }
     },
+    jobLabelsDisplay(): string {
+      return this.labels.map((label) => {
```
indiejames, line 202: I wonder if we should sort the labels here?

indiejames (approved):

indiejames:
Tested locally

chris-durbin (approved):
Tested with many labels and Looks good - just the one comment about displaying labels in alphabetical order.

ygliuvt (approved):

services/harmony/app/frontends/workflow-ui.ts
vinnyinverso: Regarding label sorting -- this is where we assign the labels property in `queryAll()` :
```
 const jobs: Job[] = items.data.map((j: JobWithLabels) => {
      const job = new Job(j);
      if (includeLabels && j.label_values) {
        job.labels = j.label_values.split(',');
      } else {
        job.labels = [];
      }
      return job;
    });
```
I'm thinking it's safe to assume that if you've set i`ncludeLabels=true` in `queryAll` that you're going to display the labels. So rather than add a new parameter (`sortLabels`), I think we should just sort alphabetically here by default. However, the runtime might not be great (sorting an array for each job returned), but I'm not sure there's a better alternative unless we restructure the join query somehow to include an alphabetical sort.

vinnyinverso: `string_agg` has an order by option, so perhaps that'll do the trick.

vinnyinverso: That worked.

services/harmony/app/models/job.ts
```diff
@@ -330,6 +330,23 @@ async function getUniqueProviderIds(tx: Transaction): Promise<string[]> {
   return results.map((job) => job.provider_id);
 }
 
+/**
+ * Sets the fields on the where clauses (see JobQuery) to be prefixed with a table name to avoid
+ * ambiguities when joining with other tables
+ * @param query - the where clauses to process
```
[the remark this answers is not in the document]
vinnyinverso: Fixed

PR #640, opened 2024-10-10
Harmony 1897

indiejames:
## Jira Issue ID
HARMONY-1897

## Description
Adds support for adding labels to jobs as well as removing labels from jobs.

## Local Test Steps
1. Run the migrations
2. Verify that any previous labels you added to jobs still show up in the job status
3. Add labels to a job:
```
curl  -XPUSH http://localhost:3000/labels -d '{"jobID": ["<YOUR JOB ID>"], "label": ["foo", "bar", "foo", "buzz"]}'  -H "content-type: application/json" -H "Authorization: bearer $MY_UAT_TOKEN"
```
4. Verify that the labels are added using the workflow-ui or job status
5. Try some other labels with leading/trailing spaces and different casing
6. Delete labels from the job
```
curl  -XDELETE http://localhost:3000/labels -d '{"jobID": ["<YOUR JOB ID>"], "label": ["foo", "bar"]}'  -H "content-type: application/json" -H "Authorization: bearer $MY_UAT_TOKEN"
```
7. Verify that the labels are removed using the workflow-ui or job status
8. Read the documentation for labels in the online documentation page

## PR Acceptance Checklist
* [x] Acceptance criteria met


chris-durbin (commented):
Tested out in sandbox with a workload run going with many jobs most of which had labels and some without labels. Migrations all worked with out issue. Most of my testing worked, but had an issue with this label:
`"one more 1234\&\&"`

It returned an error:
```
{"code":"harmony.ServerError","description":"Error: Bad escaped character in JSON at position 183 (line 1 column 184)"}%
```

services/harmony/app/markdown/labels.md
```diff
@@ -0,0 +1,18 @@
+### <a name="job-labels"></a>  Labeling Jobs
+
+Labels can be applied to a job when a Harmony request is made using the `label` parameter documented
+in the [Services API section](#using-the-service-apis). After a request is made the labels can be viewed in the
+job status page and in the workflow-ui.
+
+Labels can be added to existing jobs by the job owner, or anyone with admin permissions, using an HTTP PUT request and specifying the job IDs and labels in the body of the PUT. An EDL bearer token must be provided and a `Content-Type: application/json` header. A `curl` example that adds two labels to two different jobs follows:
+
+```
+curl -bj {{root}} -XPUT -d '{"jobID": ["<YOUR FIRST JOB ID>", "<YOUR SECOND JOB ID>"], "label": ["foo", "bar"]}'  -H "Content-Type: application/json" -H "Authorization: bearer <YOUR BEARER TOKEN>"
```
ygliuvt, line 10: The route should be `{{root}}/labels`.

services/harmony/app/markdown/labels.md
```diff
@@ -0,0 +1,18 @@
...
+```
+
+Similarly, labels can be removed from one or more jobs using an HTTP DELETE:
+
+```
+curl -bj {{root}} -XDELETE -d '{"jobID": ["<YOUR FIRST JOB ID>", "<YOUR SECOND JOB ID>"], "label": ["foo"]}'  -H "Content-Type: application/json" -H "Authorization: bearer <YOUR BEARER TOKEN>"
```
ygliuvt, line 16: The route should be `{{root}}/labels`.

services/harmony/app/models/label.ts
```diff
@@ -27,22 +31,97 @@ export function normalizeLabel(label: string): string {
   return label.trim().toLowerCase();
 }
 
+/**
+ * Verify that the user can change the labels on a give job. Currently only job owners can
```
ygliuvt, line 35: job owners and admin can change labels for a job.

ygliuvt:
For discussion purpose only: Not keeping the user labels consistent with the real state may introduce subtle unwanted behavior. For example, Say I labeled job #1 with label `investigate` to look into it. After I am done with investigation, I removed the `investigate` label. On workflow-ui, I don't see job #1 as needing investigation (i.e. without `investigate` label). Chris then decided he wants to look into job #1 as well and he labeled job #1 with the `investigate` label. Then all of a sudden, from my point of view, the job is labeled with `investigate` again on workflow-ui even though I don't want to and shouldn't see the `investigate` label on the job.

indiejames:
> For discussion purpose only: Not keeping the user labels consistent with the real state may introduce subtle unwanted behavior. For example, Say I labeled job #1 with label `investigate` to look into it. After I am done with investigation, I removed the `investigate` label. On workflow-ui, I don't see job #1 as needing investigation (i.e. without `investigate` label). Chris then decided he wants to look into job #1 as well and he labeled job #1 with the `investigate` label. Then all of a sudden, from my point of view, the job is labeled with `investigate` again on workflow-ui even though I don't want to and shouldn't see the `investigate` label on the job.

We (well, at least Chris and I) talked about whether or not users should be able to see all labels or just the ones they added, and the decision was that they should see all of them. I originally thought it should be per-user, but now I agree that just having one consistent view is the most straight-forward approach.

services/harmony/app/markdown/labels.md
indiejames: Good catch, fixed.

services/harmony/app/models/label.ts
indiejames: fixed

services/harmony/app/markdown/labels.md
indiejames: fixed

services/harmony/app/models/label.ts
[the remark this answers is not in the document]
indiejames: fixed

services/harmony/app/frontends/labels.ts
```diff
@@ -0,0 +1,61 @@
...
+ * @returns Resolves when the request is complete
+ */
+export async function addJobLabels(
+  req: HarmonyRequest, res: Response, next: NextFunction,
+): Promise<void> {
+  const isAdmin = await isAdminUser(req);
+
+  try {
+    await db.transaction(async (trx) => {
+      await addLabelsToJobs(trx, req.body.jobID, req.user, req.body.label, isAdmin);
```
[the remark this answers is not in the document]
indiejames: I wasn't sure about this, but I ended up using singular, because `label` is consistent with the `label` parameter we use with coverage and EDR requests (which was made singular to be consistent with `granuleId`, like you said). It would seem strange to me to use `labels` here, but `label` there.

indiejames:
> Tested out in sandbox with a workload run going with many jobs most of which had labels and some without labels. Migrations all worked with out issue. Most of my testing worked, but had an issue with this label: `"one more 1234\&\&"`
> 
> It returned an error:
> 
> ```
> {"code":"harmony.ServerError","description":"Error: Bad escaped character in JSON at position 183 (line 1 column 184)"}%
> ```

I wrote HARMONY-1925 to address this (handling invalid JSON) separately.

indiejames:
> > > For discussion purpose only: Not keeping the user labels consistent with the real state may introduce subtle unwanted behavior. For example, Say I labeled job #1 with label `investigate` to look into it. After I am done with investigation, I removed the `investigate` label. On workflow-ui, I don't see job #1 as needing investigation (i.e. without `investigate` label). Chris then decided he wants to look into job #1 as well and he labeled job #1 with the `investigate` label. Then all of a sudden, from my point of view, the job is labeled with `investigate` again on workflow-ui even though I don't want to and shouldn't see the `investigate` label on the job.
> > 
> > 
> > We (well, at least Chris and I) talked about whether or not users should be able to see all labels or just the ones they added, and the decision was that they should see all of them. I originally thought it should be per-user, but now I agree that just having one consistent view is the most straight-forward approach.
> 
> This being the case, shouldn't we remove the `users_labels` table? We could just add the user back to the `raw_labels` table.

No, I separated out the functionality of tracking labels for jobs from tracking labels for users, because it was not possible to add a database constraint to prevent two users from adding the same label value to a job (same value, different entries in the `labels` table).

ygliuvt:
> > > > For discussion purpose only: Not keeping the user labels consistent with the real state may introduce subtle unwanted behavior. For example, Say I labeled job #1 with label `investigate` to look into it. After I am done with investigation, I removed the `investigate` label. On workflow-ui, I don't see job #1 as needing investigation (i.e. without `investigate` label). Chris then decided he wants to look into job #1 as well and he labeled job #1 with the `investigate` label. Then all of a sudden, from my point of view, the job is labeled with `investigate` again on workflow-ui even though I don't want to and shouldn't see the `investigate` label on the job.
> > > 
> > > 
> > > We (well, at least Chris and I) talked about whether or not users should be able to see all labels or just the ones they added, and the decision was that they should see all of them. I originally thought it should be per-user, but now I agree that just having one consistent view is the most straight-forward approach.
> > 
> > 
> > This being the case, shouldn't we remove the `users_labels` table? We could just add the user back to the `raw_labels` table.
> 
> No, I separated out the functionality of tracking labels for jobs from tracking labels for users, because it was not possible to add a database constraint to prevent two users from adding the same label value to a job (same value, different entries in the `labels` table).

OK. I keep forgetting that you need the users for a label for the atuo-complete in workflow ui.

ygliuvt (approved):

chris-durbin (approved):

PR #648, opened 2024-10-29
HARMONY-1901: Search for jobs by label

chris-durbin:
## Jira Issue ID
HARMONY-1901

## Description
Adds a query parameter `label` on the /jobs and /admin/jobs endpoint that allows filtering jobs based on the provided label(s). If multiple labels are provided the jobs returned will be a logical OR of all the provided labels (return all jobs that contain any of the labels).

## Local Test Steps
* Submit a few jobs with multiple labels in each job.
* Test /jobs?label=<comma separated list of labels> as well as /admin/jobs.
* Verify that jobs are filtered correctly based on the provided labels.
* Verify that each job returned has all the expected fields for the jobs listing endpoint and that the labels field contains all of the labels for each job (not just the label that was being searched for).

## PR Acceptance Checklist
* [X] Acceptance criteria met


services/harmony/app/models/job.ts
```diff
@@ -500,16 +501,39 @@ export class Job extends DBRecord implements JobRecord {
...
-        .select(`${Job.table}.*`, tx.raw(`STRING_AGG(${LABELS_TABLE}.value, ',' order by value) AS label_values`))
-        .leftOuterJoin(`${JOBS_LABELS_TABLE}`, `${Job.table}.jobID`, '=', `${JOBS_LABELS_TABLE}.job_id`)
-        .leftOuterJoin(`${LABELS_TABLE}`, `${JOBS_LABELS_TABLE}.label_id`, '=', `${LABELS_TABLE}.id`)
-        .where(setTableNameForWhereClauses(Job.table, constraints.where))
-        .groupBy(`${Job.table}.id`)
-        .orderBy(
-          constraints?.orderBy?.field ?? 'createdAt',
-          constraints?.orderBy?.value ?? 'desc')
-        .modify((queryBuilder) => modifyQuery(queryBuilder, constraints));
+      if (constraints.labels) { // Requesting to limit the jobs based on the provided labels
```
indiejames, line 504: Is it possible that someone might want to filter on labels, but doesn't need the labels included in the result? I think these are separate concerns.

indiejames (commented):
Just a couple of comments.

services/harmony/app/models/job.ts
chris-durbin: There is currently not a use case for it or code that would request to not include labels but does filter on labels. If we do need that functionality at some point later we should add the code in then to handle that case too.

services/harmony/app/frontends/jobs.ts
```diff
@@ -126,10 +127,17 @@ export async function getJobsListing(
     req.context.logger.info(`Get jobs listing for user ${req.user}`);
     const root = getRequestRoot(req);
     const { page, limit } = getPagingParams(req, env.defaultJobListPageSize);
+    let labels = [];
```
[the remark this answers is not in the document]
chris-durbin: Nice, done.

indiejames (approved):
Tested  locally

services/harmony/app/models/job.ts
vinnyinverso: (Would need to add `includeLabels` as a parameter.)

vinnyinverso (approved):

PR #653, opened 2024-11-01
HARMONY-1918: Add support for harmony core services image versioning

ygliuvt:
## Jira Issue ID
HARMONY-1918

## Description
This PR adds support for Harmony core services versioning and changed initial .env setup to use the stable image. It is only a part of the changes to support [this feature](https://wiki.earthdata.nasa.gov/display/HARMONY/Harmony+Micro-service+Image+Tags). The other changes are in `harmony-ci-cd` and `harmony-regression-tests` projects. 

I have set up the test in SIT (we will eventually turn things on in UAT instead) to show the feature end to end.
You can see the set up in CICD harmony and harmony-regression-tests deployment plans. For end result, go to the DockerHub Harmony core services repositories to verify that the image with tags `harmony-1918-8` (release number), `sit` and `stable` exists.

## Local Test Steps
Verify at least the following:
1.`bin/bootstrap-harmony` still works without adding the IMAGE overrides in .env, and the `stable` tag of the images are used for Harmony core services. Requests are processed successfully. 
2. Add the following image overrides in .env, run bootstrap again and verify the `latest` tag of the images are used for Harmony core services. Requests are processed successfully. 
```
HARMONY_IMAGE=harmonyservices/harmony:latest
QUERY_CMR_IMAGE=harmonyservices/query-cmr:latest
QUERY_CMR_SERVICE_QUEUE_URLS='["harmonyservices/query-cmr:latest,http://sqs.us-west-2.localhost.localstack.cloud:4566/000000000000/query-cmr.fifo"]'
SERVICE_RUNNER_IMAGE=harmonyservices/service-runner:latest
WORK_ITEM_SCHEDULER_IMAGE=harmonyservices/work-scheduler:latest
WORK_ITEM_UPDATER_IMAGE=harmonyservices/work-updater:latest
WORK_REAPER_IMAGE=harmonyservices/work-reaper:latest
WORK_FAILER_IMAGE=harmonyservices/work-failer:latest
```
3. Feel free to verify the release tag and sit tag images similar to step 2 if you want.

## PR Acceptance Checklist
* [x] Acceptance criteria met


bin/create-dotenv
```diff
@@ -61,4 +61,14 @@ EOF
         echo -e "\nLOCALLY_DEPLOYED_SERVICES=harmony-service-example,giovanni-adapter\n" >> .env
     fi
 
+    # set Harmony core services to use the stable image by default for harmony-in-a-box
+    echo "HARMONY_IMAGE=harmonyservices/harmony:stable" >> .env
```
chris-durbin, line 65: Generally we set all the default values to work with harmony in a box (so we would set it to use `stable` for everything and then override if we want to change any of the images to latest).

I think that makes the most sense in this case as well to have the defaults as stable, but we can maybe talk about it at the tech meeting tomorrow.

services/harmony/bin/prebuild
```diff
@@ -3,11 +3,27 @@
...
 #
 
-rimraf built/*
-copyfiles -u 2 -e "../../db/*.sqlite3" "../../db/**/*" built
-# need to copy the db directory into the servcies/harmony directory as well to make knex migrations work because knex is stupid
-copyfiles -u 2 -e "../../db/*.sqlite3" "../../db/**/*" built/services/harmony
-copyfiles -u 2 "../../bin/start-harmony-in-container" built
-copyfiles -u 2 "../../config/services.yml" built
-copyfiles -u 4 -e "../../packages/util/node_modules/**/*" -e "../../packages/util/built/**/*" -e "../../packages/util/test/**/*" -e "../../packages/util/coverage/**/*" "../../packages/util/**/*"  built/packages/util
-copyfiles -e "./test/**/*" -e "./node_modules/**/*" -e "./coverage/**/*" -e "./built/**/*" -e "./bin/**/*" -e "./fixtures/**/*" "./**/*" built/services/harmony
```
chris-durbin, line 13: If we're no longer using copyfiles anywhere I think there might be a dependency we can drop from listing in our documentation.

ygliuvt: Other core services are still using the copyfiles in their prebuild script. I don't want to do too much unrelated stuff in this PR. We can write a new ticket and do the consolidation separately.

chris-durbin (approved):
Tested successfully locally using both harmony in a box and local development with the image overrides to use latest. Everything worked as expected with harmony in a box using the stable images. I also deployed to a sandbox environment and everything worked as expected with ECR images being used when no environment variables are overridden. I also verified I could override the image locations and everything worked as expected - changing the query-cmr image used the service deployment endpoint and all of the other harmony images were updated as part of the main harmony deployment.

indiejames (approved):

PR #655, opened 2024-11-06
Harmony 1896 - job filtering by label

indiejames:
## Jira Issue ID
HARMONY-1896

## Description
Adds the ability to filter the jobs list using labels

## Local Test Steps
1. Make some Harmony requests with various labels
2. Load the workfow-ui page
3. In the filter input start typing "label". You should a list of labels in the auto-complete (no more than ten)
4. Choose one of the labels.
5. Hit "apply". You should see your jobs only the jobs list only showing jobs with that label.
6. Add another label and hit "apply". You should see jobs that have either of the labels.

## PR Acceptance Checklist
* [x] Acceptance criteria met


ygliuvt:
I am able to run the test in Sandbox and verified filtering by label within the top 10 works as expected. One issue I find is that the label filtering does not allow user to filter on a label that is outside of the top 10 most recent labels. It is pretty counter intuitive when a user can see a label on the workflow ui, but won't be able to filter on it just because it is not one of the 10 most recently used. I think we should allow user to filter on any labels he types in.

indiejames:
> I am able to run the test in Sandbox and verified filtering by label within the top 10 works as expected. One issue I find is that the label filtering does not allow user to filter on a label that is outside of the top 10 most recent labels. It is pretty counter intuitive when a user can see a label on the workflow ui, but won't be able to filter on it just because it is not one of the 10 most recently used. I think we should allow user to filter on any labels he types in.

You should be able to type in any label, even if it doesn't show up in the auto-complete. If not, then that is an error

vinnyinverso (approved):

ygliuvt (approved):
Tested successfully in Sandbox.

PR #657, opened 2024-11-13
HARMONY-1956: Remove duplicates from label list

indiejames:
## Jira Issue ID
HARMONY-1956

## Description
Fixes a bug where duplicate labels were showing up in the dropdown filter list when using the admin route.

## Local Test Steps
* Create a job with labels
* Use dbViz to get the id for one of the labels, e.g.,`foo`:
```
select id from raw_labels where value='foo'
```
* Use dbViz to add a row to the `users_labels` with a new username pointing to an existing label, e.g., `edltest` and `foo`
```
insert into users_labels values(<id from above select>, 'edltest', 'foo', now(), now());
```
`edltest` does not have to be a real user.
* Open http://localhost:3000/admin/workflow-ui and start typing `label` in the filter box. You should to see the duplicate label more than once.

## PR Acceptance Checklist
* [x] Acceptance criteria met

### DIFF
--- a/services/harmony/app/models/label.ts
+++ b/services/harmony/app/models/label.ts
@@ -246,7 +246,8 @@ export async function getLabelsForUser(
 ): Promise<string[]> {
   const query = trx(`${USERS_LABELS_TABLE}`)
     .select('value')
-    .orderBy('updatedAt', 'desc')
+    .groupBy('value')
+    .orderByRaw(`max("${USERS_LABELS_TABLE}"."updatedAt") desc`)
     .limit(count)
     .modify((queryBuilder) => {
       if (!isAdminRoute) {
--- a/services/harmony/test/labels/label_crud.ts
+++ b/services/harmony/test/labels/label_crud.ts
@@ -18,7 +18,7 @@ describe('Get Labels', function () {
     const trx = await db.transaction();
     await joeJob.save(trx);
     await jillJob.save(trx);
-    trx.commit();
+    await trx.commit();
   });
 
   describe('When getting labels using the admin route', function () {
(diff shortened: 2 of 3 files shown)


ygliuvt (approved):
Tested successfully locally.

services/harmony/app/models/label.ts
```diff
@@ -246,7 +246,8 @@ export async function getLabelsForUser(
 ): Promise<string[]> {
   const query = trx(`${USERS_LABELS_TABLE}`)
     .select('value')
-    .orderBy('updatedAt', 'desc')
+    .groupBy('value')
```
chris-durbin, line 249: We'll probably want to run another performance test at some point with a large number of labels though not needed prior to merging this in.

services/harmony/test/labels/label_crud.ts
```diff
@@ -1,12 +1,43 @@
...
+  hookServersStartStop({ skipEarthdataLogin: false });
+  before(async function () {
+    await truncateAll();
+    const trx = await db.transaction();
+    await joeJob.save(trx);
+    await jillJob.save(trx);
+    await trx.commit();
+  });
+
+  describe('When getting labels using the admin route', function () {
```
chris-durbin, line 24: Seems like one more test should be added to sanity check the order is based on the most recent time a label has been used by any user.

chris-durbin (approved):

services/harmony/test/labels/label_crud.ts
indiejames: Done

services/harmony/app/models/label.ts
indiejames: I can use the code I wrote to do the last test for that

PR #666, opened 2024-12-02
HARMONY-1951: Ensure work failer can run every few minutes without causing problems for the work item update queue.

ygliuvt:
## Jira Issue ID
HARMONY-1951

## Description
Ensure work failer can run every few minutes without causing problems for the work item update queue.

## Local Test Steps
Tested in Sandbox to run a 110k request and verify that work failer will only be run when the work items on the small work item queue is smaller than the configured MAX_WORK_ITEMS_ON_UPDATE_QUEUE_FAILER

## PR Acceptance Checklist
* [x] Acceptance criteria met

### DIFF
--- a/services/work-failer/app/workers/failer.ts
+++ b/services/work-failer/app/workers/failer.ts
@@ -3,6 +3,8 @@ import { JobStatus } from '../../../harmony/app/models/job';
 import WorkItem, { computeWorkItemDurationOutlierThresholdForJobService, getWorkItemsByUpdateAgeAndStatus } from '../../../harmony/app/models/work-item';
 import db from '../../../harmony/app/util/db';
 import log from '../../../harmony/app/util/log';
+import { WorkItemQueueType } from '../../../harmony/app/util/queue/queue';
+import { getQueueForType } from '../../../harmony/app/util/queue/queue-factory';
 import sleep from '../../../harmony/app/util/sleep';
 import { Worker } from '../../../harmony/app/workers/worker';
 import { WorkItemStatus } from '../../../harmony/app/models/work-item-interface';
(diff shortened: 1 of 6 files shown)


services/work-failer/app/workers/failer.ts
```diff
@@ -73,16 +75,30 @@ export default class Failer implements Worker {
...
+      if (env.maxWorkItemsOnUpdateQueueFailer !== -1) {
+        const smallUpdateQueue = getQueueForType(WorkItemQueueType.SMALL_ITEM_UPDATE);
+        const updateQueueCount = await smallUpdateQueue.getApproximateNumberOfMessages();
+        if (updateQueueCount > env.maxWorkItemsOnUpdateQueueFailer) {
+          log.warn(`Work item update queue is too large with ${updateQueueCount} items, will not fail more work`);
+          await sleep(failerDisabledDelay);
+          continue;
+        }
+      }
       const { workItems, jobServiceThresholds, maxId: newId } = await this.getExpiredWorkItems(lastUpdateOlderThanMinutes, startingId, batchSize);
```
chris-durbin, line 102: I think we should make one more change to set the batchSize to the minimum of `env.workFailerBatchSize` and the number of items that can still be added to the queue.

For example if there are 997 items on the queue with a limit of 1000 and the `env.workFailerBatchSize` is 1000 we should pass in 3 rather than 1000 to the call to `getExpiredWorkItems` so we only fail 3 more things.

ygliuvt: Updated. Thanks!

chris-durbin (approved):
Tested this successfully in a sandbox environment. I simulated slow work-item updates, purged messages from the work-item update queue, and ran multiple deployments. I had the work failer running every 2 minutes and it processed everything as expected (including not adding too many items to the queue when it was backed up).

indiejames (approved):

PR #674, opened 2024-12-12
HARMONY-1966: Fix Destination URL path is lowercased for harmony-job-status-link

ygliuvt:
## Jira Issue ID
HARMONY-1966, HARMONY-1967

## Description
Fix Destination URL path is lowercased for harmony-job-status-link.
Fix new nanoid vulnerability.

## Local Test Steps
Run Harmony locally, submit the following request, verify the harmony-job-status-link is generated in `s3://dummy/bignbit-harmony-output/OPERA_L3_DSWX-S1_V1/20241212` together with other outputs.

http://localhost:3000/C1264760805-POCLOUD/ogc-api-coverages/1.0.0/collections/all/coverage/rangeset?forceAsync=true&destinationUrl=s3%3A%2F%2Fdummy%2Fbignbit-harmony-output%2FOPERA_L3_DSWX-S1_V1%2F20241212&granuleId=G1272632541-POCLOUD&width=720&height=360&format=image%2Fpng

## PR Acceptance Checklist
* [x] Acceptance criteria met


chris-durbin (approved):

services/harmony/test/destination-url.ts
```diff
@@ -72,15 +72,15 @@ describe('when setting destinationUrl on ogc request', function () {
...
+      const s3Url = 's3://dummy/P1/' + jobId + '/harmony-job-status-link';
       const statusLink = await defaultObjectStore().getObject(s3Url);
       // this.res.request.url is the job status link
       expect(statusLink).to.equal(this.res.request.url);
     });
 
     it('sets the destination_url on the job in db', async function () {
       const retrieved = await Job.forUser(this.trx, 'anonymous');
-      expect(retrieved.data[0].destination_url).to.eq('s3://dummy/p1');
+      expect(retrieved.data[0].destination_url).to.eq('s3://dummy/P1');
```
indiejames, line 83: nit: I think when you are fixing a bug like this it makes more sense to add a separate test just for that bug, rather than co-opting an existing test. In the future someone might change this URL to be all lower-case for some other reason, because it's not obvious that this is also verifying that case is preserved.

indiejames (changes requested):
Just had one comment about adding a separate test

services/harmony/test/destination-url.ts
ygliuvt: Our test is based on File store, not real s3 store. So this test will not test the real s3 scenario that is in this ticket. I changed the capitalization just indicate the the s3 key can be mixed case.

indiejames: I thought the bug that Frank reported was just what you fixed - that the whole URL was being converted to lower-case when it should not be (just the s3:// part). In which case you are testing that by using a mixed-case URL. I was just saying that it's not obvious that this is being verified by this test.

ygliuvt: It will not be verified by the test because we use file store for testing which is case insensitive. I don't think it is worth it to change the testing mock or add a unit test.

indiejames (approved):

services/harmony/test/destination-url.ts
ygliuvt: @indiejames Great observation that the test will fail in github as it runs on linux. I have added a separate test as you suggested. Thanks!

PR #677, opened 2024-12-17
Harmony 1960 - fix job counter refresh on job table reload

vinnyinverso:
## Jira Issue ID
HARMONY-1960

## Description
Screenshot 2024-12-18 at 8 36 44 AM

This fixes an issue where the job counter display (which shows how many jobs are selected) was not being updated whenever the jobs table was dynamically reloaded. (The table gets refreshed--without reloading the whole page--due to UI-triggered job status change actions/events in order to reflect the new state of the jobs once the status change has taken effect.)

The counter needs to be refreshed when the table is reloaded because some jobs that were selected may no longer be visible in the current page. E.g. if a user has a filter applied to only show running jobs, and they cancel one of those jobs, it will disappear from the current view. Thus the counter display should be decremented.

## Local Test Steps
1. Submit 4 requests:
http://localhost:3000/C1233800302-EEDTEST/ogc-api-coverages/1.0.0/collections/all/coverage/rangeset?format=image%2Ftiff
2. Pause 2 of them, and cancel 2 of them via the UI http://localhost:3000/workflow-ui.
3. Set the UI filter to show only paused and canceled jobs.
4. Select 1 paused job and one canceled job via the checkboxes. The job counter display in the top right corner should show "2 jobs".
5. Click the "resume" link in the top right corner.
6. The job counter display in the top right corner should now show "1 job", since the resumed job will disappear from the table.

## PR Acceptance Checklist
* [x] Acceptance criteria met
* [ ] Harmony in a Box tested? (if changes made to microservices)


chris-durbin (approved):
Tested successfully locally.

ygliuvt (approved):

PR #683, opened 2025-01-23
Harmony 1996: Add 'empty-result' status to WorkItem class and work-items table

indiejames:
## Jira Issue ID
HARMONY-1996

## Description
This PR adds a new status to WorkItem to represent when a service processed the work-item, but produced to data. This is to allow us to provide a better experience for users by not staging empty files.

## Local Test Steps
1. Use dbViz to update a work-item's status to 'empty-result' and verify that this is allowed. (This change will not be visible anywhere until further tickets are worked)
2. Run a Harmony request and verify that it goes through okay.
3. Check the status page and workflow UI to make sure the behavior has not changed

## PR Acceptance Checklist
* [x] Acceptance criteria met
* [ x] Harmony in a Box tested? (if changes made to microservices)

### DIFF
--- a/db/db.sql
+++ b/db/db.sql
@@ -47,7 +47,7 @@ CREATE TABLE `job_errors` (
     'warning'
   )
   ) not null default 'error',
-  `category` varchar(255) not null default 'generic',
+  `category` varchar(255),
   `createdAt` datetime not null,
   `updatedAt` datetime not null,
   FOREIGN KEY(jobID) REFERENCES jobs(jobID)
@@ -88,7 +88,7 @@ CREATE TABLE `work_items` (
   `scrollID` varchar(4096),
   `serviceID` varchar(255) not null,
   `status` varchar(255) check (`status` in ('ready', 'queued', 'running', 'successful', 'failed', 'canceled', 'warning')) not null,
-  `sub_status` varchar(255) check (`sub_status` in ('no-data')),
+  `sub_status` varchar(255),
   `stacCatalogLocation` varchar(255),
   `totalItemsSize` double precision not null default 0,
   `outputItemSizesJson` text,
--- a/db/db.sql
+++ b/db/db.sql
@@ -41,6 +41,18 @@ CREATE TABLE `job_errors` (
   `jobID` char(36) not null,
   `url` varchar(4096) not null,
   `message` varchar(4096) not null,
+  `level` varchar(255) check (
+  `level` in (
+    'error',
+    'warning'
+  )
+  ) not null default 'error',
+  `category` varchar(255) check (
+    `category` in (
+      'generic',
+      'no-data'
+    )
+  ) not null default 'generic',
   `createdAt` datetime not null,
   `updatedAt` datetime not null,
   FOREIGN KEY(jobID) REFERENCES jobs(jobID)
@@ -80,7 +92,8 @@ CREATE TABLE `work_items` (
   `workflowStepIndex` integer not null,
   `scrollID` varchar(4096),
   `serviceID` varchar(255) not null,
-  `status` text check (`status` in ('ready', 'queued', 'running', 'successful', 'failed', 'canceled', 'no-data')) not null,
+  `status` varchar(255) check (`status` in ('ready', 'queued', 'running', 'successful', 'failed', 'canceled', 'warning')) not null,
+  `subStatus` varchar(255) check (`subStatus` in ('no-data')),
   `stacCatalogLocation` varchar(255),
   `totalItemsSize` double precision not null default 0,
   `outputItemSizesJson` text,
@@ -183,9 +196,12 @@ CREATE INDEX jobs_username_idx ON jobs(jobID, username);
 CREATE INDEX job_links_jobID_idx ON job_links(jobID);
 CREATE INDEX job_links_jobID_id_idx ON job_links(jobID, id);
 CREATE INDEX job_errors_jobID_idx ON job_errors(jobID);
+CREATE INDEX job_errors_level_idx ON job_errors(level);
+CREATE INDEX job_errors_category_idx ON job_errors(category);
 CREATE INDEX work_items_jobID_idx ON work_items(jobID);
 CREATE INDEX work_items_serviceID_idx ON work_items(serviceID);
 CREATE INDEX work_items_status_idx ON work_items(status);
+CREATE INDEX work_items_subStatus_idx ON work_items(subStatus);
 CREATE INDEX workflow_steps_jobID_idx ON workflow_steps(jobID);
 CREATE INDEX workflow_steps_jobID_StepIndex_idx ON workflow_steps(jobID, stepIndex);
 CREATE INDEX workflow_steps_serviceID_idx ON workflow_steps(serviceID);
--- a/db/db.sql
+++ b/db/db.sql
@@ -47,12 +47,7 @@ CREATE TABLE `job_errors` (
     'warning'
   )
   ) not null default 'error',
-  `category` varchar(255) check (
-    `category` in (
-      'generic',
-      'no-data'
-    )
-  ) not null default 'generic',
+  `category` varchar(255) not null default 'generic',
   `createdAt` datetime not null,
   `updatedAt` datetime not null,
   FOREIGN KEY(jobID) REFERENCES jobs(jobID)
@@ -93,7 +88,7 @@ CREATE TABLE `work_items` (
   `scrollID` varchar(4096),
   `serviceID` varchar(255) not null,
   `status` varchar(255) check (`status` in ('ready', 'queued', 'running', 'successful', 'failed', 'canceled', 'warning')) not null,
-  `subStatus` varchar(255) check (`subStatus` in ('no-data')),
+  `sub_status` varchar(255) check (`sub_status` in ('no-data')),
   `stacCatalogLocation` varchar(255),
   `totalItemsSize` double precision not null default 0,
   `outputItemSizesJson` text,
@@ -201,7 +196,7 @@ CREATE INDEX job_errors_category_idx ON job_errors(category);
 CREATE INDEX work_items_jobID_idx ON work_items(jobID);
 CREATE INDEX work_items_serviceID_idx ON work_items(serviceID);
 CREATE INDEX work_items_status_idx ON work_items(status);
-CREATE INDEX work_items_subStatus_idx ON work_items(subStatus);
+CREATE INDEX work_items_sub_status_idx ON work_items(sub_status);
 CREATE INDEX workflow_steps_jobID_idx ON workflow_steps(jobID);
 CREATE INDEX workflow_steps_jobID_StepIndex_idx ON workflow_steps(jobID, stepIndex);
 CREATE INDEX workflow_steps_serviceID_idx ON workflow_steps(serviceID);
--- a/db/db.sql
+++ b/db/db.sql
@@ -80,7 +80,7 @@ CREATE TABLE `work_items` (
   `workflowStepIndex` integer not null,
   `scrollID` varchar(4096),
   `serviceID` varchar(255) not null,
-  `status` text check (`status` in ('ready', 'queued', 'running', 'successful', 'failed', 'canceled')) not null,
+  `status` text check (`status` in ('ready', 'queued', 'running', 'successful', 'failed', 'canceled', 'empty-result')) not null,
   `stacCatalogLocation` varchar(255),
   `totalItemsSize` double precision not null default 0,
   `outputItemSizesJson` text,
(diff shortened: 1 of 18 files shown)


db/db.sql
```diff
@@ -80,7 +92,8 @@ CREATE TABLE `work_items` (
   `workflowStepIndex` integer not null,
   `scrollID` varchar(4096),
   `serviceID` varchar(255) not null,
-  `status` text check (`status` in ('ready', 'queued', 'running', 'successful', 'failed', 'canceled')) not null,
+  `status` varchar(255) check (`status` in ('ready', 'queued', 'running', 'successful', 'failed', 'canceled', 'warning')) not null,
+  `subStatus` varchar(255) check (`subStatus` in ('no-data')),
```
ygliuvt, line 96: The column name should be`sub_status` to match the Postgres definition.

db/db.sql
```diff
@@ -183,9 +196,12 @@ CREATE INDEX jobs_username_idx ON jobs(jobID, username);
 CREATE INDEX job_links_jobID_idx ON job_links(jobID);
 CREATE INDEX job_links_jobID_id_idx ON job_links(jobID, id);
 CREATE INDEX job_errors_jobID_idx ON job_errors(jobID);
+CREATE INDEX job_errors_level_idx ON job_errors(level);
+CREATE INDEX job_errors_category_idx ON job_errors(category);
 CREATE INDEX work_items_jobID_idx ON work_items(jobID);
 CREATE INDEX work_items_serviceID_idx ON work_items(serviceID);
 CREATE INDEX work_items_status_idx ON work_items(status);
+CREATE INDEX work_items_subStatus_idx ON work_items(subStatus);
```
ygliuvt, line 204: The column name should be`sub_status` to match the Postgres definition.

db/db.sql
ygliuvt: The value is missing `null`.

db/db.sql
```diff
@@ -41,6 +41,18 @@ CREATE TABLE `job_errors` (
...
   `url` varchar(4096) not null,
   `message` varchar(4096) not null,
+  `level` varchar(255) check (
+  `level` in (
+    'error',
+    'warning'
+  )
+  ) not null default 'error',
+  `category` varchar(255) check (
+    `category` in (
```
vinnyinverso, line 51: Do we want to have to change this check every time a new category is added in service lib? In theory, if we're already adding the category via pull request in service lib, it has been validated already by the time it gets to Harmony. This is also missing the existing error categories like "Server", "Forbidden".

services/harmony/app/models/work-item.ts
```diff
@@ -51,6 +51,9 @@ export default class WorkItem extends Record implements WorkItemRecord {
   // The status of the operation - see WorkItemStatus
   status?: WorkItemStatus;
 
+  // The sub-status of the operation - see WorkItemSubStatus
+  subStatus?: WorkItemSubStatus;
```
ygliuvt, line 55: nit: I think we have coding standard to use underscores in new column names. i.e. `sub_status`. 
I commented above that the `sub_status` name is not consistent in db definition vs migration. If you want to go with `subStatus`, it is fine with me as long as we make it consistent.

db/migrations/20250122190038_add_warning_status_and_sub_status_to_work_items.js
```diff
@@ -0,0 +1,32 @@
+/**
+ * @param { import("knex").Knex } knex
+ * @returns { Promise<void> }
+ */
+exports.up = function (knex, Promise) {
+  return knex.schema.raw(`
+    ALTER TABLE "work_items"
+    DROP CONSTRAINT "work_items_status_check",
```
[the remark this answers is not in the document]
vinnyinverso: Same question for the "down" migration.

db/db.sql
indiejames: Fixed

db/db.sql
indiejames: Fixed

services/harmony/app/models/work-item.ts
indiejames: For some reason I thought it could be snake case in the dB, but camel case in the models. I see now that we are not doing  that (maybe it doesn't work). I'll make it snake case here.

db/migrations/20250122190038_add_warning_status_and_sub_status_to_work_items.js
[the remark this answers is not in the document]
indiejames: because there is already a constraint with that name. We can't update it, we have to drop it then add it.

db/db.sql
indiejames: That's a good point. I'll drop this constraint here as well as in the migration.

db/migrations/20250122190038_add_warning_status_and_sub_status_to_work_items.js
```diff
@@ -0,0 +1,32 @@
...
+    ADD CONSTRAINT "work_items_status_check"
+    CHECK (status IN ('ready', 'queued', 'running', 'successful', 'failed', 'canceled', 'warning')),
+    ADD COLUMN "sub_status" VARCHAR(255),
+    ADD CONSTRAINT "work_items_sub_status_check"
+    CHECK (sub_status IN (null, 'no-data'));
```
chris-durbin, line 13: I don't think we should have an enumeration on this field which I think we were calling "category" - we should allow any string and don't need this constraint.

db/migrations/20250124155544_add_level_and_category_to_job_errors.js
```diff
@@ -0,0 +1,32 @@
...
+exports.up = function (knex, Promise) {
+  return knex.schema.raw(`
+    ALTER TABLE "job_errors"
+    ADD COLUMN "level" VARCHAR(255) DEFAULT 'error' NOT NULL,
+    ADD CONSTRAINT "job_errors_level_check"
+    CHECK (level IN ('error', 'warning')),
+    ADD COLUMN "category" VARCHAR(255) DEFAULT 'generic' NOT NULL;
+
+    CREATE INDEX job_errors_level ON job_errors (level);
+    CREATE INDEX job_errors_category ON job_errors (category)
```
chris-durbin, line 14: I'm not sure if it matters, but for consistency should we have each statement end in a semi-colon?

db/migrations/20250124155544_add_level_and_category_to_job_errors.js
```diff
@@ -0,0 +1,32 @@
...
+/**
+ * @param { import("knex").Knex } knex
+ * @returns { Promise<void> }
+ */
+exports.down = function (knex) {
+  return knex.schema.raw(`
+    DROP INDEX job_errors_category;
+    DROP_INDEX job_errors_level;
+    ALTER TABLE "job_errors"
+    DROP CONSTRAINT "job_errors_category_check",
```
chris-durbin, line 27: I don't see a job_error_category_check constraint in the up migration - and I don't think we want one so assume this should be removed.

packages/util/tsconfig.build.json
```diff
@@ -10,7 +10,10 @@
       "allowJs": true,
       "noImplicitAny": false,
       "sourceMap": true,
-      "outDir": "built"
+      "outDir": "built",
+      "typeRoots": [
```
chris-durbin, line 14: Out of curiosity what does this fix - anything we should do differently?

chris-durbin (commented):
I have a general question about sub_status and category. I see in work_items we are calling the field sub_status and in job_errors we're calling it category. I believe they are the same thing - would it be better to use the same column name?

db/migrations/20250122190038_add_warning_status_and_sub_status_to_work_items.js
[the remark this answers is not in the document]
vinnyinverso: Thanks, that makes sense.

db/migrations/20250124155544_add_level_and_category_to_job_errors.js
indiejames: Done.

db/migrations/20250124155544_add_level_and_category_to_job_errors.js
indiejames: Done.

db/migrations/20250122190038_add_warning_status_and_sub_status_to_work_items.js
indiejames: Done.

packages/util/tsconfig.build.json
indiejames: I was having consistent failures from `tsc` complaining about not finding types for lots of packages. This tells it where to look. Not sure why no one else has seen these errors.

db/migrations/20250122190038_add_warning_status_and_sub_status_to_work_items.js
```diff
@@ -0,0 +1,30 @@
...
+ */
+exports.up = function (knex, Promise) {
+  return knex.schema.raw(`
+    ALTER TABLE "work_items"
+    DROP CONSTRAINT "work_items_status_check",
+    ADD CONSTRAINT "work_items_status_check"
+    CHECK (status IN ('ready', 'queued', 'running', 'successful', 'failed', 'canceled', 'warning')),
+    ADD COLUMN "sub_status" VARCHAR(255);
+
+    CREATE INDEX work_items_sub_status ON work_items (sub_status);
```
ygliuvt, line 13: nit: add `_index` to index name for consistency. i.e. `work_items_sub_status_index`.

db/migrations/20250124155544_add_level_and_category_to_job_errors.js
```diff
@@ -0,0 +1,31 @@
...
+ */
+exports.up = function (knex, Promise) {
+  return knex.schema.raw(`
+    ALTER TABLE "job_errors"
+    ADD COLUMN "level" VARCHAR(255) DEFAULT 'error' NOT NULL,
+    ADD CONSTRAINT "job_errors_level_check"
+    CHECK (level IN ('error', 'warning')),
+    ADD COLUMN "category" VARCHAR(255);
+
+    CREATE INDEX job_errors_level ON job_errors (level);
```
ygliuvt, line 13: same comment with index naming convention.

db/migrations/20250124155544_add_level_and_category_to_job_errors.js
```diff
@@ -0,0 +1,31 @@
...
+  `);
+};
+
+/**
+ * @param { import("knex").Knex } knex
+ * @returns { Promise<void> }
+ */
+exports.down = function (knex) {
+  return knex.schema.raw(`
+    DROP INDEX job_errors_category;
```
ygliuvt, line 24: I got `- syntax error at or near "DROP_INDEX"` error when running the down migration.

db/migrations/20250124155544_add_level_and_category_to_job_errors.js
```diff
@@ -0,0 +1,31 @@
...
+    DROP_INDEX job_errors_level;
+    ALTER TABLE "job_errors"
+    DROP COLUMN "category",
+    DROP CONSTRAINT "job_errors_level_check",
+    DROP COLUMN "level;"
```
ygliuvt, line 29: typo in `level;`

ygliuvt (changes requested):
Need to fix the down migration issue.

db/migrations/20250122190038_add_warning_status_and_sub_status_to_work_items.js
indiejames: Done

db/migrations/20250124155544_add_level_and_category_to_job_errors.js
indiejames: Done

db/migrations/20250124155544_add_level_and_category_to_job_errors.js
indiejames: Fixed

db/migrations/20250124155544_add_level_and_category_to_job_errors.js
indiejames: ugh - fixed here and in the other migration

indiejames:
> Need to fix the down migration issue.

Done

chris-durbin (approved):
Successfully tested the migrations and rollbacks. Verified requests continued to work as expected.

ygliuvt (approved):
Tested successfully.

PR #690, opened 2025-02-05
Ldds 528

cbrownstevenson:
## Jira Issue ID
Objective: https://bugs.earthdata.nasa.gov/browse/LDDS-528

## Description

- Add dimension/L1b band subsetting to subset-band-name
- Service chain subset-band-name in geoloco's service and turn on band subsetting

## PR Acceptance Checklist
* [x] Acceptance criteria met


cbrownstevenson:
I re-added variable subsetting to geoloco in CMR UAT but do not know how to re-kick off the test.

config/services.yml
```diff
@@ -652,12 +653,15 @@ https://cmr.uat.earthdata.nasa.gov:
...
          multiple_variable: true
+         dimension: true
        output_formats:
          - application/x-hdf
        reprojection: true
     steps:
       - image: !Env ${QUERY_CMR_IMAGE}
         is_sequential: true
+      - image: !Env ${SUBSET_BAND_NAME_IMAGE}
+        is_sequential: true
```
chris-durbin, line 664: I don't think this one should be marked as sequential. Sequential is for something like query-cmr with pages through CMR results and can possibly create additional work items for the same step as part of its process.

I believe you'll want to make this step optional though. For example if a user does not ask for variable subsetting I think you want to skip subset-band-name and go straight to Geoloco. I'm not sure what other operations subset-band-name performs besides variable subsetting.

Something like:
```
        conditional:
          exists: ['variableSubset']
```

Similarly are there ever cases where you would only want to perform Geoloco and not subset-band-name? You would want to add a conditional for it as well.

chris-durbin (commented):
Double-check the output from the compare-services script which is currently failing to make sure your UMM-S record and harmony configuration match.

Also I think you can remove the `ldds/subset-band-name` service chain. The Geoloco chain can support everything you need once you make the steps conditional based on the transformations requested by the user.

cbrownstevenson:
> Double-check the output from the compare-services script which is currently failing to make sure your UMM-S record and harmony configuration match.
> 
> Also I think you can remove the `ldds/subset-band-name` service chain. The Geoloco chain can support everything you need once you make the steps conditional based on the transformations requested by the user.

@chris-durbin I removed it before I learned how service chaining worked. I added variable subsetting back in after looking at the check results.

config/services.yml
cbrownstevenson: Does the variableSubset include dimension subsetting? Or do I add something like dimensionSubset?

cbrownstevenson: Never mind. Found it.

chris-durbin (approved):
Verified that with http://localhost:3000/C1260442414-LAADSCDUAT/ogc-api-coverages/1.0.0/collections/EV_500_RefSB/coverage/rangeset?maxResults=1&subset=lon(-157.5%3A-155)&subset=lat(70%3A72.5) the request now first goes to subset-band-name prior to being sent to the geoloco service.

ygliuvt (approved):

PR #698, opened 2025-02-13
HARMONY-2007: Add is_sequential: true where missing for query-cmr in services.yml and validations for same

indiejames:
Add is_sequential: true where missing for query-cmr in services.yml and validation for same

## Jira Issue ID
HARMONY-2007

## Description
This fixes a configuration error where some query-cmr steps did not have `sequential: true`. This was causing jobs to move from `RUNNING_WITH_ERRORS` to `COMPLETED_WITH_ERRORS` prematurely. It also adds a validation for this setting and tests for the validation.

## Local Test Steps
1. Do a deployment configured to point to production CMR/EDL. Or use my sandbox deployment here https://internal-harmony-jn-frontend-13863418.us-west-2.elb.amazonaws.com
2. Run the following query
```
http://localhost:3000/OPERA_L2_RTC-S1_V1/ogc-api-coverages/1.0.0/collections/parameter_vars/coverage/rangeset?forceAsync=true&subset=time(%222025-01-01T00%3A00%3A00%22%3A%222025-01-02T00%3A00%3A00%22)&destinationUrl=s3%3A%2F%2Fcdd-harmony-sandbox-staging%2Fopera-browse-test&format=image%2Fpng&skipPreview=true&variable=all
```
This job will run very slowly due to issues with getting file sizes from S3 when updating work-items from query-cmr. Eventually it should move to `RUNNING_WITH_ERRORS`, and then after about an hour `COMPLETED_WITH_ERRORS`. 
3. Use the workflow-ui to verify that there are no running work-items. Or check the database.
You should see something like this if you set a `status: running` filter (no work-items runniing)
image

## PR Acceptance Checklist
* [x] Acceptance criteria met
* [x] Harmony in a Box tested? (if changes made to microservices or new dependencies added)

### DIFF
--- a/services/harmony/app/models/services/index.ts
+++ b/services/harmony/app/models/services/index.ts
@@ -1,20 +1,21 @@
 import * as fs from 'fs';
-import * as path from 'path';
 import * as yaml from 'js-yaml';
 import _, { get as getIn } from 'lodash';
+import * as path from 'path';
 
-import logger from '../../util/log';
-import { HttpError, NotFoundError, ServerError } from '../../util/errors';
-import { isMimeTypeAccepted, allowsAny } from '../../util/content-negotiation';
-import { CmrCollection } from '../../util/cmr';
+import { Conjunction, isInteger, listToText } from '@harmony/util/string';
+
 import { addCollectionsToServicesByAssociation } from '../../middleware/service-selection';
-import { listToText, Conjunction, isInteger } from '@harmony/util/string';
-import TurboService from './turbo-service';
-import HttpService from './http-service';
-import DataOperation from '../data-operation';
-import BaseService, { ServiceConfig } from './base-service';
-import RequestContext from '../request-context';
+import { CmrCollection } from '../../util/cmr';
+import { allowsAny, isMimeTypeAccepted } from '../../util/content-negotiation';
 import env from '../../util/env';
+import { HttpError, NotFoundError, ServerError } from '../../util/errors';
+import logger from '../../util/log';
+import DataOperation from '../data-operation';
+import RequestContext from '../request-context';
+import BaseService, { ServiceConfig } from './base-service';
+import HttpService from './http-service';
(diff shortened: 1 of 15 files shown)


chris-durbin (approved):
I couldn't find documentation for `is_sequential` - can you make sure to add something in the adapting services guide in the section where we have an example service chain definition.

services/harmony/app/models/services/index.ts
```diff
@@ -109,6 +110,10 @@ function validateServiceConfigSteps(config: ServiceConfig<unknown>): void {
           + `Configured to use ${maxBatchInputs}, but will be limited to ${env.maxGranuleLimit}`);
       }
     }
+    if (step.image.match(/harmonyservices\/query\-cmr:.*/) && !step.is_sequential) {
+
+      throw new TypeError(`Invalid is_sequential ${step.is_sequential}. query-cmr steps must always have sequential = true.`);
```
ygliuvt, line 115: nit: should we remove the `is_sequential: true` setting from `query-cmr` step and enforce it in the code? This feels like leaking unnecessary implementation details to the client.

ygliuvt (approved):

services/harmony/app/models/services/index.ts
indiejames: This is not a client error. This is an error that shows up when you start Harmony if it is misconfigured. I think I want to keep it this way instead of adding separate handling for query-cmr.

ygliuvt: I should have said service provider rather than client. The service providers shouldn't have to know the `query-cmr` step is sequential when they set up the service chain in services.yml. The fact that Harmony always execute query cmr sequentially shouldn't have to be spelled out in the service configuration especially when it is something the service providers doesn't care or know. Like I said that I am ok with having it spelled out because we do have other steps that also use the `is_sequential` attribute, but I don't think it is necessary.

PR #708, opened 2025-03-04
HARMONY-2016: Fix race condition in work reaper.

ygliuvt:
## Jira Issue ID
HARMONY-2016

## Description
I think the issue is caused by the race condition between old work items cleanup and old workflow steps cleanup. Currently, work reaper uses the current time minus the `reapableWorkAgeMinutes` configuration value to find and cleanup old work items and old workflow steps, but because it takes time to run old work items cleanup, by the time it gets to workflow steps cleanup, there might be more work items that are eligible for cleanup based on the current time, thus causes the error documented in this ticket. 

The fix is to pre-calculate the cutoff time before executing the work items cleanup and the workflow steps cleanup to avoid the race condition.

## Local Test Steps
npm run test

## PR Acceptance Checklist
* [x] Acceptance criteria met

### DIFF
--- a/services/cron-service/app/cronjobs/work-reaper.ts
+++ b/services/cron-service/app/cronjobs/work-reaper.ts
@@ -1,3 +1,4 @@
+import { subMinutes } from 'date-fns';
 import { JobStatus, terminalStates } from '../../../harmony/app/models/job';
 import {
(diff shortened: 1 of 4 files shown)


indiejames (approved):
Tested locally, verified that work-items and steps are deleted

chris-durbin (approved):
Good catch.

PR #710, opened 2025-03-06
Harmony 1963 - data size change information in job status page

indiejames:
## Jira Issue ID
HARMONY-1963

## Description
Adds information about data size change from original to output

## Local Test Steps
Run a few queries and look at the job status pages. Verify that the size information is at the bottom. For example
```
"originalDataSize": "1.37 MiB",
"outputDataSize": "92.53 KiB",
"dataSizePercentChange": "93.41% reduction"
```
Make sure the size change looks right.

## PR Acceptance Checklist
* [x] Acceptance criteria met
* [ ] ~Harmony in a Box tested (if changes made to microservices or new dependencies added)~


services/harmony/app/frontends/jobs.ts
```diff
@@ -171,6 +172,58 @@ export async function getJobsListing(
...
+  }
+  let result: string;
+  const diff = sizes.originalSize - sizes.outputSize;
+  if (diff < 0) {
+    const percent = (-diff / sizes.originalSize * 100.0).toFixed(precision);
+    result = `${percent}% increase`;
+  } else if (diff > 0) {
+    let percent = (diff / sizes.originalSize * 100.0).toFixed(precision);
+    // due to JS precision issues, big changes will appear to be 100% reduction, which is impossible
+    if (percent === '100.00') percent = '99.99';
```
ygliuvt, line 199: nit: This will not work if the precision is not the default (2).

indiejames: Good catch. Fixed and test added.

ygliuvt (approved):
Tested successfully locally.

services/harmony/test/jobs/jobs-status.ts
```diff
@@ -74,16 +79,39 @@ describe('Individual job status route', function () {
   hookTransaction();
   before(async function () {
     await aJob.save(this.trx);
+    console.log(`JOB ID: ${aJob.jobID}`);
```
chris-durbin, line 82: Remove

chris-durbin (commented):
My initial reaction is that I think we need to store these values and calculate them once.
1. Work items are transient (they get cleaned up after a job completes based on a configuration interval).
2. Calculating it on every call to get the job status for a completed request could be slow for something with a lot of granules.

services/harmony/test/jobs/jobs-status.ts
indiejames: Done

indiejames:
> My initial reaction is that I think we need to store these values and calculate them once.
> 
> 1. Work items are transient (they get cleaned up after a job completes based on a configuration interval).
> 2. Calculating it on every call to get the job status for a completed request could be slow for something with a lot of granules.

Done

chris-durbin (approved):
Tested successfully locally.

PR #717, opened 2025-03-12
HARMONY-2035: Fix POST OGC requests boolean parameter validation 

ygliuvt:
## Jira Issue ID
HARMONY-2035

## Description
This is Harmony part of the code change to support OGC POST requests submitted by harmony-py client. These harmony-py requests have string values for boolean parameters that is not currently supported by Harmony. This PR adds such support.

## Local Test Steps
See [the corresponding harmony-py PR](https://github.com/nasa/harmony-py/pull/109) for testing instructions.

## PR Acceptance Checklist
* [x] Acceptance criteria met


services/harmony/app/util/parameter-parsers.ts
```diff
@@ -220,3 +220,79 @@ export function handleAveragingType(
...
+ * Handle the forceAsync parameter in a Harmony query, adding it to the DataOperation
+ * if necessary.
+ *
+ * @param operation - the DataOperation for the request
+ * @param query - the query for the request
+ */
+export function handleForceAsync(
+  operation: DataOperation,
+  query: Record<string, string | boolean>): void {
+  if (query.forceasync !== undefined) {
```
indiejames, line 234: I think this code should be pulled out into a separate function like `queryValueToBoolean` that can be used here and in the other two cases. That would make it easy to unit test as well.

ygliuvt: The three boolean field values are parsed differently as to if undefined is allowed and when will the default value apply. I thought about it hard and ended up with the current code.

chris-durbin (changes requested):
We should change the OGC coverages and EDR schemas to make the types string instead of boolean for all 3 fields so that the validations work for GET requests as well.

ygliuvt:
> We should change the OGC coverages and EDR schemas to make the types string instead of boolean for all 3 fields so that the validations work for GET requests as well.

That is ticket HARMONY-2024. This is just to make harmony-py requests which is POST working. My thought was that we will fix things properly in HARMONY-2024 which may revamp all these code changes.

ygliuvt:
Close this PR as we have decided to pull in HARMONY-2024 and work on the whole OGC GET/POST boolean parameter validation as a whole rather than this stop gap PR just for harmony-py request without the full validations. I will do a new PR for HARMONY-2024 with this change as part of it.

services/harmony/app/util/parameter-parsers.ts
ygliuvt: As I was refactoring the code, I realized that your suggestion is a great idea.

PR #721, opened 2025-03-19
HARMONY-1986: As a Harmony operator I want the Harmony health endpoint to provide more detailed and fine-grained information

ygliuvt:
## Jira Issue ID
HARMONY-1986

## Description
Add CMR and EDL dependencies to health check.

## Local Test Steps
http://localhost:3000/health now shows the CMR and EDL status in addition to db status:
```
{
  status: "up",
  message: "Harmony is operating normally.",
  dependencies: [
    {
      name: "db",
      status: "up"
    },
    {
      name: "cmr",
      status: "up"
    },
    {
      name: "edl",
      status: "up"
    }
  ]
}
```

## PR Acceptance Checklist
* [x] Acceptance criteria met

### DIFF
--- a/services/harmony/app/frontends/health.ts
+++ b/services/harmony/app/frontends/health.ts
@@ -1,5 +1,5 @@
 import { Response, NextFunction } from 'express';
-import { isCmrHealthy } from '../util/cmr';
+import { getCmrHealth } from '../util/cmr';
 import { isEdlHealthy } from '../util/edl-api';
 import HarmonyRequest from '../models/harmony-request';
 import { Job } from '../models/job';
@@ -7,25 +7,24 @@ import logger from '../util/log';
 import RequestContext from '../models/request-context';
 import db from '../util/db';
 
-const UP = 'up';
-const DOWN = 'down';
-const _DEGRADED = 'degraded';
+export enum HealthStatus {
+  UP = 'up',
+  DOWN = 'down',
+}
 
 const HEALTHY_MESSAGE = 'Harmony is operating normally.';
 const DOWN_MESSAGE = 'Harmony is currently down.';
-const _DEGRADED_MESSAGE = 'Harmony is currently degraded.';
 
-export interface DependencyStatus {
+interface DependencyStatus {
   name: String;
   status: String;
   message?: String;
 }
 
-interface HealthStatus {
+interface HealthInfo {
   status: String;
   message?: String;
   dependencies: DependencyStatus[];
-
 }
 
 /**
@@ -34,25 +33,12 @@ interface HealthStatus {
  * @returns a promise resolving to the DependencyStatus of database
  */
 async function getDbHealth(): Promise<DependencyStatus> {
-  let dbHealthy = true;
   try {
     await Job.getTimeOfMostRecentlyUpdatedJob(db);
+    return { name: 'db', status: HealthStatus.UP };
   } catch (e) {
     logger.error(e);
-    dbHealthy = false;
-  }
-
-  if (dbHealthy) {
-    return {
-      name: 'db',
-      status: UP,
-    };
-  } else {
-    return {
-      name: 'db',
-      status: DOWN,
-      message: 'Unable to query the database',
-    };
+    return { name: 'db', status: HealthStatus.DOWN, message: 'Unable to query the database' };
   }
 }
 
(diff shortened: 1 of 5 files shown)


services/harmony/app/util/cmr.ts
```diff
@@ -1091,3 +1091,33 @@ export function filterGranuleLinks(
...
+    if (response.status === 200) {
+      return {
+        name: 'cmr',
+        status: 'up',
+      };
+    } else {
+      return {
+        name: 'cmr',
+        status: 'down',
+        message: `${JSON.stringify(response.data)}`,
```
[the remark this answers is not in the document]
ygliuvt: If CMR is down and responding (CMR returns 503 in this case), the message will be the CMR health status with details. The hope is that when other errors, it will go to the exception clause. But I think you raised a good point that what if CMR or cloudfront returns something that is not a JSON. 

To make it safe, I think I will just change the message to: `CMR is down. For details see: https://<cmr_root>/search/helath`.

services/harmony/app/util/cmr.ts
```diff
@@ -1091,3 +1091,33 @@ export function filterGranuleLinks(
   return granule.links.filter((g) => (g.rel.endsWith('/data#') || g.rel.endsWith('/service#'))
     && !g.inherited);
 }
+
+/**
+ * Returns CMR health information
+ *
+ * @returns a promise resolving to the DependencyStatus of CMR
+ */
+export async function getCmrHealth(): Promise<object> {
```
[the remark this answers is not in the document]
ygliuvt: Sure. Especially when I no longer passing the detailed message from CMR back.

services/harmony/app/util/cmr.ts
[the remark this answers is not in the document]
ygliuvt: Decided to still pass the CMR health info back. It would be useful.

chris-durbin (approved):

services/harmony/app/frontends/health.ts
```diff
@@ -12,7 +15,7 @@ const HEALTHY_MESSAGE = 'Harmony is operating normally.';
 const DOWN_MESSAGE = 'Harmony is currently down.';
 const _DEGRADED_MESSAGE = 'Harmony is currently degraded.';
 
-interface ComponentStatus {
+export interface DependencyStatus {
```
indiejames, line 18: I don't think you need this interface. You can just make the `dependencies` property on `HeathStatus` and optional of the type `HealthStatus[]`, like this
```
interface HealthStatus {
  name: String;
  status: String;
  message? String;
  dependencies?: HealthStatus[];
  }
  ```

File: services/harmony/app/frontends/health.ts [hunk 4]
```diff
@@ -21,46 +24,79 @@ interface ComponentStatus {
...
+}
+
+/**
+ * Returns the health information
+ *
+ * @param context - the request context
+ * @returns a promise resolving to the health check response
+ */
+async function getGeneralHealth(context: RequestContext): Promise<HealthStatus> {
+  const dbHealth = await getDbHealth();
```
indiejames, line 66: nit: I _think_ it is slightly more efficient to not await each request, instead, gather them all in a `Promise.all` and `await` that. Probably not a big deal for this use case, though.

services/harmony/app/util/edl-api.ts
```diff
@@ -215,3 +215,21 @@ export async function validateUserIsInCoreGroup(
...
+/**
+ * Returns true if EDL is up
+ *
+ * @param context - request context
+ * @returns A Promise containing `true` if EDL is up
+ */
+export async function isEdlHealthy(context: RequestContext): Promise<boolean> {
+  const { logger } = context;
+  try {
+    const response = await axios.default.get(`${env.oauthHost}/home`);
```
indiejames, line 228: nit: you are using `fetch` in `isCmrHealthy`, but `axios` here. Any reason for not using the same in both?

services/harmony/app/util/cmr.ts
```diff
@@ -1091,3 +1092,30 @@ export function filterGranuleLinks(
   return granule.links.filter((g) => (g.rel.endsWith('/data#') || g.rel.endsWith('/service#'))
     && !g.inherited);
 }
+
+/**
+ * Returns CMR health information
+ *
+ * @returns A Promise with healthy and message indicating the CMR health info
+ */
+export async function isCmrHealthy(): Promise<{ healthy: boolean; message: string }> {
```
indiejames, line 1101: nit: kind of weird to have a function name `isSomething` that doesn't return a boolean, or Promise<boolean>

services/harmony/app/frontends/health.ts
```diff
@@ -1,8 +1,11 @@
...
+import { isCmrHealthy } from '../util/cmr';
+import { isEdlHealthy } from '../util/edl-api';
 import HarmonyRequest from '../models/harmony-request';
 import { Job } from '../models/job';
 import logger from '../util/log';
-import db, { Transaction } from '../util/db';
+import RequestContext from '../models/request-context';
+import db from '../util/db';
 
 const UP = 'up';
```
indiejames, line 10: an enumerated type would be better here rather than strings. That way you get type checking that can maker sure someone doesn't accidentally return an arbitrary string instead of using one of these

services/harmony/app/util/edl-api.ts
```diff
@@ -215,3 +215,21 @@ export async function validateUserIsInCoreGroup(
...
   return true;
 }
+
+/**
+ * Returns true if EDL is up
+ *
+ * @param context - request context
+ * @returns A Promise containing `true` if EDL is up
+ */
+export async function isEdlHealthy(context: RequestContext): Promise<boolean> {
```
indiejames, line 225: I think it might be better if all of these types of functions were consistent in returning a `Promise<HealthStatus>` instead of `Promise<boolean>` or `Promise<{health: boolean; message: string}>`, etc. And I would name them `getEdlHealth`, `getCmrHealth`, etc.

indiejames (commented):
Tested locally and it is working I had some general comment that are up for debate.

services/harmony/app/frontends/health.ts
ygliuvt: I didn't make this interface. Just renamed it. Since `DependencyStatus` does not have `dependencies`, it doesn't hurt to make a different interface to make this clear.

services/harmony/app/util/cmr.ts
```diff
@@ -1091,3 +1092,30 @@ export function filterGranuleLinks(
...
+  try {
+    const response: CmrResponse = await fetch(`${cmrApiConfig.baseURL}/search/health`);
+
+    if (response.status === 200) {
+      return { healthy: true, message: '' };
```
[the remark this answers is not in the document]
ygliuvt: `message` is not optional currently in the function definition.

services/harmony/app/util/edl-api.ts
ygliuvt: Because CMR namespace is using `fetch` and edl-api namespace is using `axios`. I am just trying to be consistent within the same namespace. As to why these namesapces use different libraries, it is beyond me.

services/harmony/app/util/edl-api.ts
ygliuvt: It was sort like what you suggested in the original draft and Chris requested changing cmr health check to be something like the `isEdlHealthy`, and since I also want to preserve the CMR health check message, it became this way.

services/harmony/app/frontends/health.ts
ygliuvt: Changed to enum.

services/harmony/app/util/edl-api.ts
ygliuvt: I will rename the function back to `getCmrHealth`.

services/harmony/app/frontends/health.ts
ygliuvt: Refactored.

services/harmony/app/util/cmr.ts
ygliuvt: Renamed function.

services/harmony/app/util/cmr.ts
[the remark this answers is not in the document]
ygliuvt: I made it optional now.

indiejames (approved):

PR #722, opened 2025-03-19
Harmony 2037 - fix for SAMBAH bug

indiejames:
## Jira Issue ID
HARMONY-2037

## Description
Fixes a bug where stitchee was being treated as if it were an aggregating service and therefore only one instance was being created when multiple should have been.

## Local Test Steps
Run the following query
```
http://localhost:3000/C1254854453-LARC_CLOUD/ogc-api-coverages/1.0.0/collections/all/coverage/rangeset?concatenate=true&maxResults=40&extend=true
```
You should see one batchee and six stitchee, plus one concise.

## PR Acceptance Checklist
* [x] Acceptance criteria met

### DIFF
--- a/services/harmony/app/models/services/base-service.ts
+++ b/services/harmony/app/models/services/base-service.ts
@@ -148,7 +148,7 @@ const multiCatalogOperations = [
  * @param operation - The operation
  * @returns true if the step uses more than one input catalog, false otherwise
  */
-function stepUsesMultipleInputCatalogs(step: ServiceStep, operation: DataOperation): boolean {
+export function stepUsesMultipleInputCatalogs(step: ServiceStep, operation: DataOperation): boolean {
   // get the operations for this step that support multiple input catalogs
   const multiCatOps = _.intersection(multiCatalogOperations, step.operations);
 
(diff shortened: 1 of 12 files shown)


services/harmony/app/models/services/base-service.ts
```diff
@@ -134,25 +134,35 @@ const conditionToOperationField = {
...
-  return (
-    (operation.shouldConcatenate || operation.shouldExtend) &&
-    _.intersection(aggregatingOperations, step.operations).length > 0
-  );
+export function stepUsesMultipleInputCatalogs(step: ServiceStep, operation: DataOperation): boolean {
+  // get the operations for this step that support multiple input catalogs
+  const multiCatOps = _.intersection(multiCatalogOperations, step.operations);
+
+  // check to see if the user has actually requested any of the multi-catalog operations
+  for (const op of multiCatOps) {
```
chris-durbin, line 156: I worry about things that rely on convention. I'd almost rather see an explicit key value map defined between the multi-catalog step and the function to call on the operation. I don't feel that strongly about which way is better, but if we're relying on convention I think it's important to update the docstring for this function to make it clear what the expectations are.

chris-durbin (approved):
Tested and verified that 6 STITCHEE work items are created. My concise request failed, but I believe that's just because I didn't have enough memory allocated for it locally.

ygliuvt (approved):

PR #725, opened 2025-03-24
HARMONY-2032: Update work-failer default timeout to 5 minutes

chris-durbin:
## Jira Issue ID
HARMONY-2032

## Description
This fixes an issue we often ran into when regression tests were running where a work item would stay running for a long time. It turned out that we were defaulting to waiting for 2 hours before timing out a work item if no other work item for that job had completed.

This change sets the default timeout to 5 minutes for all services except concise. Concise will allow up to 15 minutes for the first concise work item for a job since it often takes longer to complete.

## Local Test Steps
Change a service such as harmony-service-example to sleep for a long time. For example:
```
import time

time.sleep(100000)
```
Verify that work failer sets up the work item to be retried after 5 minutes. You might want to change the frequency at which the work-failer runs to make sure it happens sooner (otherwise you could wait for up to 10 minutes with the default configuration): e.g. `FAILABLE_WORK_AGE_MINUTES=1`

Since we only run one pod for each service locally you won't see the work item get queued to run again. If you want to test through all retries you'd need to deploy to harmony sandbox.

If testing with harmony in a box make sure to build the images (at a minimum the work-failer image).

## PR Acceptance Checklist
* [X] Acceptance criteria met


indiejames (approved):
Tested locally, verified retries were happening on schedule

ygliuvt (approved):

PR #729, opened 2025-03-31
HARMONY-1620: Add cron job to reset user-work

indiejames:
## Jira Issue ID
HARMONY-1620

## Description
Adds a cron job that resets the user_work rows for any jobs with rows that have `ready_count` or `runnning_count` greater than zero and have not been updated in the last hour.

## Local Test Steps
* set the cron job to run every minute
```
export USER_WORK_UPDATER_CRON="*/1 * * * *"
```
* Insert a row into your user_work table with an old `last_worked` datetime
```
INSERT INTO user_work values(100,'joe', 'abc', 10, 13, 'job4', true,'2025-03-31 16:19:55', '2025-03-31 16:19:55','2025-03-31 16:19:55')
```
* start harmony and services
* select the user_work row you inserted from the table after a minute or so and verify that the counts have been reset to 0
* to test paused jobs, run a big job with `skipPreview=false`
* after it pauses set the `ready_count` and `running_count` to big values in the `user_work` table for the job.
* tail the cron-service log to verify that the `UserWorkUpdater` has reset the counts on the job
* Check the row in `user_work` to verify that the `ready_count` and `running_count` are zeros
## PR Acceptance Checklist
* [x] Acceptance criteria met


services/cron-service/app/cronjobs/update-user-work.ts
```diff
@@ -0,0 +1,97 @@
...
+      query = query.andWhere(tx.raw(`last_worked <= ${timeStamp}`));
+    } else {
+      query = query.andWhere('last_worked', '<=', tx.raw(`now() - interval '${env.userWorkUpdateAge}'`));
+    }
+
+    const results = await query;
+    const jobIDs = results.map((r) => r.job_id);
+
+    // reset the counts for the jobs
+    for (const jobID of jobIDs) {
```
chris-durbin, line 74: This will cause the counts for paused jobs to be set to the number of ready and running items which will be non-zero and cause work to start for them and spin up pods. We should add a second query to pull back the jobs so that we get their statuses.

Then we want two different behaviors.
1. For the paused jobs we want to set both the ready and running counts to 0 (reset them and prevent any work from being started and any extra pods sitting around for items that aren't being worked).
2. For the active jobs we want to call recalculateCount to set the values accurately.

indiejames: Done.

services/cron-service/app/cronjobs/update-user-work.ts
```diff
@@ -0,0 +1,105 @@
...
+      logger.info(`Resetting user-work counts for job ${jobID}`);
+      const { job } = await Job.byJobID(tx, jobID);
+      if (job.status === JobStatus.PAUSED) {
```
ygliuvt, line 80: Do we need to consider other job status e.g. failed, canceled, etc.?

services/cron-service/app/cronjobs/update-user-work.ts
```diff
@@ -0,0 +1,105 @@
...
+        await setReadyAndRunningCountToZero(tx, jobID);
+      } else {
+        await recalculateCount(tx, jobID, 'ready');
```
ygliuvt, line 83: We can replace these two calls with one call to `recalculateCounts`.

services/cron-service/env-defaults
```diff
@@ -12,3 +12,9 @@ WORK_REAPER_BATCH_SIZE=2000
 
 # The cron schedule for restarting Prometheus if it is down
 RESTART_PROMETHEUS_CRON="*/10 * * * *" # every 10 minutes
+
+# The cron schedule for the user-work updater
+USER_WORK_UPDATER_CRON="*/30 * * * *" # every thirty minutes
+
+# Rows in user-work older than this will cause the user-work for a job to be recalculated
+USER_WORK_UPDATE_AGE="1 HOUR"
```
ygliuvt, line 20: I am lazy, so I don't quite see the point of making this configuration parameter value that we control so complicated. Why not just call it something like `USER_WORK_EXPIRATION_MINUTES` with a value type of number?

ygliuvt: Then we can delete most of the code in this PR.

chris-durbin: I would prefer a numeric value in minutes over a string as well.

services/cron-service/app/cronjobs/update-user-work.ts
indiejames: didn't see that function - done.

services/cron-service/app/cronjobs/update-user-work.ts
indiejames: no, they would not have ready or running work-items

indiejames: I don't really understand what Chris is saying here, but jobs in terminal states that still have rows (that are expired) will have their counts set to zero by the call to `recalculateCounts` since they will have no `ready` or `running` work-items. Are you two saying we should just check for terminal states and call `setReadyAndRunningCountToZero` for those to avoid the recalculation which is slightly more expensive?

indiejames: Okay, I made that change to zero out the user_work for other states instead of calling `recalculateCounts`

services/cron-service/env-defaults
indiejames: done

services/cron-service/app/cronjobs/update-user-work.ts
chris-durbin: I was saying we delete entries from the user_work table when jobs move to a terminal state so you shouldn't ever hit this scenario.

services/cron-service/app/cronjobs/update-user-work.ts
```diff
@@ -0,0 +1,60 @@
...
+    } else {
+      query = query.andWhere('last_worked', '<=', tx.raw(`now() - interval '${env.userWorkExpirationMinutes} minutes'`));
+    }
+
+    const results = await query;
+    const jobIDs = results.map((r) => r.job_id);
+
+    // reset the counts for the jobs
+    for (const jobID of jobIDs) {
+      logger.info(`Resetting user-work counts for job ${jobID}`);
```
chris-durbin, line 34: Can you log different things for resetting vs recalculating - maybe:
logger.warn(`Resetting user_work counts to 0 for job ${jobID} with status ${job.status}`); // If
logger.warn(`Recalculating user_work counts for job ${jobID} with status ${job.status}`); // Else

services/cron-service/env-defaults
```diff
@@ -12,3 +12,9 @@ WORK_REAPER_BATCH_SIZE=2000
 
 # The cron schedule for restarting Prometheus if it is down
 RESTART_PROMETHEUS_CRON="*/10 * * * *" # every 10 minutes
+
+# The cron schedule for the user-work updater
+USER_WORK_UPDATER_CRON="*/30 * * * *" # every thirty minutes
+
+# Rows in user-work older than this will cause the user-work for a job to be recalculated
+USER_WORK_EXPIRATION_MINUTES="60"
```
chris-durbin, line 20: Any reason this one needs to be a string (for other parameters like this we use integers e.g. REAPABLE_WORK_AGE_MINUTES=20160)?

chris-durbin (approved):

services/cron-service/app/cronjobs/update-user-work.ts
indiejames: done

services/cron-service/app/cronjobs/update-user-work.ts
```diff
@@ -0,0 +1,61 @@
...
+ */
+export class UserWorkUpdater extends CronJob {
+
+  static async run(ctx: Context): Promise<void> {
+    const { logger } = ctx;
+    logger.debug('Running');
+    try {
+      await updateUserWork(ctx);
+    } catch (e) {
+      logger.error('User work udpater failed to update user-work table');
```
ygliuvt, line 57: nit: The table name is `user_work`, not `user-work`.

indiejames: fixed

PR #732, opened 2025-04-02
HARMONY-2051: Do not fail a job immediately if a work item does not provide outputs

chris-durbin:
## Jira Issue ID
HARMONY-2051

## Description
We've started seeing an error in production and reproduced in sandbox where work items are returning a successful status but without any output results. I'm not sure of the root cause yet, but this PR will change the error to be retried and only considered as a work item error as opposed to immediately failing the job.

## Local Test Steps
Since I've been unable to figure out the exact root cause I'm just testing in the sandbox environment with an updated work-item-updater image. I'm finding that during an initial scale up I'm seeing the problem occur for quite a few work items and then the issues stopped. In looking at the pods I found some errors like the following (though could be unrelated):

```
{"env_name":"harmony-cdd-sandbox","level":"debug","message":"SIDECAR STATUS: {\n  \"metadata\": {},\n  \"status\": \"Failure\",\n  \"message\": \"Internal error occurred: unable to upgrade connection: rpc error: code = Unavailable desc = connection error: desc = \\\"transport: Error while dialing: dial unix /run/containerd/containerd.sock: connect: no such file or directory\\\"\",\n  \"reason\": \"InternalError\",\n  \"details\": {\n    \"causes\": [\n      {\n        \"message\": \"unable to upgrade connection: rpc error: code = Unavailable desc = connection error: desc = \\\"transport: Error while dialing: dial unix /run/containerd/containerd.sock: connect: no such file or directory\\\"\"\n      }\n    ]\n  },\n  \"code\": 500\n}","timestamp":"2025-04-02T20:25:19.728Z","workItemId":116800}
```

In my testing I'm finding some work items failing 5 times and being marked as failed, but this change will at least be a first step to prevent jobs from immediately failing.

## PR Acceptance Checklist
* [X] Acceptance criteria met

### DIFF
--- a/services/harmony/app/backends/workflow-orchestration/work-item-updates.ts
+++ b/services/harmony/app/backends/workflow-orchestration/work-item-updates.ts
@@ -696,7 +696,7 @@ export async function processWorkItem(
       return;
     }
 
-    const hasOutput = results?.length > 0;
+    const hasOutput = results && results.length > 0;
     if (status === WorkItemStatus.SUCCESSFUL && !hasOutput) {
       // To return no output the status needs to be a warning, otherwise we treat it as an error
       logger.error('The work item update should have contained results, but it did not.');
--- a/services/harmony/app/backends/workflow-orchestration/work-item-updates.ts
+++ b/services/harmony/app/backends/workflow-orchestration/work-item-updates.ts
@@ -27,7 +27,7 @@ import WorkflowStep, {
 import { handleBatching, outputStacItemUrls, resultItemSizes } from '../../util/aggregation-batch';
 import db, { batchSize, Transaction } from '../../util/db';
 import env from '../../util/env';
-import { ServiceError } from '../../util/errors';
+import { ServerError, ServiceError } from '../../util/errors';
 import { completeJob } from '../../util/job';
 import { logAsyncExecutionTime } from '../../util/log-execution';
 import { objectStoreForProtocol } from '../../util/object-store';
@@ -83,8 +83,6 @@ async function addJobLinksForFinishedWorkItem(
     }
   }
 }
-
-
 
 /**
  * Returns the final job status and message for the request based on whether all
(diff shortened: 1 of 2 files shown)


services/harmony/app/backends/workflow-orchestration/work-item-updates.ts
```diff
@@ -698,6 +696,14 @@ export async function processWorkItem(
       return;
     }
 
+    const hasOutput = results?.length > 0;
```
indiejames, line 699: I'm surprised this line doesn't give you a compiler error. if results can be null then `results?.length` will be null, which can't be used with `>`

indiejames: I think
```
const hasOutput = results &&  results.length > 0
```
is better

indiejames (approved):
Just the one comment about `results?length > 0`, but otherwise looks good

services/harmony/app/backends/workflow-orchestration/work-item-updates.ts
chris-durbin: The ? is shorthand for what you wrote.

ygliuvt (approved):
Run a test in Sandbox with a 110k granule l2 subset request and by the time I cancelled it at 70% complete, there are 91 failed work items (when there should only be 4 work items fail for legitimate issues). But this is still better than failing the request as a whole without most of the granules processed.

services/harmony/app/backends/workflow-orchestration/work-item-updates.ts
chris-durbin: I was wrong, James is correct - well mostly - the documentation says it will short circuit to undefined rather than null: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining

PR #740, opened 2025-04-22
Harmony 2025 - Add support for custom environment variables for services

indiejames:
## Jira Issue ID
HARMONY-2052

## Description
This PR moves service specific environment variables out of the `harmony_env` config map and into the yaml template file for the service, so that services can't see environment variables for other services.

## Local Test Steps
* add s service specific variable, e.g.,` BATCHEE_FOO=bar`, to your `.env` file and another one to `services/harmony/env-defaults` for one of the services in your `LOCALLY_DEPLOYED_SERVICES`.
* run ./bin/deploy and make sure queries still work. 
* use `kubectl` or `k9s` to describe the pod for the service
* verify that the environment variables are defined
* (optional) ssh into the worker container for the pod and `echo` the variables to make sure they are correct 

## PR Acceptance Checklist
* [x] Acceptance criteria met

### DIFF
--- a/bin/create-k8s-config-maps-and-secrets
+++ b/bin/create-k8s-config-maps-and-secrets
@@ -1,5 +1,25 @@
 #!/bin/bash
 
+# check if the input string is a service specific variable name, e.g.
+# BATCHEE_IMAGE. note that variable names that end in QUEUE_URLS are not service
+# specific variables - they are used by harmony itself.
+# returns 0 if the string is the name of a service specific variable name, 1 otherwise
+is_service_var() {
+   if [[ "$1" == *"QUEUE_URLS" ]]; then
+    return 1 # we need to include QUEUE_URLS env vars in the main config map
+  fi
+
+  for service in ${LOCALLY_DEPLOYED_SERVICES//,/ }
+  do
+    local var_prefix=`echo $service | tr 'a-z' 'A-Z' | tr '-' '_'`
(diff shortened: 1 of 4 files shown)


bin/create-k8s-config-maps-and-secrets
```diff
@@ -1,5 +1,25 @@
 #!/bin/bash
 
+# check if the input string is a service specific variable name, e.g.
+# BATCHEE_IMAGE. note that variable names that end in QUEUE_URLS are not service
+# specific variables - they are used by harmony itself.
+# returns 0 if the string is the name of a service specific variable name, 1 otherwise
```
chris-durbin, line 6: Nit: This is backwards for most boolean logic where 1 means true and 0 means false

chris-durbin (approved):
I tested in sandbox and locally. I noticed that generating the config maps is really slow (it was already slow, but seems to be getting worse). I wonder if there's something we can improve in the script for generating the config maps.

bin/create-k8s-config-maps-and-secrets
indiejames: Yes, I realized that, but for scripts - 0 means success, non-zero means error, so this is why I did it this way. Also, it's what claude suggested

indiejames: Also, the way it is used depends on this convention
```
$(is_service_var "$name") && continue # skip service specific variables
```

PR #741, opened 2025-04-23
HARMONY-2070: As a harmony user, I want my EDL bearer token cached so that I can submit many simultaneous requests

ygliuvt:
## Jira Issue ID
HARMONY-2070

## Description
Cache EDL bearer token lookup to improve performance.

## Local Test Steps
Add the following line to line 54 of services/harmony/app/util/edl-api.ts to facilitate the testing:
console.log(`calling getUserIdRequest with token: ${userToken}`);

Use the following curl to submit an example harmony request with EDL bearer token:
curl -i -H "Authorization: Bearer $HARMONY_UAT_TOKEN" "http://localhost:3000/C1233800302-EEDTEST/ogc-api-coverages/1.0.0/collections/blue_var/coverage/rangeset?subset=lat(20%3A60)&subset=lon(-140%3A-50)&granuleId=G1233800343-EEDTEST&outputCrs=EPSG%3A31975&format=image%2Fpng"

Verify the `getUserIdRequest` function is called for only the first request, not the subsequent ones.
Verify after 5 minutes (the cache will expire), the `getUserIdRequest` function is called again to repopulate the cache and then skipped for the subsequent requests.

Verify the following request (invalid token) will return a 403 status code with proper error message:
curl -i -H "Authorization: Bearer HARMONY_UAT_TOKEN" "http://localhost:3000/C1233800302-EEDTEST/ogc-api-coverages/1.0.0/collections/blue_var/coverage/rangeset?subset=lat(20%3A60)&subset=lon(-140%3A-50)&granuleId=G1233800343-EEDTEST&outputCrs=EPSG%3A31975&format=image%2Fpng"

Add `TOKEN_CACHE_TTL=60000` to .env, restart server, repeat the tests above and verify the EDL token lookup is called after a minute.

## PR Acceptance Checklist
* [x] Acceptance criteria met

### DIFF
--- a/services/harmony/app/util/cache/memory-cache.ts
+++ b/services/harmony/app/util/cache/memory-cache.ts
@@ -40,26 +40,23 @@ export class MemoryCache extends Cache {
     const cached = this.data.get(key);
     if (cached !== undefined) return cached;
 
-    // Check for in-progress fetch
-    const existingPromise = this.pending.get(key);
-    if (existingPromise) return existingPromise;
+    if (this.pending.has(key)) {
+      return this.pending.get(key)!;
+    }
 
-    // Initiate fetch and store the promise
-    const fetchPromise = this.fetchMethod(key, context)
-      .then((result) => {
-        if (result !== undefined) {
-          this.data.set(key, result);
-        }
+    const promise = this.fetchMethod(key, context)
+      .then((value) => {
+        this.data.set(key, value);
         this.pending.delete(key);
-        return result;
+        return value;
       })
       .catch((err) => {
         this.pending.delete(key);
         throw err;
       });
 
-    this.pending.set(key, fetchPromise);
-    return fetchPromise;
+    this.pending.set(key, promise);
+    return promise;
   }
 
   async set(key: string, value: string): Promise<void> {
--- a/services/harmony/app/util/cache/memory-cache.ts
+++ b/services/harmony/app/util/cache/memory-cache.ts
@@ -4,27 +4,43 @@ import env from '../env';
 
 // Simple implementation of a string cache backed by an in-memory
 // least-recently-used (LRU) cache
-export class MemoryCache extends Cache {
-  data: LRUCache<string, string>;
 
-  constructor(fetchMethod: (k: string) => Promise<string>) {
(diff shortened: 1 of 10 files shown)


services/harmony/app/util/cache/memory-cache.ts
```diff
@@ -4,30 +4,65 @@ import env from '../env';
...
+        }
+        this.pending.delete(key);
+        return result;
+      })
+      .catch((err) => {
+        this.pending.delete(key);
+        throw err;
+      });
+
+    this.pending.set(key, fetchPromise);
```
chris-durbin, line 61: Looks like a race condition between the check on line 44 and then locking here, but I tested with 2000 requests as fast as possible and didn't run into a problem. It just results in an extra lookup if the race condition is hit - no error - so this seems ok.

services/harmony/app/util/cache/memory-cache.ts
```diff
@@ -4,30 +4,65 @@ import env from '../env';
...
 
-  async fetch(key: string): Promise<string> {
-    return this.data.fetch(key);
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  async fetch(key: string, context?: any): Promise<string> {
+    const cached = this.data.get(key);
+    if (cached !== undefined) return cached;
+
+    // Check for in-progress fetch
+    const existingPromise = this.pending.get(key);
```
chris-durbin, line 44: We should add a unit test to verify that if the key is in the pending list the existing promise is returned and a new one is not created.

chris-durbin (approved):
Tested successfully locally.

services/harmony/app/util/cache/memory-cache.ts
ygliuvt: I added a unit test.

services/harmony/app/util/cache/memory-cache.ts
indiejames: Everything between those steps is synchronous code, so I don't think it will ever get run by multiple requests simultaneously. Running an aync function is not the same as running multiple threads in other languages.

PR #742, opened 2025-04-24
HARMONY-2070: As a harmony user, I want my EDL bearer token cached so that I can submit many simultaneous requests.

ygliuvt:
## Jira Issue ID
HARMONY-2070

## Description
Here is a new approach that abandons the reuse of MemoryCache and just use LRUCache directly. I think it is much cleaner than the previous implementation.

## Local Test Steps
Add the following line to line 54 of services/harmony/app/util/edl-api.ts to facilitate the testing:
console.log(`calling getUserIdRequest with token: ${userToken}`);

Use the following curl to submit an example harmony request with EDL bearer token:
curl -i -H "Authorization: Bearer $HARMONY_UAT_TOKEN" "http://localhost:3000/C1233800302-EEDTEST/ogc-api-coverages/1.0.0/collections/blue_var/coverage/rangeset?subset=lat(20%3A60)&subset=lon(-140%3A-50)&granuleId=G1233800343-EEDTEST&outputCrs=EPSG%3A31975&format=image%2Fpng"

Verify the `getUserIdRequest` function is called for only the first request, not the subsequent ones.
Verify after 5 minutes (the cache will expire), the `getUserIdRequest` function is called again to repopulate the cache and then skipped for the subsequent requests.

Verify the following request (invalid token) will return a 403 status code with proper error message:
curl -i -H "Authorization: Bearer HARMONY_UAT_TOKEN" "http://localhost:3000/C1233800302-EEDTEST/ogc-api-coverages/1.0.0/collections/blue_var/coverage/rangeset?subset=lat(20%3A60)&subset=lon(-140%3A-50)&granuleId=G1233800343-EEDTEST&outputCrs=EPSG%3A31975&format=image%2Fpng"

Add `TOKEN_CACHE_TTL=60000` to .env, restart server, repeat the tests above and verify the EDL token lookup is called after a minute.

## PR Acceptance Checklist
* [x] Acceptance criteria met

### DIFF
--- a/services/harmony/app/middleware/earthdata-login-token-authorizer.ts
+++ b/services/harmony/app/middleware/earthdata-login-token-authorizer.ts
@@ -1,9 +1,32 @@
 import { RequestHandler } from 'express';
+import { LRUCache } from 'lru-cache';
 
 import HarmonyRequest from '../models/harmony-request';
 import { getUserIdRequest } from '../util/edl-api';
+import env from '../util/env';
 
 const BEARER_TOKEN_REGEX = new RegExp('^Bearer ([-a-zA-Z0-9._~+/]+)$', 'i');
+
+/**
+ * Wrapper function of getUserIdRequest to be set to fetchMethod of LRUCache.
+ *
+ * @param token - EDL bearer token
+ * @param _sv - default value parameter of LRUCache fetchMethod, unused here
+ * @param options - options parameter of LRUCache fetchMethod, carries the request context
+ * @returns Promise of user name associated with the EDL token
+ */
+async function fetchUserId(token: string, _sv: string, { context }): Promise<string> {
+  return getUserIdRequest(context, token);
+}
+
+// In memory cache with 5 min TTL for EDL token to username.
+// The token is valid if it exists in the cache.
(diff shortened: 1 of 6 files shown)


services/harmony/app/middleware/earthdata-login-token-authorizer.ts
```diff
@@ -1,10 +1,33 @@
...
 import { getUserIdRequest } from '../util/edl-api';
+import env from '../util/env';
 
 const BEARER_TOKEN_REGEX = new RegExp('^Bearer ([-a-zA-Z0-9._~+/]+)$', 'i');
 
+/**
+ * Wrapper function of getUserIdRequest to be set to fetchMethod of LRUCache.
+ *
+ * @param token - EDL bearer token
+ * @param _sv - default value parameter of LRUCache fetchMethod, unused here
```
indiejames, line 14: nit: technically `_sv` is the "stale value", not the default value. I guess there are some use cases where knowing what the old value was is necessary to get the new value (like maybe a counter or something), but not in our case.

indiejames (approved):
Tested locally

PR #743, opened 2025-04-25
HARMONY-2079: Fix issue with worker container not being ready

chris-durbin:
## Jira Issue ID
HARMONY-2079

## Description
Fixes a bug where the service-runner was attempting to make a request to the worker container but an error was being returned indicating the worker container was not ready.

I made a few changes to get more logging information saved in the work item logs we push to S3 in addition to changing the service runner to first make a k8s call to check if the worker is in a running state. It seems like that change to check fixed the problem altogether despite not finding any evidence that the check failed. I also changed the service runner to not immediately return an error to harmony if there was a k8s issue with the exec call (not the worker container returning a non zero exit code which is still handled in the same way). If there is a k8s issue it will check to make sure the worker container is running and then perform up to 5 internal retries before finally giving up and sending an "Internal server error" back to harmony.

## Local Test Steps
Build the service-runner image and submit requests - verify that requests work as expected and you can see additional log information about the pod that the request ran on and details about the command line that was run (by looking at the logs through the workflow-ui).

In sandbox build and push the service-runner image and submit a request for 110K granules. Verify that only the 4 expected failures happen for the request.

I've submitted several 110K granule requests as part of my sandbox testing and in every case only the 4 expected failures occurred.

## PR Acceptance Checklist
* [X] Acceptance criteria met

### DIFF
--- a/services/service-runner/app/service/service-runner.ts
+++ b/services/service-runner/app/service/service-runner.ts
@@ -404,9 +404,9 @@ export async function runServiceFromPull(
       });
 
       if ('retryable' in result) {
-        workItemLogger.debug(`Retryable error encountered (attempt ${retryCount + 1} of ${maxRetries})`);
-        await waitForContainerToStart('worker');
         retryCount += 1;
+        workItemLogger.debug(`Retryable error encountered (attempt ${retryCount} of ${maxRetries})`);
+        await waitForContainerToStart('worker');
       } else {
         return result;
       }
--- a/services/service-runner/app/service/service-runner.ts
+++ b/services/service-runner/app/service/service-runner.ts
@@ -13,6 +13,7 @@ import logger from '../../../harmony/app/util/log';
 import { objectStoreForProtocol } from '../../../harmony/app/util/object-store';
 import { resolve as resolveUrl } from '../../../harmony/app/util/url';
 import env from '../util/env';
+import { waitForContainerToStart } from '../util/k8s';
 
 const kc = new k8s.KubeConfig();
 kc.loadFromDefault();
@@ -28,6 +29,7 @@ export interface ServiceResponse {
   errorCategory?: string;
   hits?: number;
   scrollID?: string;
+  retryable?: boolean;
 }
 
 // how long to let a worker run before giving up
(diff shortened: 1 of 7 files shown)


services/service-runner/app/service/service-runner.ts
```diff
@@ -235,13 +239,13 @@ export async function runQueryCmrFromPull(
...
 ): Promise<object> {
-  let newFileContent;
-  const retryMessage = `Start of service execution (retryCount=${workItem.retryCount}, id=${workItem.id})`;
-  if (logs.length > 0 && (typeof logs[0] === 'string' || logs[0] instanceof String)) {
-    newFileContent = [retryMessage, ...logs];
-  } else {
-    newFileContent = [{ message: retryMessage }, ...logs];
-  }
+  let newFileContent = logs;
+  // const retryMessage = `${new Date().toISOString()} Start of service execution (retryCount=${workItem.retryCount}, id=${workItem.id})`;
```
indiejames, line 243: remove commented out code

services/service-runner/app/service/service-runner.ts
```diff
@@ -288,84 +287,138 @@ export async function runServiceFromPull(
...
+          clearTimeout(timeout);
+          workItemLogger.error('Kubernetes client exec caught exception:');
+          workItemLogger.error(e);
+          resolve({ error, errorLevel: 'error' });
+        });
       });
-    });
+
+      if ('retryable' in result) {
+        workItemLogger.debug(`Retryable error encountered (attempt ${retryCount + 1} of ${maxRetries})`);
```
ygliuvt, line 407: nit: Move line 409 up above this line, so you can log `...attempt ${retryCount}...`, rather than +1.

services/service-runner/app/service/service-runner.ts
```diff
@@ -288,84 +287,138 @@ export async function runServiceFromPull(
...
+    const commandAndArgs = [
+      ...commandLine,
+      '--harmony-action',
+      'invoke',
+      operationCommandLine,
+      operationCommandLineValue,
+      '--harmony-sources',
+      stacCatalogLocation,
+      '--harmony-metadata-dir',
+      `${catalogDir}`,
```
ygliuvt, line 318: why not just `catalogDir,`?

ygliuvt (approved):
Tested successfully locally. Sandbox testing will be straightforward, so I didn't repeat it.

indiejames (approved):

PR #753, opened 2025-05-05
Set up Harmony Filtering Service for UAT

srizvi-NASA:
Updated services-uat/env-defaults,/versions.ts for the deployment of the Filtering Service on UAT

## Jira Issue ID
HARMONY-2067

## Description
This PR defines the filtering service chain for UAT.

## Local Test Steps
The Test URL for this service:
/C1262899964-LARC_CLOUD/ogc-api-coverages/1.0.0/collections/all/coverage/rangeset?maxresults=1&forceasync=true

## PR Acceptance Checklist
* [ ] Acceptance criteria met

### DIFF
--- a/services/harmony/env-defaults
+++ b/services/harmony/env-defaults
@@ -538,3 +538,9 @@ HARMONY_SMAP_L2_GRIDDER_REQUESTS_MEMORY=128Mi
 HARMONY_SMAP_L2_GRIDDER_LIMITS_MEMORY=8Gi
 HARMONY_SMAP_L2_GRIDDER_INVOCATION_ARGS='python -m harmony_service'
 HARMONY_SMAP_L2_GRIDDER_SERVICE_QUEUE_URLS='["ghcr.io/nasa/harmony-smap-l2-gridder:latest,http://sqs.us-west-2.localhost.localstack.cloud:4566/000000000000/harmony-smap-l2-gridder.fifo"]'
+
+FILTERING_IMAGE=harmony/filtering:latest
+FILTERING_REQUESTS_MEMORY=4Gi
+FILTERING_LIMITS_MEMORY=6Gi
+FILTERING_INVOCATION_ARGS='./docker-entrypoint.sh'
+FILTERING_SERVICE_QUEUE_URLS='["harmony/filtering:latest,http://sqs.us-west-2.localhost.localstack.cloud:4566/000000000000/filtering.fifo"]'
--- a/config/services-uat.yml
+++ b/config/services-uat.yml
@@ -30,7 +30,7 @@ https://cmr.uat.earthdata.nasa.gov:
         env:
           <<: *default-turbo-env
           STAGING_PATH: public/harmony/filtering
-    umm_s: 'S1273163940-LARC_CLOUD'
+    umm_s: 'S1273529849-LARC_CLOUD'
     capabilities:
       concatenation: false
       subsetting:
--- a/config/services-uat.yml
+++ b/config/services-uat.yml
@@ -20,6 +20,31 @@ x-turbo-config: &default-turbo-config
 
 https://cmr.uat.earthdata.nasa.gov:
 
+  - name: asdc/filtering
+    description: Harmony service to filter netCDF based on user specified filtering criteria (spatial, min/max, glamping, and variable).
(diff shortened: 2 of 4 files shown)


chris-durbin:
The services.yml and UMM-S comparison script returned an error. The UMM-S record should be updated to drop the Aggregation section since the filtering service does not aggregate:

```
ServiceOptions: {
  Aggregation: {
    Concatenate: {
      ConcatenateDefault: true
    }
  }
}
```

services/harmony/env-defaults
```diff
@@ -538,3 +538,9 @@ HARMONY_SMAP_L2_GRIDDER_REQUESTS_MEMORY=128Mi
 HARMONY_SMAP_L2_GRIDDER_LIMITS_MEMORY=8Gi
 HARMONY_SMAP_L2_GRIDDER_INVOCATION_ARGS='python -m harmony_service'
 HARMONY_SMAP_L2_GRIDDER_SERVICE_QUEUE_URLS='["ghcr.io/nasa/harmony-smap-l2-gridder:latest,http://sqs.us-west-2.localhost.localstack.cloud:4566/000000000000/harmony-smap-l2-gridder.fifo"]'
+
+FILTERING_IMAGE=harmony/filtering:latest
+FILTERING_REQUESTS_MEMORY=4Gi
```
chris-durbin, line 543: We tend to set requests memory close to the minimum amount of memory the service will need for an average request since this reserves the memory on the node. Is every request going to take at least 4Gi, or can we set it to something lower like 256Mi?

services/harmony/env-defaults
```diff
@@ -538,3 +538,9 @@ HARMONY_SMAP_L2_GRIDDER_REQUESTS_MEMORY=128Mi
...
+FILTERING_LIMITS_MEMORY=6Gi
+FILTERING_INVOCATION_ARGS='./docker-entrypoint.sh'
```
chris-durbin, line 545: When attempting to run a request I get the error:
OCI runtime exec failed: exec failed: unable to start container process: exec: "./docker-entrypoint.sh": stat ./docker-entrypoint.sh: no such file or directory: unknown

services/harmony/env-defaults
```diff
@@ -538,3 +538,9 @@ HARMONY_SMAP_L2_GRIDDER_REQUESTS_MEMORY=128Mi
...
+FILTERING_SERVICE_QUEUE_URLS='["harmony/filtering:latest,http://sqs.us-west-2.localhost.localstack.cloud:4566/000000000000/filtering.fifo"]'
```
chris-durbin, line 546: @indiejames - should the queue urls be defined here or in `packages/util/env-defaults`? In deployed environments the default isn't used so it won't matter, but for local testing if you build the image locally I think it needs to be available for multiple harmony microservices.

chris-durbin (changes requested):
Requests are failing for me because the configured invocation args are pointing to `./docker-entrypoint.sh`, but the docker image says that file does not exist.

services/harmony/env-defaults
chris-durbin: This is fixed with the latest docker image.

services/harmony/env-defaults
srizvi-NASA: Not every request is going to take at least 4Gi. You can set it to something lower like 256Mi.

srizvi-NASA:
> The services.yml and UMM-S comparison script returned an error. The UMM-S record should be updated to drop the Aggregation section since the filtering service does not aggregate:
> 
> ```
> ServiceOptions: {
>   Aggregation: {
>     Concatenate: {
>       ConcatenateDefault: true
>     }
>   }
> }
> ```

A new umm-s record was created and referenced in the services-uat.yml.

chris-durbin (approved):
Successfully tested locally and in a sandbox environment.

services/harmony/env-defaults
indiejames: Every other service has it in `packages/util/env-defaults`, so I would put it there

indiejames (changes requested):
Will approve after `FILTERING_SERVICE_QUEUE_URLS` env var is moved to `packages/util/env-defaults`

indiejames (approved):

PR #757, opened 2025-05-14
HARMONY-2093: Add exponential backoff on servie-runner retries

indiejames:
## Jira Issue ID
HARMONY-2093

## Description
Add exponential backoff to service-runner retries to help with errors connecting to the worker. Previously we were checking the container status to see if it was ready before retrying, but possibly due to this bug (https://github.com/kubernetes/kubernetes/issues/106896) the container status was reporting `ready` very quickly and subsequent retries would fail quickly. This change simply waits an increasing amount before retrying. 

## Local Test Steps
There is no easy way to force an error like the ones we were seeing in production. We were unable to reproduce it in the sandbox. To test, build the service-runner image then restart your local deployment. Then run some queries of various sizes and make sure they complete.

## PR Acceptance Checklist
* [?] Acceptance criteria met

### DIFF
--- a/services/service-runner/app/service/service-runner.ts
+++ b/services/service-runner/app/service/service-runner.ts
@@ -320,7 +320,7 @@ export async function runServiceFromPull(
 
     let retryCount = 0;
     const maxRetries = 5;
-    let retryDelay = 3_000; // time in ms to wait before retrying
+    let retryDelay = 5_000; // time in ms to wait before retrying
     const retryDelayMultiplier = 2.0; // used to increase delay each retry
 
     while (retryCount < maxRetries) {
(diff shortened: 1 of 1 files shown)


services/service-runner/app/service/service-runner.ts
```diff
@@ -320,6 +320,8 @@ export async function runServiceFromPull(
 
     let retryCount = 0;
     const maxRetries = 5;
+    let retryDelay = 3_000; // time in ms to wait before retrying
```
chris-durbin, line 323: I think I want to see a slightly longer delay - maybe 5 seconds which would result in a little over 2 and a half minutes total to exhaust all retries.

chris-durbin (approved):

PR #760, opened 2025-05-16
TRT-668: Remove NetCDF-to-Zarr service chains.

owenlittlejohns:
## Jira Issue ID

[TRT-668](https://bugs.earthdata.nasa.gov/browse/TRT-668)

## Description

This PR removes the Harmony NetCDF-to-Zarr service and any service chains that had it as a step. (These longer service chains have versions already existing that run all the other steps)

I also had to regenerate a bunch of text fixtures. That got a bit involved, with some of the underlying metadata having changed in the years since some of the fixtures were generated.

## Local Test Steps

* Run this version of Harmony locally.
* Ensure it works for some sample requests (maybe running a regression test notebook)

## PR Acceptance Checklist
* [x] Acceptance criteria met


docs/guides/adapting-new-services.md
owenlittlejohns, line 1: I tried to update documentation where the NetCDF-to-Zarr service had been used as an example. I think I got most places. 🤞

services/harmony/fixtures/cmr.uat.earthdata.nasa.gov-443/160951032500011846
owenlittlejohns, line 1: There were a whole bunch of fixtures that needed updating. I probably could/should have been more targeted than nuking all the CMR fixtures. But it was a bit late by the time I realised that.

services/harmony/test/collection-capabilities.ts
```diff
@@ -4,18 +4,24 @@ import { currentApiVersion } from '../app/frontends/capabilities';
 import { hookGetCollectionCapabilities } from './helpers/capabilities';
 import hookServersStartStop from './helpers/servers';
 
+// This is the concept ID of the most recently updated collection with a
+// short name of "harmony_example". There are 3 collections with the same
+// short name, so if either of the other two are updated these tests will no
+// longer pass when regenerating CMR test fixtures.
```
owenlittlejohns, line 10: Hopefully this is a helpful note. It caught me out for a while where the wrong collection was being returned for this test.

services/harmony/test/ignore-errors.ts
```diff
@@ -15,14 +15,11 @@ import { resetQueues } from './helpers/queue';
...
   maxResults: 1,
-  outputCrs: 'EPSG:4326',
-  interpolation: 'near',
-  scaleExtent: '0,2500000.3,1500000,3300000',
-  scaleSize: '1.1,2',
-  format: 'application/x-zarr',
-  concatenate: false,
+  subset: 'lat(60:65)',
+  format: 'application/x-netcdf4',
+  forceAsync: true,
```
owenlittlejohns, line 22: I had to add `forceAsync: true` to the HOSS/MaskFill requests in a couple of places, as otherwise the condition for `numInputs <= maxSyncGranules` was letting Harmony try and do a synchronous request and that was timing out.

This feels like the simplest way to not end up doing that (without changing the `services-uat.yml` file), but definitely open to other ideas.

services/harmony/test/jobs/job-sharing.ts
```diff
@@ -11,6 +11,10 @@ const notJobOwner = 'jill'; // jill wants to access the results of joe's jobs
 
 const collectionWithEULAFalseAndGuestReadTrue = 'C1233800302-EEDTEST';
 const collectionWithEULATrueAndGuestsReadTrue = 'C1233860183-EEDTEST';
+// All collections in EEDTEST are in the All Collections permission group,
+// which provides read and order permissions to all guest users. This means
+// if a fixture is regenerated for C1233147317-EEDTEST it will need to be
+// manually updated to remove the "read" permission.
```
owenlittlejohns, line 17: I spent a little while on this one, too. It's not very satisfying to manually tweak the CMR response, but that permission that is set for all collections in EEDTEST will mean there's no EEDTEST collection we can use without guest read permissions.

services/harmony/test/ogc-api-coverages/describe-collections.ts
```diff
@@ -7,7 +7,7 @@ import { ServiceConfig } from '../../app/models/services/base-service';
 import { hookServices } from '../helpers/stub-service';
 
 describe('OGC API Coverages - describeCollections', function () {
-  const collection = 'C1225776654-ASF';
+  const collection = 'C1244141250-EEDTEST';
```
owenlittlejohns, line 10: The metadata for the ASF version of this Sentinel 1 collection has changed since the last time the test fixtures were run. This EEDTEST version is equivalent, and in the state the the ASF collection used to be.

services/harmony/test/parameters/concatenation.ts
```diff
@@ -182,52 +182,4 @@ describe('testing concatenation', function () {
       });
     });
   });
-
```
owenlittlejohns, line 185: I pulled this set of testing because there is no NetCDF-to-Zarr workflow. I'm not sure if this is a cause for concern, though, as it was testing a workflow where the default concatenation behaviour is to not, unlike concise.

owenlittlejohns (commented):
Also - I'll work up a PR for `harmony-ci-cd` early next week.

chris-durbin (approved):
Successfully verified in sandbox. Test changes look good in continuing to test the desired functionality just using different collections and services.

config/services-prod.yml
```diff
@@ -534,34 +534,6 @@ https://cmr.earthdata.nasa.gov:
...
-      The service attempts to optimize chunking in both time and space via heuristic algorithm in the output Zarr store.
-    data_operation_version: '0.21.0'
-    type:
-      <<: *default-turbo-config
-      params:
-        <<: *default-turbo-params
-        env:
-          <<: *default-turbo-env
-          STAGING_PATH: public/harmony/netcdf-to-zarr
-    umm_s: S2839491596-XYZ_PROV
```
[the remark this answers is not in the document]
owenlittlejohns: Totally agreed. I have [a ticket for that](https://bugs.earthdata.nasa.gov/browse/TRT-669). I'm not sure the best order of operations as this propagates:

Pieces of the puzzle:

* Removing the UMM-S records.
* Removing the NetCDF-to-Zarr value from `ng_LOCALLY_DEPLOYED_SERVICES` in Bamboo (and at the same time may as well pull any Bamboo environment variables for things like Docker images for the service).
* These changes hitting a Harmony environment. (Well, these and the `harmony-ci-cd` changes you've got in PR)

Maybe the second and third items happen at the same time, so the real question is when to remove the UMM-S records?

First maybe? Doing that would remove things from EDSC, and I think Harmony would just return an [UnsupportedOperation HttpError](https://github.com/nasa/harmony/blob/ba94e4c19aabb33c5d8c82fe0fdaa1d40d828d88/services/harmony/app/models/services/index.ts#L560), which would be consistent with later behaviour anyway. How does that sound?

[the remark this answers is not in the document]
chris-durbin: I ran a test to confirm harmony works as expected with services.yml pointing to a non-existent record (it will say the operation isn't supported), so from the harmony perspective order does not matter. I agree we should remove the UMM-S record first so that EDSC no longer presents the option. You can go ahead and remove the records now.

PR #763, opened 2025-05-21
HARMONY-2086: Run granule validation as part of query-cmr task when forceAsync is true.

ygliuvt:
## Jira Issue ID
HARMONY-2086

## Description
This PR moves the granule validation from the middleware to the query-cmr pod to minimize the time spent in request submission for `forceAsync=true` requests. It adds a `granValidation` object in extraArgs of the operation to indicate to query-cmr that granule validation needs to be performed rather than the normal granule queries done in query-cmr task. query-cmr uses `errorCategory: 'granValidation'` to indicate that the query-cmr result is the result of granule validation to service-runner and work-updater which will process the result accordingly. The work-updater will remove the `granValidation` from operation extraArgs when it sees a work item update with `message_category: 'granValidation'` and queue the query-cmr work item again so that it will be processed by query-cmr for the normal granule queries.

Note: We still have to save the query params in S3 in the middleware which could be time consuming for `forceAsync=true` requests, but there is no way around it.

### test specific granule id with forceAsync=true
http://localhost:3000/C1233800302-EEDTEST/ogc-api-coverages/1.0.0/collections/all/coverage/rangeset?subset=lat(20%3A60)&subset=lon(-140%3A-50)&granuleId=G1233800343-EEDTEST&outputCrs=EPSG%3A31975&format=image%2Fpng&forceAsync=true

### test maxResults with forceAsync=true, verify the correct granule limitation message is display on the job in workflow-ui
http://localhost:3000/C1233800302-EEDTEST/ogc-api-coverages/1.0.0/collections/all/coverage/rangeset?subset=lat(20%3A60)&subset=lon(-140%3A-50)&maxResults=1&outputCrs=EPSG%3A31975&format=image%2Fpng&forceAsync=true

### test bad granule id with forceAsync=true
http://localhost:3000/C1233800302-EEDTEST/ogc-api-coverages/1.0.0/collections/all/coverage/rangeset?subset=lat(20%3A60)&subset=lon(-140%3A-50)&granuleId=G123-EEDTEST&outputCrs=EPSG%3A31975&format=image%2Fpng&forceAsync=true

### test direct download with forceAsync=true, verify that the opendap links are not included in the result page
http://localhost:3000/C1233800302-EEDTEST/ogc-api-coverages/1.0.0/collections/parameter_vars/coverage/rangeset?variable=all&granuleId=G1233800343-EEDTEST&forceAsync=true

### test direct download with validation error
http://localhost:3000/C1233800302-EEDTEST/ogc-api-coverages/1.0.0/collections/parameter_vars/coverage/rangeset?variable=all&granuleId=G12300343-EEDTEST&forceAsync=true

Also test with regular requests (without forceAsync=true) and verify things work as before.

## PR Acceptance Checklist
* [x] Acceptance criteria met


services/harmony/app/models/services/base-service.ts
```diff
@@ -510,15 +510,18 @@ export default abstract class BaseService<ServiceParamType> {
 
           if (step.extra_args) {
             this.operation.extraArgs = step.extra_args;
-          } else {
-            // clear out extraArgs used by other steps
+          } else if (i > 1) {
```
chris-durbin, line 513: We shouldn't base this on the step number - we should change the check to explicitly list which extra args you do not want to remove.

I think rather than doing the check here it should be in the `removeExtraArgs` function.

However, I'm also realizing there's an issue with the download service - it uses extraArgs for the queryCmr step so line 512 is going to replace the granule validation extra args with the extraArgs from services.yml.

services/harmony/app/backends/workflow-orchestration/work-item-updates.ts
```diff
@@ -966,19 +1032,25 @@ export async function handleBatchWorkItemUpdatesWithJobId(
...
-
-    const preprocessedWorkItems: WorkItemUpdateQueueItem[] = await Promise.all(
-      groups[workflowStepIndex].map(async (item: WorkItemUpdateQueueItem) => {
-        const { update, operation } = item;
-        const result = await preprocessWorkItem(update, operation, logger, nextWorkflowStep);
-        item.preprocessResult = result;
-        return item;
-      }));
-    await processWorkItems(jobID, parseInt(workflowStepIndex), preprocessedWorkItems, logger);
+    if (parseInt(workflowStepIndex) === 1
```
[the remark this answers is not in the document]
ygliuvt: We could as the other two checks are like assertions -- they must be true when `groups[workflowStepIndex][0].update.message_category === 'granValidation'`.

services/harmony/app/models/services/base-service.ts
ygliuvt: Good point with the download service. I changed code to make download service work with granule validation when forceAsync=true.

services/harmony/app/backends/workflow-orchestration/work-item-polling.ts
```diff
@@ -224,15 +224,23 @@ export async function getWorkFromQueue(serviceID: string, reqLogger: Logger): Pr
     await queue.deleteMessage(queueItem.receipt);
     reqLogger.debug(`Deleted work item with receipt ${queueItem.receipt} from queue ${queueUrl}`);
     const item = JSON.parse(queueItem.body) as WorkItemData;
-    // get the operation for the step
-    const operationJson = await operationCache.fetch(`${item.workItem.jobID},${item.workItem.serviceID}`);
+
+    // always fetch for query-cmr service to make sure the update to operation
+    // for removing the extraArgs used for granule validation is picked up
+    const operationKey = `${item.workItem.jobID},${item.workItem.serviceID}`;
+    if (QUERY_CMR_SERVICE_REGEX.test(item.workItem.serviceID)) {
```
indiejames, line 231: Not sure I would even bother inserting into the cache here for queryCMR. I get that it makes the call on line 235 the same for all services, but I would probably just include that in an `else` and use the value returned by this fetch directly.

services/harmony/app/backends/workflow-orchestration/work-item-updates.ts
```diff
@@ -949,6 +949,72 @@ export async function processWorkItems(
...
+
+      if (status === WorkItemStatus.FAILED) {
+        // update job status and message
+        await completeJob(tx, job, JobStatus.FAILED, logger, message);
+      } else {
+        // update workflowstep operation
+        const thisWorkflowStep = await (await logAsyncExecutionTime(
+          getWorkflowStepByJobIdStepIndex,
+          'HWIUWJI.getWorkflowStepByJobIdStepIndex',
+          logger))(db, jobID, 1);
```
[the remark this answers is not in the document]
indiejames: Better to use `update.workflowStepIndex`

services/harmony/app/backends/workflow-orchestration/work-item-updates.ts
```diff
@@ -966,19 +1032,23 @@ export async function handleBatchWorkItemUpdatesWithJobId(
...
-
-    const preprocessedWorkItems: WorkItemUpdateQueueItem[] = await Promise.all(
-      groups[workflowStepIndex].map(async (item: WorkItemUpdateQueueItem) => {
-        const { update, operation } = item;
-        const result = await preprocessWorkItem(update, operation, logger, nextWorkflowStep);
-        item.preprocessResult = result;
-        return item;
-      }));
-    await processWorkItems(jobID, parseInt(workflowStepIndex), preprocessedWorkItems, logger);
+    if (groups[workflowStepIndex][0].update.message_category === 'granValidation') {
```
indiejames, line 1035: I think `granValidation` should be a `const` string defined somewhere or better yet an enumerated type.

ygliuvt: I will punt this for now -- hard to find a good location for this constant.

PR #781, opened 2025-06-17
HARMONY-2074: Allow harmony to be configured to fork one node process for each available CPU

chris-durbin:
## Jira Issue ID
HARMONY-2074

## Description
By default node will only use a single CPU. This change allows harmony to run on multiple CPUs by starting up one instance of the harmony application on each CPU.

## Local Test Steps
You can set `USE_CLUSTERING=true` in your .env file and when you start harmony up you will see that it forks multiple node processes (10 on my machine) that are all running harmony - note that each has its own separate memory space so local caching only works for a single process.

You would never want to do this locally - this is just for test purposes. In our AWS environments we will enable it which will start up two harmony processes on each of our EC2 instances.

By making this change in sandbox along with a different EC2 instance type we are able to scale up to handle 2000 simultaneous requests. Without this change we were still seeing many 504 errors during SSL negotiation.

I had to change the debug ports for the other apps in start-dev-services since starting harmony now starts a new debug port for each process beginning with 9200.

## PR Acceptance Checklist
* [X] Acceptance criteria met

### DIFF
--- a/services/harmony/app/util/cmr.ts
+++ b/services/harmony/app/util/cmr.ts
@@ -892,7 +892,7 @@ export enum cmrQueryType {
   COLL_JSON = 'coll_json',
   COLL_UMM = 'coll_umm',
   GRID = 'grid',
-  PERMISIONS = 'permissions',
+  PERMISSION = 'permission',
   SERVICE = 'service',
   VARIABLE = 'variable',
   VISUALIZATION = 'visualization',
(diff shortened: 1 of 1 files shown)


ygliuvt (approved):

services/harmony/app/server.ts
```diff
@@ -256,5 +260,26 @@ export async function stop({
 }
 
 if (require.main === module) {
-  start(process.env);
+  if (process.env.USE_CLUSTERING === 'true') {
```
indiejames, line 263: The only downside to this is that the forked processes won't share a caches

indiejames (approved):

PR #879, opened 2026-03-18
HARMONY-2248 Remove extraneous eslint-disable directives

flamingbear:
Pre-work for migration of eslint. This just removes unused (confusing) directives from the code.

These directives suppress rules that are either not configured at all
or explicitly set to "off" in the resolved ESLint configuration, meaning
they were never doing any work. Verified via `eslint --print-config`
and `eslint --report-unused-disable-directives`.

Rules removed:
- no-restricted-globals (not configured)
- no-param-reassign (not configured)
- no-loop-func (off)
- no-lonely-if (not configured)
- no-bitwise (not configured)
- import/first (not configured)
- import/prefer-default-export (not configured)
- import/no-named-as-default-member (off)
- import/order (not configured)
- max-classes-per-file (not configured)
- max-len (not configured)
- @typescript-eslint/no-empty-interface (not configured)
- @typescript-eslint/ban-types (off)
- node/no-unpublished-require (off in test overrides)
- no-new (not configured)

## Jira Issue ID

HARMONY-2248 (a)

## Description

Pre-work for migration of eslint.  This just removes unused (confusing) directives from the code.

  How we know these are safe to remove

  The project's default lint configuration does not flag unused eslint-disable directives. To find them, we ran ESLint with an additional flag:

  `npx eslint --report-unused-disable-directives .`

  This tells ESLint to actually run every enabled rule and then check whether each eslint-disable comment prevented a real violation. If ESLint reports zero problems for a directive, the directive is doing nothing — either:

  1. The rule isn't enabled in the resolved config, so there's nothing to suppress
  2. The code no longer triggers the rule, so the suppression is stale

  To distinguish between these two cases, we inspected the fully-resolved rule configuration (after all extends, plugins, and overrides are merged) using:

  `npx eslint --print-config <file>`

  This outputs every rule's final severity for a given file path. If the rule resolves to undefined or "off", the directive was always dead code (Commit 1). If the rule is active at "error" or "warn" but ESLint still reports no problem on that line, the code was refactored and the directive is stale (Commit 2).

  Commit 1: Rules not enabled in config (27 directives, 24 files)

  These directives suppress rules that resolve to undefined (not configured) or "off" in the merged ESLint config. They were never doing any work. This includes rules like no-param-reassign, no-restricted-globals, no-bitwise, max-len, etc. — commonly enabled in full airbnb but not included in airbnb-typescript/base which this project extends.

  Notably:
  - no-loop-func is explicitly "off" — the 4 test file disables were always redundant
  - node/no-unpublished-require is turned off in the test overrides, so test-file disables for it are redundant
  - @typescript-eslint/ban-types is explicitly "off" in root config

  Commit 2: Rules enabled but code no longer triggers them (16 directives, 9 files)

  These directives suppress rules that are active, but the code was refactored after the disable was added and no longer produces a
  violation:

  - @typescript-eslint/dot-notation (10 directives): Bracket notation was replaced with dot notation in prior changes
  - @typescript-eslint/naming-convention (5 directives): 4 were redundant with an existing file-level disable in cmr.ts; 1 file-level disable in memory-usage-collector.ts where no violations exist
  - prefer-destructuring (1 directive): Code in geo.ts already uses destructuring

  How to verify

  On main — shows 43 unused directives:
  `npx eslint --report-unused-disable-directives .`

  On this branch — shows 0:
  `npx eslint --report-unused-disable-directives .`

  Further Recommendation:

  Consider adding "reportUnusedDisableDirectives": true to the ESLint config (or --report-unused-disable-directives-severity error in the
  lint script) to catch stale directives automatically going forward.

## Local Test Steps

Make sure tests and lint still pass as before.

Also run `eslint --report-unused-disable-directives` and you should see no output 

## PR Acceptance Checklist
* [ ] Acceptance criteria met

## Summary by CodeRabbit

* **Style**
  * Removed linting directives across multiple files to enforce consistent code standards throughout the codebase.

### DIFF
--- a/packages/util/env.ts
+++ b/packages/util/env.ts
@@ -1,4 +1,3 @@
-/* eslint-disable @typescript-eslint/dot-notation */
 import _ from 'lodash';
 import * as dotenv from 'dotenv';
 import * as fs from 'fs';
--- a/services/cron-service/app/cronjobs/memory-usage-collector.ts
+++ b/services/cron-service/app/cronjobs/memory-usage-collector.ts
@@ -1,4 +1,3 @@
-/* eslint-disable @typescript-eslint/naming-convention */
 import { CloudWatchClient, GetMetricStatisticsCommand } from '@aws-sdk/client-cloudwatch';
 import { PutObjectCommand, S3Client } from '@aws-sdk/client-s3';
 import * as k8s from '@kubernetes/client-node';
--- a/services/cron-service/test/helpers/env.ts
+++ b/services/cron-service/test/helpers/env.ts
@@ -25,7 +25,6 @@ process.env.MAX_BATCH_SIZE_IN_BYTES = '10000';
 
 process.env.PORT = '4000';
 
-// eslint-disable-next-line import/first
 import env from '../../app/util/env'; // Must set required env before loading the env file
 
 env.nodeEnv = 'test';
--- a/services/harmony/app/backends/service-response.ts
+++ b/services/harmony/app/backends/service-response.ts
@@ -31,7 +31,6 @@ export interface CallbackQuery {
  * @param bbox - A bounding box
  */
 export function validateBbox(bbox: number[]): void {
-  // eslint-disable-next-line no-restricted-globals
   if (bbox.length !== 4 || bbox.some(isNaN)) {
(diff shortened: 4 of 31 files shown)


chris-durbin (approved):
In the next PR I'd recommended updating all the lint commands in package.json files to add --report-unused-disable-directives

PR #907, opened 2026-05-12
HARMONY-2342: Allows timeout overrides to consider collections.

flamingbear:
## Jira Issue ID

[HARMONY-2342](https://bugs.earthdata.nasa.gov/browse/HARMONY-2342)

## Description

Adds the ability to override work-failer's service timeouts by collection and adds net2cog overrides for the ASF GCOV collections.

getDefaultTimeoutSeconds updated to take collection concept IDs with precedence and now 
returns the override by precedence:
  1. per-collection override
  2. service-level default
  3. harmony default env.defaultTimeoutSeconds.

## Local Test Steps

Pull this branch and run the tests.

Update your .env with at least 16Gi (it will actually probably run with 8Gi)
NET2COG_LIMITS_MEMORY=16Gi

Run this request in your local Harmony-in-a-box.

http://localhost:3000/NISAR_L2_GCOV_BETA_V1/ogc-api-coverages/1.0.0/collections/parameter_vars/coverage/rangeset?forceAsync=true&granuleName=NISAR_L2_PR_GCOV_009_076_A_021_2005_QPDH_A_20260102T110439_20260102T110517_X05009_N_F_J_001&format=image%2Ftiff&variable=all&label=harmony-2342

Verify that the request completes without an retries having triggered and that it takes well over the default 5 min timeout.

## PR Acceptance Checklist
* [X] Acceptance criteria met

## Summary by CodeRabbit

* **New Features**
  * Enhanced timeout configuration to support per-collection overrides with sensible fallbacks to service or default thresholds.

* **Tests**
  * Added tests covering per-collection timeout behavior, matching logic, and fallback scenarios.

* **Chores**
  * Updated vulnerability policy to add/remove ignores.
  * Removed an unused dependency from a service package.

[Review Change Stack](https://app.coderabbit.ai/change-stack/nasa/harmony/pull/907)


services/work-failer/app/workers/failer.ts
```diff
@@ -70,7 +103,21 @@ export async function computeWorkItemDurationOutlierThresholdForJobService(
   serviceID: string,
   workflowStepIndex: number,
 ): Promise<number> {
-  let threshold = getDefaultTimeoutSeconds(serviceID) * 1000;
+  let collectionIds: string[] = [];
+  const serviceName = extractServiceName(serviceID);
+  const entry = serviceName ? serviceToDefaultTimeoutSeconds[serviceName] : undefined;
+  if (entry?.collections) {
+    try {
+      const step = await getWorkflowStepByJobIdStepIndex(db, jobID, workflowStepIndex);
```
flamingbear, line 111: If this is the wrong place to get the Ids, I'm all ears. But it's only called when there's a collection override for the service, but that's every net2cog job.

services/work-failer/app/workers/failer.ts
```diff
@@ -70,7 +103,21 @@ export async function computeWorkItemDurationOutlierThresholdForJobService(
...
+      if (step) {
+        collectionIds = step.collectionsForOperation();
+      }
+    } catch (e) {
+      log.warning(`Failed to load workflow step ${jobID}/${workflowStepIndex} for collection timeout lookup`);
+    }
+  }
+
+  let threshold = getDefaultTimeoutSeconds(serviceID, collectionIds) * 1000;
```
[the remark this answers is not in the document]
flamingbear: It's intended to default to the service default if it can't load the information

services/work-failer/app/workers/failer.ts
chris-durbin: An alternative would be to directly pull back the collection IDs from the job (maybe a new function so you are only pulling that single field back). You could also cache the mapping from job id to collection ids but that would only help for requests with a large number of granules - the bignbit case is only a single granule and that's the vast majority of the usage for net2cog.

In either case though I don't think this is much of a performance concern and what you have should work fine.

chris-durbin (approved):
Tested successfully by changing the timeout to 60 seconds for the specific collection and verified the work failer correctly used the 60 second timeout setting causing the work item to be retried.

PR #912, opened 2026-05-22
HARMONY-2300: Update dependencies and base images to reduce vulnerabilities in built microservices images.

chris-durbin:
## Jira Issue ID
HARMONY-2300

## Description
Updates base image used for our Docker images and ensures we update packages as part of building the images.

## Local Test Steps
```
npm run build && npm run build-all
```
Then bring up harmony in a box and verify that all the services continue to work without issue and requests complete successfully. There are no critical vulnerabilities reported after this change.

I also pushed the images to sandbox and tested there.

## PR Acceptance Checklist
* [X] Acceptance criteria met

### DIFF
--- a/services/cron-service/Dockerfile
+++ b/services/cron-service/Dockerfile
@@ -1,14 +1,14 @@
-FROM node:22.14.0
+FROM node:22-alpine3.22
 
-RUN apt-get update
-RUN apt-get install -y vim git python3 postgresql-client make gcc g++ libc-dev libpq-dev
+RUN apk update && apk upgrade --no-cache
+RUN apk add --no-cache vim git python3 postgresql-client make gcc g++ libc-dev postgresql-dev ca-certificates
 RUN git config --global url."https://".insteadOf ssh://
 
 RUN mkdir -p /cron-service/services/cron-service
 
 COPY package.json package-lock.json /cron-service/services/cron-service/
 WORKDIR /cron-service/services/cron-service
-RUN npm ci
+RUN npm ci --omit=dev
 
 COPY env-defaults /cron-service/services/cron-service/
 COPY built /cron-service/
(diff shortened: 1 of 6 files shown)


services/query-cmr/Dockerfile
```diff
@@ -1,15 +1,15 @@
-FROM node:22.14.0
+FROM node:22-alpine3.22
 
-RUN apt-get update
-RUN apt-get install -y vim
+RUN apk update && apk upgrade --no-cache
+RUN apk add --no-cache vim git
```
flamingbear: I think you can pull both vim and git here.

services/cron-service/Dockerfile
```diff
@@ -1,14 +1,14 @@
-FROM node:22.14.0
+FROM node:22-alpine3.22
 
-RUN apt-get update
-RUN apt-get install -y vim git python3 postgresql-client make gcc g++ libc-dev libpq-dev
+RUN apk update && apk upgrade --no-cache
+RUN apk add --no-cache vim git python3 postgresql-client make gcc g++ libc-dev postgresql-dev ca-certificates
 RUN git config --global url."https://".insteadOf ssh://
```
flamingbear: I don't want to second guess too many things, but what else in here is using git? I think git and vim could be nuked? it goes all the way back to the beginning.

flamingbear (commented):
Just commenting.  These are much improved according to dockers' vulnerability scans.  The only critical one I see is vim and even though that's not really reachable we could remove it and have none.  Also consider removing git and the git config ?

flamingbear (approved):
Looks good, none of the harmonyservice/serviceimages have critical vulnerabilities any longer.  I was able to fire up Harmony-In-A-Box and verify that requests are still processed as expected.

PR #917, opened 2026-06-04
HARMONY-2364: Make sure as work item updates are actively being worked SQS messages are not redelivered

chris-durbin:
## Jira Issue ID
HARMONY-2364

## Description
Make sure as work item updates are actively being worked SQS messages are not redelivered. The root cause of the bug for HARMONY-2364 is that the work item updates for the Geoloco work items in the request took over 100 seconds to process and the messages were being redelivered every 30 seconds because the message had not been deleted from the queue. When the message is redelivered because of the design of SQS the original worker is not able to delete the message from the queue (the only valid receipt handle that can delete the message is the last worker that received the message).

With this change the worker will extend the visibility timeout every 30 seconds to prevent the message from being redelivered.

## Local Test Steps
I built and pushed the work-updater image to sandbox and then tested in sandbox pointed against production using {root}/C1379758607-LAADS/ogc-api-coverages/1.0.0/collections/all/coverage/rangeset?subset=lon(77.19259999%3A77.19260000999999)&subset=lat(28.54499999%3A28.545000010000003)&format=application%2Fx-hdf&maxResults=100&skipPreview=true

I verified the message was logged every 30 seconds: 
```
{"env_name":"harmony-cdd-sandbox","level":"warn","message":"Extending message visibility by 60 seconds","timestamp":"2026-06-04T15:05:00.732Z"}
```

I saw work items completed successfully in the workflow-ui and no messages were added to the dead letter queue.

## PR Acceptance Checklist
* [X] Acceptance criteria met

### DIFF
--- a/services/harmony/app/util/queue/queue.ts
+++ b/services/harmony/app/util/queue/queue.ts
@@ -18,4 +18,10 @@ export abstract class Queue {
   abstract deleteMessage(receipt: string): Promise<void>;
   abstract deleteMessages(receipts: string[]): Promise<void>;
   abstract purge(): Promise<void>;
+  abstract changeMessageVisibility(
+    receipt: string, visibilityTimeoutSeconds: number
+  ): Promise<void>;
+  abstract changeMessageVisibilities(
+    receipts: string[], visibilityTimeoutSeconds: number
+  ): Promise<void>;
 }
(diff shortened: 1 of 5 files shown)


flamingbear (approved):
This code looks clean and the tests look comprehensive. While I'm still a bit uncertain how the bug was actually showing up -- I understand the message getting redelivered before it could complete and then not being able to be deleted -- I did see the previous infrastructure change to shuttle items into a DLQ after 5 retries, and now I am seeing the logging for "Extending message visibility"  (that's a long way to say it looks good to me, but you might want a second opinion)

PR #919, opened 2026-06-09
HARMONY-2354: Implements per-step paging on the steps endpoint.

flamingbear:
## Jira Issue ID

[HARMONY-2354](https://bugs.earthdata.nasa.gov/browse/HARMONY-2354)

## Description

Adds per-step paging to the steps endpoint:

Each step is paged independently via its `step<stepIndex>Page` query param  (e.g. `?step1Page=2&step2Page=1`)

`limit` is used as the page size for all steps

Each step's `paging` block is now `{ currentPage, lastPage, total, links }`  with `links` reusing `getPagingLinks`.

The block is shown when a step has more than one page or the requested page is  past the end (so paging off the end after a filter still links back)

## Local Test Steps

check out this branch and run the tests.
```
npm run test
```

deploy to Harmony-In-A-Box

Run a command that generates a number of pages, to test independent paging, we'll use the `limit` query param.

This request will find 7 granules and 6 that will complete.

http://localhost:3000/C1268429762-EEDTEST/ogc-api-coverages/1.0.0/collections/parameter_vars/coverage/rangeset?forceAsync=true&subset=lat(11.73504%3A43.2472)&subset=lon(5.0196%3A79.91089)&subset=time(%222015-03-31T17%3A06%3A00Z%22%3A%222015-03-31T17%3A10%3A00Z%22)&label=harmony-py&label=HARMONY-2254-verify&label=HARMONY-2254&format=image%2Ftiff&variable=Soil_Moisture_Retrieval_Data_1km%2Fretrieval_qual_flag_1km&variable=Soil_Moisture_Retrieval_Data_1km%2Fretrieval_qual_flag_apm_1km&variable=Soil_Moisture_Retrieval_Data_1km%2Fsoil_moisture_1km&variable=Soil_Moisture_Retrieval_Data_1km%2Fsoil_moisture_apm_1km&maxResults=20

(use your UUID:)
`> export UUID=77cc875f-4fd0-4c20-87ba-184f25718094`

Visualize all of the outputs you won't see any paging or paging objects.
`> open http://localhost:3000/jobs/${UUID}/steps`

default to limit=3
`> open "http://localhost:3000/jobs/${UUID}/steps?limit=3"`

open the metadata-annotator's second page:
`> open "http://localhost:3000/jobs/${UUID}/steps?limit=3&step3Page=2"`

Navigate around, verifying that only the step that was changed updates and the rest of the pages stay as they are (but also that the next/previous values on other steps have the correct pages to preserve other steps locations).

Change the page length and open metadata-annotator's second page, and HOSS's 10th:
This shows you that you can page multiple steps and that if you are beyond the last page, the last page is loaded for you and the page object will exist that has normal paging for HOSS step2 (current, previous, first) and the same for metadata annotator, step3

Old behavior was: ~you can still get back to results (HOSS shows a `first` link that will take you back to results, `previous` and `current` (that are both still beyond the bounds)~
`> open "http://localhost:3000/jobs/${UUID}/steps?limit=5&step3Page=2&step2Page=10"`

For more fun, you can just pop open a request that will generate 177 workItems:

http://localhost:3000/C1234088182-EEDTEST/ogc-api-coverages/1.0.0/collections/all/coverage/rangeset?format=image%2Ftiff

go to the steps endpoint for the job, verify the default of 50 workitems and that the paging works as expected.

  ## PR Acceptance Checklist
  * [X] Acceptance criteria met
  * [X] Harmony in a Box tested

### DIFF
--- a/services/harmony/app/frontends/steps.ts
+++ b/services/harmony/app/frontends/steps.ts
@@ -340,8 +340,8 @@ function buildSteps(
   const result: JobStep[] = [];
   const filtering = q.status !== undefined || q.workItem !== undefined;
   for (const { step, workItems, pagination } of stepResults) {
-    // Don't show steps without workitems.
-    if (filtering && workItems.length === 0) continue;
+    // Drop steps with no matching work items at all
+    if (filtering && pagination.total === 0) continue;
 
     const jobStep: JobStep = {
       serviceID: sanitizeImage(step.serviceID),
@@ -350,11 +350,12 @@ function buildSteps(
       statuses: statusCounts.get(step.stepIndex) ?? {},
       workItems: workItems.map((wi) => buildWorkItem(wi, resolved)),
     };
-    if (pagination.lastPage > 1) {
+    const { currentPage, lastPage, total } = pagination;
+    if (lastPage > 1 || currentPage > Math.max(lastPage, 1)) {
       jobStep.paging = {
-        currentPage: pagination.currentPage,
-        lastPage: pagination.lastPage,
-        total: pagination.total,
+        currentPage,
+        lastPage,
+        total,
         links: getPagingLinks(req, pagination, true, `step${step.stepIndex}Page`),
       };
     }
--- a/services/harmony/app/frontends/steps.ts
+++ b/services/harmony/app/frontends/steps.ts
@@ -22,7 +22,7 @@ import { getPagingLinks, parseIntegerParam } from '../util/pagination';
 import { readCatalogItems, StacItem } from '../util/stac';
 import { getRequestRoot } from '../util/url';
 
-export const DEFAULT_PER_PAGE = 50;
+const DEFAULT_PER_PAGE = 50;
 const MAX_STEP_PAGE_SIZE = 1000;
 const MAX_BATCH_CATALOGS = 5;
 const VALID_STATUSES = Object.values(WorkItemStatus);
@@ -308,8 +308,7 @@ function buildWorkItem(
 }
 
 
-// A workflow step paired with its requested page of work items and that page's
-// pagination info (used to build the step's paging links).
+// A workflow step, its page of work items and the pagination info.
 interface StepWorkItems {
   step: WorkflowStep;
   workItems: WorkItem[];
@@ -340,7 +339,7 @@ function buildSteps(
   const result: JobStep[] = [];
   const filtering = q.status !== undefined || q.workItem !== undefined;
   for (const { step, workItems, pagination } of stepResults) {
-    // Drop steps with no matching work items at all
+    // Don't show steps having no matching work items.
     if (filtering && pagination.total === 0) continue;
 
     const jobStep: JobStep = {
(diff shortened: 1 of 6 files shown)


services/harmony/app/util/pagination.ts
chris-durbin, line 52: Testing with `Limit=3` caused the parameter to be silently ignored and similar for `step2page=2`. I think we should treat the parameter as case insensitive similar to what we do on the coverages API and also validate the query parameters are known on the steps endpoint otherwise return a request validation error.

chris-durbin (commented):
Looks great just a couple of comments around edge cases and parameter handling.

When asking for a page that is beyond the last page (in my case `limit=3&step2Page=20` where there are only 50 items) the paging section has some confusing info - maybe in that scenario redirect it to the last page instead?:
```
"paging": {
"currentPage": 20,
"lastPage": 17,
"total": 50,
"links": [
{
"title": "The first page (1 of 17)",
"href": "http://localhost:3000/jobs/a4ee422d-e8b2-405b-b3f9-855610a649e2/steps?limit=3&step2Page=1",
"rel": "first",
"type": "application/json"
},
{
"title": "The previous page (19 of 17)",
"href": "http://localhost:3000/jobs/a4ee422d-e8b2-405b-b3f9-855610a649e2/steps?limit=3&step2Page=19",
"rel": "prev",
"type": "application/json"
},
{
"title": "The current page (20 of 17)",
"href": "http://localhost:3000/jobs/a4ee422d-e8b2-405b-b3f9-855610a649e2/steps?limit=3&step2Page=20",
"rel": "self",
"type": "application/json"
}
]
}
}
```

services/harmony/app/util/pagination.ts
chris-durbin: We have a function for putting all the query parameters in lower case:
`const query = keysToLowerCase(req.query);`

flamingbear: 3a463dd3f

services/harmony/app/frontends/steps.ts
```diff
@@ -366,6 +381,7 @@ export async function getJobSteps(
 ): Promise<void> {
   const { jobID } = req.params;
   try {
+    req.query = keysToLowerCase(req.query);
```
flamingbear, line 384: I didn't love modifying the input query, but I think the downstream user is the logger. I could do something else here if this is unacceptable.

services/harmony/test/jobs/job-status-pagination.ts
```diff
@@ -193,7 +193,7 @@ describe('Individual job status route - pagination', function () {
         const dataLinks = JSON.parse(this.res.text).links.filter((link) => link.rel === 'data');
         expect(dataLinks.length).to.equal(0);
       });
-      itIncludesPagingRelations(5, `jobs/${jobID}`, { first: 1, prev: 5, self: 6, next: null, last: null });
+      itIncludesPagingRelations(5, `jobs/${jobID}`, { first: 1, prev: null, self: 6, next: null, last: 5 });
     });
```
flamingbear, line 197: this is the new behavior for beyond the page. before it would give you the self of page 6 and previous as page 5.  Now we get rid of previous and add last. (also before if the requested page was 7, the previous would point to 6 which also didn't exist.)

chris-durbin (approved):

PR #920, opened 2026-06-11
HARMONY-2355: Implements multiple value query params.

flamingbear:
## Jira Issue ID

[HARMONY-2355](https://bugs.earthdata.nasa.gov/browse/HARMONY-2355) 

## Description

Update query parameters to handle multiple comma separated values. 
e.g. `?step=1,2`
Also, multiple repeated values are supported 
`?step=1&step=2`

## Local Test Steps

Pull and deploy this to Harmony-In-A-Box.

Run a request that will generate multiple pages of steps and will pause so you have multiple statuses.

http://localhost:3000/C1234088182-EEDTEST/ogc-api-coverages/1.0.0/collections/all/coverage/rangeset?format=image%2Ftiff

visit the endpoint and compare the outputs of: verify they don't error

http://localhost:3000/jobs/{:UUID}/steps
http://localhost:3000/jobs/{:UUID}/steps?status=ready
http://localhost:3000/jobs/{:UUID}/steps?status=ready,successful

## PR Acceptance Checklist
* [X] Acceptance criteria met

### DIFF
--- a/services/harmony/app/frontends/steps.ts
+++ b/services/harmony/app/frontends/steps.ts
@@ -20,6 +20,7 @@ import { Link } from '../util/links';
 import { keysToLowerCase } from '../util/object';
 import { defaultObjectStore } from '../util/object-store';
 import { getPagingLinks, parseIntegerParam } from '../util/pagination';
+import { parseMultiValueParameter } from '../util/parameter-parsing-helpers';
 import { readCatalogItems, StacItem } from '../util/stac';
 import { getRequestRoot } from '../util/url';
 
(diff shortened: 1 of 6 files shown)


chris-durbin (approved):
Tested successfully. The only thing I was thinking about is whether an out of bounds step number - for example step=3 when there are only 2 steps should be an error or return empty steps. It's currently returning emptying steps with a 200 response, and I'm thinking that probably is the best behavior.

flamingbear:
> It's currently returning emptying steps with a 200 response, and I'm thinking that probably is the best behavior.

~That was what we (maybe I) decided on for the case when you've gone off the end somehow, the return leaves you enough paging information to get back to the first page (and maybe last?). It can happen if you're deep in the steps and then apply a different query filter, like success only or failed only.  It should return paging information so that you can quickly get to a valid page~

~But now I'm thinking this is behavior from the last PR that I'm describing... let me see what happens with multiple query params and one off the end. And for paging I actually return the last page of values.~

Reading is fundamental. Ok steps, I'm following.  I think it's fine, but happy to change it if you change your mind.

PR #956, opened 2026-09-17
HARMONY-2444: Force datetime strings to UTC to avoid local timezone issues

indiejames:
Force datetime strings to UTC to avoid local timezone issues

## Jira Issue ID
HARMONY-2444

## Description
UTC based timestamps read from Iceberg were being interpreted as local timezone times. This forces them to be interpreted as UTC regardless of the local timezone, which fixes a bug where rows were being skipped when running in the local dev environment (unless you happened to be running in Caimbridge).

## Local Test Steps
1. Start Harmony
2. Execute a query.
3. Use duckdb to access the Iceberg `jobs` table to count a row count
4. Execute another query
5. Wait until the next minute
6. Get the row count for the jobs table using duckdb
7. Verify that it has increased by 1 

## PR Acceptance Checklist
* [x] Acceptance criteria met

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of UTC timestamps when determining the latest analytics table update.
  - Explicit numeric UTC offsets are now preserved and normalized consistently.
  - Timestamps without an offset are interpreted as UTC, while zero offsets are normalized to standard UTC notation.
  - Surrounding whitespace is handled consistently during timestamp parsing.
  - Invalid date values continue to be rejected with an error instead of being silently misinterpreted.


services/cron-service/app/cronjobs/update-analytics.ts
```diff
@@ -75,6 +75,30 @@ async function getPostgresRows(ctx: Context, table: string, latestUpdateTime: st
...
+ * @returns {Date}
+ */
+function parseUtcTimestamp(dateString: string): Date {
+  let formatted = dateString.trim().replace(' ', 'T');
+
+  // Handle "+00" or explicit offset at the end
+  if (formatted.endsWith('+00')) {
+    formatted = formatted.slice(0, -3) + 'Z';
+  } else if (!formatted.endsWith('Z')) {
+    formatted += 'Z';
```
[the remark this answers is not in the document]
indiejames: The connection is explicitly set to use UTC, so values are always returned in UTC

services/cron-service/app/cronjobs/update-analytics.ts
```diff
@@ -75,6 +75,36 @@ async function getPostgresRows(ctx: Context, table: string, latestUpdateTime: st
   return result;
 }
 
+/**
+ * Parse a datetime string for a UTC time into a Date
+ * @param dateString - the string to parse
+ * @returns the Date object for the UTC timestamp
+ */
+function parseUtcTimestamp(dateString: string): Date {
```
chris-durbin: Might be worth a couple of unit tests with examples to make it clear what this handles - although this is not a generic function meant to handle everything it just needs to handle the one case of timestamp strings returned from DuckDB.

chris-durbin (approved):

ygliuvt (approved):

services/cron-service/app/cronjobs/update-analytics.ts
indiejames: Done

